PR #10, opened 2022-08-02
Implement bitmap setters, getters, and element-wise builder

### LINKED ISSUE #4: Implement bitmap helpers
Dealing with bitmaps is hard and we should provide some tools to do the finicky hard-to-get-right math:

- `ArrowBitmapIsNull(const void* bitmap, int64_t i)`
- `ArrowBitmapSetNull(void* bitmap, int64_t i, char value)`
- `ArrowBitmapCountNull(const void* bitmap, int64_t bitmap_size_bits)`

I think we also need a builder dedicated to bitmaps since building them up incrementally (without allocating if there are no null) is much of the use case. Something like

```
struct ArrowBitmapBuilder { struct ArrowBuffer buffer; int64_t n_values; uint8_t pending_values[64]; int8_t pending_values_offset; }
ArrowBitmapBuilderAppendInt8(struct ArrowBitmapBuilder* bitmap_builder, const int8_t* values, int64_t n_values);
ArrowBitmapBuilderAppendInt32(struct ArrowBitmapBuilder* bitmap_builder, const int32_t* values, int64_t n_values);
ArrowBitmapBuilderReset(struct ArrowBitmapBuilder* bitmap_builder);
```

(because char and int arrays are common ways that these things are stored)

paleolimbot:
Fixes #4.


lidavidm:
I guess the general comment is: as Wes suggests, many of these should be defined inline in the header so they can get inlined into the caller, right?

wesm:
Also, keep in mind that in C, unlike C++, it takes more work to guarantee that functions will actually be inlined in all compilers. You need something like

```
#ifdef _MSC_VER
#   define always_inline __forceinline
#elif defined(__GNUC__)
#   define always_inline __attribute__((always_inline)) inline
#else
#   define always_inline inline
#endif
```

I've always found it easier to use macros (as long as you don't put anything too complex in the macro which can make debugging more difficult)

pitrou:
> Also, keep in mind that in C, unlike C++, it takes more work to guarantee that functions will actually be inlined in all compilers.

But it's also most of the time pointless. The compiler is usually able to reason on code performance better than the developer, except in fringe cases.

I would be in favor of inline functions over macros, which have a non-trivial debugging and maintainability cost.

For the record, CPython has started migrating from macros over to inline functions.

wesm:
> I would be in favor of inline functions over macros, which have a non-trivial debugging and maintainability cost.
>
> For the record, CPython has started migrating from macros over to inline functions.

That's great feedback @pitrou, sounds good to me

lidavidm:
I don't think we need to split up header files or rearrange things, though, and I don't think we need to inline the buffer functions - it could just be moving the definitions from `bitmap.c` to `nanoarrow.h` and adding `inline`

lidavidm:
But if it's easier to do things in stages then we can just review the current PR?

src/nanoarrow/typedefs_inline.h
```diff
@@ -0,0 +1,172 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#ifndef NANOARROW_TYPEDEFS_INLINE_H_INCLUDED
+#define NANOARROW_TYPEDEFS_INLINE_H_INCLUDED
+
+#include <stdint.h>
+
+/// \defgroup nanoarrow-inline-typedef Type definitions used in inlined implementations
```
lidavidm, line 23: This still needs the `extern "C"`

src/nanoarrow/nanoarrow.h
```diff
@@ -483,82 +372,117 @@ ArrowErrorCode ArrowSchemaViewInit(struct ArrowSchemaView* schema_view,
...
-                                 int64_t size_bytes);
+static inline ArrowErrorCode ArrowBufferAppend(struct ArrowBuffer* buffer,
+                                               const void* data, int64_t size_bytes);
+
+/// }@
+
+/// \defgroup nanoarrow-bitmap Bitmap utilities
+
+/// \brief Extract a boolean value from a bitmap
+static inline int8_t ArrowBitmapElement(const void* bitmap, int64_t i);
```
lidavidm, line 439: nit: `ArrowBitmapGet`/Set/CountSet/CountUnset/etc.?

src/nanoarrow/nanoarrow.c
```diff
@@ -17,6 +17,7 @@
 
 #include "allocator.c"
 #include "buffer.c"
+#include "bitmap.c"
```
lidavidm, line 20: Does this file exist?

src/nanoarrow/bitmap_inline.h
```diff
@@ -0,0 +1,131 @@
...
+#ifndef NANOARROW_BITMAP_INLINE_H_INCLUDED
+#define NANOARROW_BITMAP_INLINE_H_INCLUDED
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "buffer_inline.h"
+#include "typedefs_inline.h"
+
+static inline int8_t ArrowBitmapElement(const void* bitmap, int64_t i) {
```
lidavidm, line 27: It may be worth stealing Arrow's implementations of these utilities: https://github.com/apache/arrow/blob/3b987d92d14ce7b5f5ccd2afb7366273e20348d4/cpp/src/arrow/util/bit_util.h#L296

src/nanoarrow/bitmap_inline.h
```diff
@@ -0,0 +1,131 @@
...
+  } else {
+    bitmap_char[i / 8] &= ~mask;
+  }
+}
+
+static inline int64_t ArrowBitmapCountTrue(const void* bitmap, int64_t i_from,
+                                           int64_t i_to) {
+  int64_t count = 0;
+  for (int64_t i = i_from; i < i_to; i++) {
+    count += ArrowBitmapElement(bitmap, i);
```
lidavidm, line 46: I wonder if this optimized to popcnt or not…but Arrow also has implementations to steal!

lidavidm (commented):
I suppose I'm a little confused on why we have the buffer (to avoid data dependencies?)

lidavidm:
> I'm not sure which buffer you mean! The `struct ArrowBuffer`?

Sorry, I mean the buffer of values inside the bitmap builder

src/nanoarrow/nanoarrow.h
paleolimbot: When I stole Arrow's implementations I also stole all the names I saw! The raw bit functions became `ArrowBitXXX` and the functions that operate on an owning `struct ArrowBitmap` became `ArrowBitmapXXX` to match the `ArrowBuffer` functions...

src/nanoarrow/bitmap_inline.h
paleolimbot: Done! At least to the extent needed to bitpack a char (DuckDB), an int (R), append a bunch of nulls at once, and calculate a null count.

src/nanoarrow/bitmap_inline.h
paleolimbot: I passed on the compiler intrinsics for now because I don't have CI to test multiple compilers and make sure that they work or benchmarks set up to make sure they're worth it...I used the pre-computed `kpopcount` array which is much better than the previous version.

src/nanoarrow/buffer_inline.h
```diff
@@ -15,14 +15,20 @@
...
 
-#include "nanoarrow.h"
+#include "typedefs_inline.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
 
-static int64_t ArrowGrowByFactor(int64_t current_capacity, int64_t new_capacity) {
+static inline int64_t _ArrowGrowByFactor(int64_t current_capacity, int64_t new_capacity) {
```
lidavidm, line 31: It's not allowed to start names with an underscore: https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html

src/nanoarrow/bitmap_inline.h
```diff
@@ -0,0 +1,323 @@
...
+#include "typedefs_inline.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static const uint8_t _ArrowkBitmask[] = {1, 2, 4, 8, 16, 32, 64, 128};
+static const uint8_t _ArrowkFlippedBitmask[] = {254, 253, 251, 247, 239, 223, 191, 127};
+static const uint8_t _ArrowkPrecedingBitmask[] = {0, 1, 3, 7, 15, 31, 63, 127};
+static const uint8_t _ArrowkTrailingBitmask[] = {255, 254, 252, 248, 240, 224, 192, 128};
```
lidavidm, line 34: Ditto the comment about underscores in names here (unfortunately).

lidavidm (commented):
Generally looks good to me, just the comments about naming.

src/nanoarrow/buffer_inline.h
pitrou: It's used in many C projects though, so most probably can be considered safe.

paleolimbot: I see...I was copying the pattern used by headers generated by nanopb ("private" inline functions). Is there a better pattern for functions that have to be visible for inline functions but that shouldn't be accessed otherwise?

paleolimbot: Maybe `ArrowPrivateXXX`?

lidavidm: Ah, interesting. I agree it's probably safe. It wouldn't be an issue if we have to change it later for some reason.

lidavidm (approved):

PR #12, opened 2022-08-03
Add metadata builder functions

### LINKED ISSUE #6: Implement metadata building utility
Now that we have growable buffers we can implement a more natural way to build key/value metadata. It might just take one function:

```
ArrowMetadataBuild(struct ArrowBuffer* buffer, const char* key, const char* value);
```

Sending it to a schema is a tiny bit awkward without a helper:

```
ArrowSchemaSetMetadata(schema, (const char*)buffer.data);
ArrowBufferReset(buffer);
```

(or maybe there's a more natural way to do this with better names)

paleolimbot:
Fixes #6.

This PR adds functions `ArrowMetadataBuilderAppend()` (for blindly but efficiently adding a key/value pair to the end of some metadata) and `ArrowMetadataBuilderSet()` (to less efficiently replace or remove a value for key).

The use case I had in mind is building extension type metadata from some input (i.e., make an output schema like the input except with new extension type or with new serialized extension type metadata). It's rather difficult to replicate the "replace" or "remove" behaviour otherwise.

### DIFF
--- a/src/nanoarrow/metadata.c
+++ b/src/nanoarrow/metadata.c
@@ -21,7 +21,7 @@
 
 #include "nanoarrow.h"
 
-struct ArrowStringView ArrowStringViewCreate(const char* value) {
+struct ArrowStringView ArrowCharView(const char* value) {
   struct ArrowStringView out;
 
   out.data = value;
@@ -131,7 +131,7 @@ ArrowErrorCode ArrowMetadataGetValue(const char* metadata, const char* key,
 }
 
 char ArrowMetadataHasKey(const char* metadata, const char* key) {
-  struct ArrowStringView value = ArrowStringViewCreate(NULL);
+  struct ArrowStringView value = ArrowCharView(NULL);
   ArrowMetadataGetValue(metadata, key, &value);
   return value.data != NULL;
 }
@@ -194,7 +194,7 @@ static ArrowErrorCode ArrowMetadataBuilderSetInternal(struct ArrowBuffer* buffer
                                                       struct ArrowStringView* key,
                                                       struct ArrowStringView* value) {
   // Inspect the current value to see if we can avoid copying the buffer
-  struct ArrowStringView current_value = ArrowStringViewCreate(NULL);
+  struct ArrowStringView current_value = ArrowCharView(NULL);
   int result =
       ArrowMetadataGetValueInternal((const char*)buffer->data, key, &current_value);
   if (result != NANOARROW_OK) {
--- a/src/nanoarrow/metadata.c
+++ b/src/nanoarrow/metadata.c
@@ -84,10 +84,10 @@ int64_t ArrowMetadataSizeOf(const char* metadata) {
   return size;
 }
 
-ArrowErrorCode ArrowMetadataGetValueView(const char* metadata,
-                                         struct ArrowStringView* key,
-                                         const char* default_value,
-                                         struct ArrowStringView* value_out) {
+static ArrowErrorCode ArrowMetadataGetValueView(const char* metadata,
+                                                struct ArrowStringView* key,
+                                                const char* default_value,
+                                                struct ArrowStringView* value_out) {
   value_out->data = default_value;
   if (default_value != NULL) {
     value_out->n_bytes = strlen(default_value);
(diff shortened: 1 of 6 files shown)


src/nanoarrow/nanoarrow.h
```diff
@@ -261,6 +261,24 @@ ArrowErrorCode ArrowMetadataGetValue(const char* metadata, const char* key,
                                      const char* default_value,
                                      struct ArrowStringView* value_out);
 
+/// \brief Initialize a builder for schema metadata from key/value pairs
+ArrowErrorCode ArrowMetadataBuilderInit(struct ArrowBuffer* buffer, const char* metadata);
```
lidavidm, line 265: The `metadata` param is an existing metadata buffer? (It's also not tested)

src/nanoarrow/metadata.c
```diff
@@ -114,8 +114,156 @@ ArrowErrorCode ArrowMetadataGetValue(const char* metadata, const char* key,
...
+  ArrowBufferInit(buffer);
+  int result = ArrowBufferAppend(buffer, metadata, ArrowMetadataSizeOf(metadata));
+  if (result != NANOARROW_OK) {
+    return result;
+  }
+
+  return NANOARROW_OK;
+}
+
+ArrowErrorCode ArrowMetadataBuilderAppendView(struct ArrowBuffer* buffer,
```
lidavidm, line 141: Worth possibly exposing this variant too?

src/nanoarrow/metadata.c
```diff
@@ -114,8 +114,156 @@ ArrowErrorCode ArrowMetadataGetValue(const char* metadata, const char* key,
...
+                                              struct ArrowStringView* key,
+                                              struct ArrowStringView* value) {
+  if (value == NULL) {
+    return NANOARROW_OK;
+  }
```
lidavidm, line 146: Hmm, to me it's a little weird to accept NULL as the value and then just do nothing with it. If we just considered `append(key, NULL)` to be an error, we could drop this, and then we could pass the views by value instead of indirecting through a pointer

src/nanoarrow/nanoarrow.h
paleolimbot: (Documented and tested now)

src/nanoarrow/metadata.c
paleolimbot: I pushed a compromise of sorts...all values are now `ArrowStringView` but all keys stay `const char*`, which is somewhere inbetween the programmability and non-null terminated-ness of the `ArrowStringView` and the reality that keys are almost always `"a literal string"`. I tried having both be `ArrowStringView` but that was rather painful to actually use.

src/nanoarrow/metadata.c
paleolimbot: I moved these to be internal implementation details...the `NULL` mostly just helps with minimizing the number of times one has to copy a metadata string (correspondingly, I added `ArrowMetadatBuilderRemove()` to make it more explicit).

src/nanoarrow/metadata.c
lidavidm: Wonder if a helper macro to convert `const char*` to `struct ArrowStringView` might help work around that?

paleolimbot: I added `ArrowCharView()`:

https://github.com/apache/arrow-nanoarrow/blob/1a4bb88165ad659c0db274df24a44987c8a81bf3/src/nanoarrow/nanoarrow.h#L237-L238

It's still a little awkward feeling if I convert keys to use `struct ArrowStringView` but I don't really mind either way.

lidavidm: not a big deal for me either, though it is a little awkward to have the types differ between key/value even if there is a pragmatic reason

paleolimbot: I'll change them to be the string views...anybody looking for convenience isn't writing C code anyway!

src/nanoarrow/typedefs_inline.h
```diff
@@ -166,6 +166,34 @@ enum ArrowType {
...
+};
+
+ /// \brief Create a string view from a null-terminated string
+static inline struct ArrowStringView ArrowCharView(const char* value) {
+  struct ArrowStringView out;
+
+  out.data = value;
+  out.n_bytes = 0;
+  if (value) {
+    while (*value++) {
```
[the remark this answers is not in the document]
paleolimbot: Yes...too much of a hack? I'll check if string.h gets included in another inline file anyway (it probably does).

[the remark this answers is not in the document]
lidavidm: +I don't think it's a big deal to include string.h (it'll certainly raise less eyebrows than this)

PR #14, opened 2022-08-05
Owning/mutable `struct ArrowArray`

### LINKED ISSUE #5: Implement `ArrowArray*` helpers
Now that we have buffer holders (`struct ArrowBuffer`) we can implement an owning `struct ArrowArray`. I envision the API something like the `ArrowSchema*` helpers:

- `ArrowArrayInit(struct ArrowArray* array, int64_t n_buffers)`
- `ArrowArrayAllocateChildren(struct ArrowArray* array, int64_t n_children)`
- `ArrowArraySetBuffer(struct ArrowArray* array, int64_t i, struct ArrowBuffer* buffer)` (moves the `struct ArrowBuffer` so that it lives in memory owned by `array->private_data` and sets `array->buffers[i] = buffer->data`).

@lidavidm commented:

Maybe `Init` could take the type and then it would know how many buffers to allocate? (Ditto for `AllocateChildren`)

@paleolimbot commented:

Ah that makes sense, and parallels `ArrowSchemaInit()` (where the caller can fall back on `NANOARROW_TYPE_UNINITIALIZED` if they're implementing their own logic). I think `ArrowArrayAllocateChildren()` has to exist on its own, but the utility of a helper like `ArrowArrayInitFromSchemaView(struct ArrowArray* array, struct ArrowSchemaView* view, struct ArrowError* error)` might become clear at some point (since the schema view contains all the information needed to initialize all the buffers for all the recursive children).

paleolimbot:
Fixes #5 by implementing an Array whose buffer lifecycle is handled by `struct ArrowBuffer`.

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -15,9 +15,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include "allocator.c"
-#include "array.c"
-#include "error.c"
-#include "metadata.c"
-#include "schema.c"
-#include "schema_view.c"
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "nanoarrow.h"
+
+ArrowErrorCode ArrowArrayInit(struct ArrowArray* array, struct ArrowSchema* schema) {
+    return NANOARROW_OK;
+}
+
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -153,3 +153,68 @@ ArrowErrorCode ArrowArrayInit(struct ArrowArray* array, enum ArrowType storage_t
 
   return NANOARROW_OK;
 }
+
+ArrowErrorCode ArrowArrayAllocateChildren(struct ArrowArray* array, int64_t n_children) {
+  if (array->children != NULL) {
+    return EINVAL;
+  }
+
+  if (n_children == 0) {
+    return NANOARROW_OK;
+  }
+
+  array->children =
+      (struct ArrowArray**)ArrowMalloc(n_children * sizeof(struct ArrowArray*));
+  if (array->children == NULL) {
+    return ENOMEM;
+  }
+
+  for (int64_t i = 0; i < n_children; i++) {
+    array->children[i] = NULL;
+  }
+
+  for (int64_t i = 0; i < n_children; i++) {
+    array->children[i] = (struct ArrowArray*)ArrowMalloc(sizeof(struct ArrowArray));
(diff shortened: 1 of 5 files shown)


src/nanoarrow/typedefs_inline.h
```diff
@@ -165,6 +212,20 @@ struct ArrowBitmap {
   int64_t size_bits;
 };
 
+/// \brief A structure used as the private data member for ArrowArrays allocated here
```
lidavidm, line 215: nit: does this need to be in the public header?

lidavidm (approved):

src/nanoarrow/typedefs_inline.h
paleolimbot: The definition needs to be visible for the appenders to be inlined (I removed the documentation comments and gave it a scarier name of `ArrowArrayPrivateData` to hopefully make that clearer).

src/nanoarrow/array.c
```diff
@@ -0,0 +1,220 @@
...
+    case NANOARROW_TYPE_UNINITIALIZED:
+    case NANOARROW_TYPE_NA:
+      array->n_buffers = 0;
+      break;
+
+    case NANOARROW_TYPE_LIST:
+    case NANOARROW_TYPE_LARGE_LIST:
+    case NANOARROW_TYPE_STRUCT:
+    case NANOARROW_TYPE_MAP:
+    case NANOARROW_TYPE_SPARSE_UNION:
```
lidavidm, line 79: Shouldn't fixed-size list also go here?

src/nanoarrow/nanoarrow.h
```diff
@@ -503,6 +456,35 @@ static inline void ArrowBitmapReset(struct ArrowBitmap* bitmap);
...
+
+/// \brief Allocate the array->children array
+///
+/// Includes the memory for each child struct ArrowArray.
+/// schema must have been allocated using ArrowArrayInit.
+ArrowErrorCode ArrowArrayAllocateChildren(struct ArrowArray* array, int64_t n_children);
+
+/// \brief Allocate the array->dictionary member
+///
+/// array must have been allocated using ArrowArrayInit
```
lidavidm, line 477: I suppose afterwards, the dictionary/each child should be in turn initialized with ArrowArrayInit?

paleolimbot: They don't have to to avoid leaking memory (each member is marked released initially), but functionally that's probably what most people should do (or create the child arrays separately and move them). I added both those bits to the documentation.

lidavidm (approved):

PR #17, opened 2022-08-10
Buffer element appenders

paleolimbot:
It turns out this is really annoying to do otherwise! Declaring a variable of an appropriate type gets verbose when switching on type, and it sounds like these functions might be useful for database drivers, too.

### DIFF
--- a/src/nanoarrow/buffer_test.cc
+++ b/src/nanoarrow/buffer_test.cc
@@ -160,3 +160,31 @@ TEST(BufferTest, BufferTestError) {
 
   ArrowBufferReset(&buffer);
 }
+
+TEST(BufferTest, BufferTestAppendHelpers) {
+  struct ArrowBuffer buffer;
+  ArrowBufferInit(&buffer);
+
+  EXPECT_EQ(ArrowBufferAppendInt8(&buffer, 123), NANOARROW_OK);
+  EXPECT_EQ(reinterpret_cast<int8_t*>(buffer.data + buffer.size_bytes)[-1], 123);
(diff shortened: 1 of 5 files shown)


src/nanoarrow/buffer_test.cc
```diff
@@ -160,3 +160,31 @@ TEST(BufferTest, BufferTestError) {
...
+
+TEST(BufferTest, BufferTestAppendHelpers) {
+  struct ArrowBuffer buffer;
+  ArrowBufferInit(&buffer);
+
+  EXPECT_EQ(ArrowBufferAppendInt8(&buffer, 123), NANOARROW_OK);
+  EXPECT_EQ(reinterpret_cast<int8_t*>(buffer.data + buffer.size_bytes)[-1], 123);
+  EXPECT_EQ(ArrowBufferAppendUInt8(&buffer, 123), NANOARROW_OK);
+  EXPECT_EQ(reinterpret_cast<uint8_t*>(buffer.data + buffer.size_bytes)[-1], 123);
+  EXPECT_EQ(ArrowBufferAppendInt16(&buffer, 123), NANOARROW_OK);
```
lidavidm, line 172: I think these are all unaligned accesses; that's probably OK, but I think Arrow itself tries to avoid this (or uses memcpy to do unaligned reads) to be portable. That said this is test code so not a big deal.

lidavidm (approved):

src/nanoarrow/buffer_test.cc
paleolimbot: It's easy enough to reset!

lidavidm (approved):

PR #19, opened 2022-08-11
ArrowArray consumer buffer helpers

paleolimbot:
Implements a few types whose goal is to provide access to buffers. This will require a `switch(storage_type)` by the consumer and requires some knowledge of the spec (getters to parallel the setters in #16 is another option for a future PR!).


src/nanoarrow/typedefs_inline.h
```diff
@@ -179,6 +217,19 @@ struct ArrowStringView {
...
+struct ArrowBufferView {
+  /// \brief A pointer to the start of the buffer
+  ///
+  /// If n_bytes is 0, this value may be NULL.
+  const union ArrowBufferDataPointer data;
+
+  /// \brief The size of the string in bytes,
+  ///
+  /// (Not including the null terminator.)
+  int64_t n_bytes;
```
lidavidm, line 230:
```suggestion
  /// \brief The size of the buffer in bytes.
  int64_t n_bytes;
```

src/nanoarrow/utils_inline.h
```diff
@@ -26,6 +26,115 @@
 extern "C" {
 #endif
 
+static inline void ArrowLayoutInit(struct ArrowLayout* layout,
```
lidavidm, line 29: Just an observation, but I think this sort of function doesn't necessarily have to be inline - it shouldn't be called frequently and shouldn't be a bottleneck

lidavidm: Or, of course, you could go the other way and put _everything_ in headers, making this a header-only library

src/nanoarrow/typedefs_inline.h
paleolimbot: Done!

src/nanoarrow/utils_inline.h
paleolimbot: Done! I'm sure it *could* be header-only but that's a discussion/battle for another day that I'm not all that qualified to weigh in on.

src/nanoarrow/typedefs_inline.h
```diff
@@ -166,6 +166,32 @@ enum ArrowType {
...
+/// the length and offset of the array.
+struct ArrowLayout {
+  /// \brief The function of each buffer
+  enum ArrowBufferType buffer_type[3];
+
+  /// \brief The size of an element each buffer or 0 if this size is variable or unknown
+  int64_t element_size_bits[3];
+
+  /// \brief The fixed size of a child element
+  int64_t child_size_elements;
```
lidavidm, line 192: How is this different from `element_size_bits`?

lidavidm (approved):

src/nanoarrow/typedefs_inline.h
paleolimbot: It's needed to calculate the length of a child of a fixed-size list (I should clarify + add a test for that, though, to make sure).

PR #24, opened 2022-08-19
Pass the ArrowBufferAllocator by value and implement a custom deallocator

paleolimbot:
This feature will allow R/Python runtimes to use the built-in `ArrowArrayInit()`/`ArrowArraySetBuffer()` infrastructure rather than roll their own (e.g., an R package could set a custom deallocator that calls `R_ReleaseObject()` on whatever object was protecting the buffer from garbage collection). Clients that are also using Arrow C++ could use this to with a shared pointer to a Buffer to avoid a copy as well.


lidavidm (approved):
This does bloat the ArrowBuffer structure a bit but I suppose that's  unavoidable

PR #25, opened 2022-08-19
Fix memcheck in CI (and memcheck errors in C)

### LINKED ISSUE #22: Fix memtest check to actually error on leak
Right now it's reporting errors but returning with exit code 0, so the CI still passes, plus the supplement file is not actually suppressing errors and there's no way to access the logs (i.e., we need to upload an artifact with those files if the build fails).

paleolimbot:
Fixes #22.


lidavidm (commented):
LGTM, though it seems there's still CI issues

lidavidm:
Ah…seems it's probably the suppressions not taking effect

PR #26, opened 2022-08-19
Implement getters

### LINKED ISSUE #18: Add element-wise getters?
This isn't strictly necessary - but I wonder what you'd think of having inline getters, things like

```c
int ArrowArrayIsSet(struct ArrowArray* array, int64_t i) {
  // TODO: unions
  if (array->n_buffers == 0) return 0;  // NA
  if (!array->buffers[0]) return 1;     // common optimization
  return ArrowBitGet((const uint8_t*) array->buffers[0], i);
}
```

and other counterparts for the 'easy' appenders. (possibly also written in terms of 'unsafe' type-width-specific getters, as with the appenders.)

paleolimbot:
Fixes #18.


src/nanoarrow/array_inline.h
```diff
@@ -445,6 +445,179 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+static inline int8_t ArrowArrayViewIsNull(struct ArrowArrayView* array_view, int64_t i) {
+  const uint8_t* validity_buffer = array_view->buffer_views[0].data.as_uint8;
+  i += array_view->array->offset;
+  switch (array_view->storage_type) {
+    case NANOARROW_TYPE_NA:
+      return 0x01;
+    case NANOARROW_TYPE_DENSE_UNION:
+    case NANOARROW_TYPE_SPARSE_UNION:
+      // Not supported yet
+      return 0xff;
```
lidavidm, line 457: Arguably it's technically "never null"

lidavidm (approved):

src/nanoarrow/array_inline.h
paleolimbot: Isn't the nullness determined by each sub-array? (i.e., the "right" way to do it would be `ArrowArrayViewIsNull(array_view->children[array_view->buffer_view[0].data.as_int8[i]], i)` for the sparse version)?

src/nanoarrow/array_view_test.cc
```diff
@@ -354,3 +360,101 @@ TEST(ArrayTest, ArrayViewTestFixedSizeListArray) {
...
+
+  EXPECT_EQ(ArrowArrayViewGetIntUnsafe(&array_view, 0), 1);
+  EXPECT_EQ(ArrowArrayViewGetUIntUnsafe(&array_view, 1), 2);
+
+  ArrowArrayViewReset(&array_view);
+  array.release(&array);
+  schema.release(&schema);
+}
+
+TEST(ArrayViewTest, ArrayViewTestGetNumeric) {
```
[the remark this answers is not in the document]
paleolimbot: There are a few tests with some repetition here, some avoidable, some not. I'm a poor candidate to do refactoring because I don't know what tools are available to help. I'm also not worried about the maintainability of the tests, though...for the most part they're all pretty valid things that a user of the library might type.

src/nanoarrow/array_inline.h
lidavidm: Right, but I'd say conceptually, the union array itself is always valid.

src/nanoarrow/array_view_test.cc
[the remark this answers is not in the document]
lidavidm: Yeah, the main 'benefit' is that it would expose each type as its own test case. But it's not critical.

src/nanoarrow/array_inline.h
lidavidm: Looking at C++, since type codes are buffer 1, then buffer 0 is nullptr and so the array will consider every slot valid

paleolimbot: I think I personally would have expected it to forward to whatever array was specified by the type ID (along with `ArrowArrayViewGetXXX()`). For somebody attempting to do a rowwise append of (e.g.) an integer column, I'd expect them to do something like:

```c
if (ArrowArrayViewIsNull(array_view, i)) {
  // whatever the rowwise format's version of "set this column as null" is
} else {
  int64_t value = ArrowArrayViewGetInt(array_view, i)
  // whatever the rowwise format's version of "set this column as int" is
}
```

If one treats every value of a union as non-null, users have to special case unions?

lidavidm: Fair - I guess the Arrow union array expects you to do that manually. In that case leaving this as a TODO is fine.

PR #31, opened 2022-08-22
Simplify file structure + add amalgamation option

### LINKED ISSUE #20: Consolidate implementation into files, or implement an amalgamation step?
In the spirit of making vendoring simpler for users, I'm wondering what we can do to reduce the total number of files in the project. In very large projects that are intending to be vendored (like DuckDB or SQLite), they implement an amalgamation that does a topological sort of dependencies and stitches together all the files to create one big header and one big compiled file. 

I see two different approaches here:

* Do all the implementation work in a single .C file (or a few .C files) and try to make it as brief as possible
* Implement an amalgamation similar to DuckDB or SQLite. 

If you look at https://github.com/nanopb/nanopb, there are 4 headers and 3 .C files. Do you think we could get closer to that?

paleolimbot:
Fixes #20.

- Reduces the number of source and header files (still maintaining some separation for sanity's sake while developing)
- Adds an amalgamation generator

Suggestions for how to make it easier to do are welcome! I'm not sure my CMake generator is the magical syntax we're collectively looking for but it is nice for testing.

```bash
git clone https://github.com/paleolimbot/arrow-nanoarrow.git && \
  cd arrow-nanoarrow && \
  git checkout  more-vendorable && \
  cmake . -DNANOARROW_AMALGAMATION=ON && \
  cmake --install . --prefix ..
```


lidavidm (approved):
It might be nice to just check in the amalgamated files, since that way the project would be a one-step download/clone-and-use?

lidavidm:
Commit somewhere, yes. A prerelease is also fine I think, my understanding is that we must just be careful not to refer to it as any sort of official release (so, it can't be in the docs, it must be clearly marked, etc.)

Issue #21, opened 2022-08-15
Demonstrate vendoring + symbol renaming


lidavidm:
Just to prove our claim of being easily embeddable, and to make sure that things work if multiple versions of nanoarrow end up in a binary (e.g. through two different dependencies)

paleolimbot:
I was thinking that an `examples/` directory should exist...this should definitely be one of them!

With respect to symbol renaming, I've only seen this in H3: https://github.com/uber/h3/blob/master/src/h3lib/include/h3api.h.in#L27-L50 . Should we do that here or is there another thing I should be using? Privately linking to a static library should also work, correct?

One could also use external function declarations as `NANOARROW_DLL void ArrowSomething();`, which might allow something like `#define NANOARROW_DLL static (inline?)` to get a header-only version? I don't want to go too crazy on the `#define` options because I find that kind of thing annoying when other libraries do it, but that one seems like it might allow some flexibility.

lidavidm:
jemalloc lets you do it as well (it's used in Arrow C++), though I'm not sure how it works there. (In that case, symbols are renamed, but still accessible.)

I think this is orthogonal to whether it's header-only, we can also use compiler/linker options (e.g. to give all nanoarrow symbols internal linkage).

PR #38, opened 2022-08-29
[C] Add namespacing option

### LINKED ISSUE #28: Allow namespacing
Since one of the selling points of nanoarrow is easier embedding and vendoring, we should probably make it possible to avoid conflicts between different nanoarrow versions loaded in the same process.

See for example a similar configuration option offered by xxhash:
https://github.com/Cyan4973/xxHash/blob/c4359b17db082888fdc18371eba918b957a6baaa/xxhash.h#L210-L225

paleolimbot:
Fixes #28 (will be documented in a PR for #21).

This is far outside my expertise...I chose an implementation between H3 ( https://github.com/uber/h3/blob/master/src/h3lib/include/h3api.h.in#L27-L50 ) and xxHash ( https://github.com/Cyan4973/xxHash/blob/c4359b17db082888fdc18371eba918b957a6baaa/xxhash.h#L210-L225 ) but am open to suggestions!


src/nanoarrow/nanoarrow.h
```diff
@@ -24,6 +24,81 @@
...
+// If using CMake, optionally pass -DNANOARROW_NAMESPACE=MyNamespace which will set this
+// define in build_id.h. If not, you can optionally #define NANOARROW_NAMESPACE
+// MyNamespace here.
+
+// This section remaps the non-prefixed symbols to the prefixed symbols so that
+// code written against this build can be used independent of the value of
+// NANOARROW_NAMESPACE.
+#ifdef NANOARROW_NAMESPACE
+#define NANOARROW_CAT(A, B) A##B
+#define NANOARROW_SYMBOL(A, B) NANOARROW_CAT(A, B)
```
pitrou, line 36: `#undef` these at the end?

.github/workflows/build-and-test.yaml
```diff
@@ -112,6 +113,16 @@ jobs:
...
+      
+      - name: Build nanoarrow (namespaced)
+        if: matrix.config.label == 'namespaced-build'
+        run: |
+          export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:`pwd`/dist/lib
+          sudo ldconfig
+          mkdir build
+          cd build
+          cmake .. -DCMAKE_BUILD_TYPE=Debug -DGTest_DIR=`pwd`/../dist/lib/cmake/GTest -DArrow_DIR=`pwd`/../dist/lib/cmake/arrow -DArrowTesting_DIR=`pwd`/../dist/lib/cmake/arrow -DNANOARROW_CODE_COVERAGE=ON -DNANOARROW_BUILD_TESTS=ON -DNANOARROW_NAMESPACE=SomeUserNamespace
+          cmake --build .
```
pitrou, line 125: Should you add a test step that inspects the produced library and checks that no symbol is non-namespaced?

paleolimbot: Done! (Also this highlighted a number of symbols I'd still defined by accident). Is there any check beyond the output of `nm --extern-only build/libnanoarrow.a` that I need to check?

PR #39, opened 2022-08-29
[R] Scaffold nanoarrow_schema, nanoarrow_array, and nanoarrow_array_stream S3 classes

paleolimbot:
This PR defines the nanoarrow_schema, nanoarrow_array, and nanoarrow_array_stream S3 classes. These objects are R external pointers wrapping a `struct ArrowSchema*`, `struct ArrowArray*`, and `struct ArrowArrayStream*`, respectively.


r/NAMESPACE
```diff
@@ -1,4 +1,33 @@
...
+S3method(as_nanoarrow_array,RecordBatch)
+S3method(as_nanoarrow_array,Table)
+S3method(as_nanoarrow_array,default)
+S3method(as_nanoarrow_array_stream,RecordBatchReader)
+S3method(as_nanoarrow_array_stream,default)
+S3method(as_nanoarrow_array_stream,nanoarrow_array_stream)
+S3method(as_nanoarrow_schema,DataType)
+S3method(as_nanoarrow_schema,Field)
+S3method(as_nanoarrow_schema,Schema)
+S3method(as_nanoarrow_schema,nanoarrow_schema)
```
nealrichardson, line 14: My first reaction here is that `as_nanoarrow_array()` feels like a lot to type

r/R/pkg-arrow.R
```diff
@@ -0,0 +1,168 @@
...
+
+# exported in zzz.R
+infer_type.nanoarrow_array <- function(x, ...) {
+  arrow::as_data_type(infer_nanoarrow_schema(x, ...))
+}
+
+as_data_type.nanoarrow_schema <- function(x, ...) {
+  exportable_schema <- nanoarrow_allocate_schema()
+  nanoarrow_pointer_export(x, exportable_schema)
+  asNamespace("arrow")$DataType$import_from_c(exportable_schema)
```
nealrichardson, line 26: Not exported from arrow? We should fix that. 

```suggestion
  getFromNamespace("DataType", "arrow")$import_from_c(exportable_schema)
```

r/R/pkg-arrow.R
```diff
@@ -0,0 +1,168 @@
...
+  nanoarrow_array_set_schema(array, imported_schema)
+  array
+}
+
+#' @export
+as_nanoarrow_array.ChunkedArray <- function(x, ..., schema = NULL) {
+  if (is.null(schema)) {
+    array <- arrow::as_arrow_array(x)
+  } else {
+    array <- arrow::as_arrow_array(x, type = arrow::as_data_type(schema))
```
nealrichardson, line 130: This concatenates the chunks, right? Is that what we want to do? That's not what we do when passing R <-> Python in `arrow`.

r/R/pkg-arrow.R
```diff
@@ -0,0 +1,168 @@
...
+  } else {
+    batch <- arrow::as_record_batch(x, schema = arrow::as_schema(schema))
+  }
+
+  as_nanoarrow_array.RecordBatch(batch)
+}
+
+#' @export
+as_nanoarrow_array_stream.RecordBatchReader <- function(x, ..., schema = NULL) {
+  stopifnot(is.null(schema))
```
nealrichardson, line 164: this is a TODO?

r/R/zzz.R
```diff
@@ -0,0 +1,88 @@
...
+  s3_register("arrow::as_data_type", "nanoarrow_schema")
+  s3_register("arrow::as_schema", "nanoarrow_schema")
+  s3_register("arrow::as_arrow_array", "nanoarrow_array")
+  s3_register("arrow::as_chunked_array", "nanoarrow_array")
+  s3_register("arrow::as_record_batch", "nanoarrow_array")
+  s3_register("arrow::as_arrow_table", "nanoarrow_array")
+  s3_register("arrow::as_record_batch_reader", "nanoarrow_array_stream")
+}
+
+s3_register <- function(generic, class, method = NULL) {
```
nealrichardson, line 30: This is copied from another package, yeah? Should provide attribution.

r/R/array.R
```diff
@@ -0,0 +1,68 @@
...
+#' @rdname as_nanoarrow_array
+#' @export
+from_nanoarrow_array <- function(array, to = NULL, ...) {
+  stopifnot(inherits(array, "nanoarrow_array"))
+  UseMethod("from_nanoarrow_array", to)
+}
+
+#' @export
+as_nanoarrow_array.default <- function(x, ..., schema = NULL) {
+  # For now, use arrow's conversion for everything
```
nealrichardson, line 43: Most of the `as` functions call `arrow::something()`, is that temporary? What does this package do currently that doesn't require `arrow`?

r/DESCRIPTION
```diff
@@ -22,5 +22,6 @@ RoxygenNote: 7.2.1
 URL: https://github.com/apache/arrow-nanoarrow
 BugReports: https://github.com/apache/arrow-nanoarrow/issues
 Suggests: 
+    arrow (>= 8.0.0),
```
nealrichardson, line 25: 8.0.0 is for the `as_*` methods?

r/NAMESPACE
paleolimbot: Yeah, `as_nanoarrow_anything()` is a mouthful and is mostly because I couldn't think of anything better. I suppose the namespace `na` instead of `nanoarrow` could work, too, but but it's also mostly developer facing and users shouldn't really have to know these classes/functions exist.

The reason why I didn't use `to_nanoarrow()` is that each S3 class really does need its own coerce method. I used them in geoarrow to do stuff like:

```r
do_something_cool <- function(array) {
  array <- as_nanoarrow_array(array)
  .Call(some_c_function, array)
}
```

(i.e., a similar way that we use `arrow::as_arrow_array()` although the nanoarrow array doesn't have any user-facing conveniences like `[` subsetting or other vector-ish S3 generics).

The `as_nanoarrow_array()` / `from_nanoarrow_array()` bit is inspired by `vctrs::vec_restore()` and `vctrs::vec_cast()`, which both take a ptype `to` argument. In arrow we're user-facing and we don't let the user pick what type of object gets returned from `as.vector()`. In nanoarrow we're developer-facing, and the ability to do something like `from_nanoarrow_array(array, double())` (or recursively, `from_nanoarrow_array(array, data.frame(x = double())`) makes R type stability easier. readr does something like this, too with its `cols(x = col_double())`. Importantly, this also lets an extension package author define the `as_nanoarrow_array()` and `from_nanoarrow_array()` S3 generics in their own package and nanoarrow doesn't have to know it exists for `from_nanoarrow_array(array, data.frame(x = some_extension_type())` to do its thing.

All of that is a bit esoteric, though, because at the end of the day `as.data.frame()` will be the entry point for most people most times. The main thing I like about the `as_()`/`from_()` system is that the S3 object author decides how or if that conversion takes place, not nanoarrow.

r/R/array.R
paleolimbot: Yes, it's temporary. When did narrow the first time, I started from the "let's assemble C structs from R vectors first" side of things and it was a long road to anything useful. Here I'm trying it from the other side...let's let Arrow create valid C structs for us and implement conversions one at a time.

r/DESCRIPTION
paleolimbot: Yes, and the ability to pass an external pointer directly into `import_from_c`/`export_from_c`. With a tiny bit of work I could probably relax that to whenever the C interface was implemented in the arrow package.

r/R/pkg-arrow.R
paleolimbot: (I added todo notes)

r/R/zzz.R
paleolimbot: I added a note here linking to the comment in vctrs where it says you can do this!

nealrichardson: This may require a note in a LICENSE file, as we do in apache/arrow. Fortunately it is [permissively licensed (MIT)](https://github.com/r-lib/vctrs/blob/main/DESCRIPTION#L26).

paleolimbot: I'm certainly happy to do that to be safe, although this particular function seems like it has been licensed differently to support this sort of usage:

```
#' @section Usage in other packages:
#' To avoid taking a dependency on vctrs, you copy the source of
#' [`s3_register()`](https://github.com/r-lib/vctrs/blob/main/R/register-s3.R)
#' into your own package. It is licensed under the permissive
#' [unlicense](https://choosealicense.com/licenses/unlicense/) to make it
#' crystal clear that we're happy for you to do this. There's no need to include
#' the license or even credit us when using this function.
```

( https://github.com/r-lib/vctrs/blob/c2a7710fe55e3a2249c4fdfe75bbccbafcf38804/R/register-s3.R#L25-L31 )

r/NAMESPACE
paleolimbot: I added `as.vector()` and `as.data.frame()` implementations and will use those in the tests (so that it's easy to remove the `from_arrow_array()` system if it does in fact never get used)!

r/R/zzz.R
nealrichardson: Oh interesting

PR #40, opened 2022-09-01
Add example usage to examples/ directory

### LINKED ISSUE #21: Demonstrate vendoring + symbol renaming

paleolimbot:
Fixes #21. Will demonstrate:

- A minimal CMake configuration that builds + privately links a static nanoarrow build
- A minimal vendored configuration

The principles I'd like to communicate are:

- You must vendor nanoarrow.c/nanoarrow.h or privately link a static library
- You must not expose any parts of nanoarrow.h in one of your library's headers (except the ABI stable Arrow C Data/Stream interface)
- When using the vendored nanoarrow.h/nanoarrow.c version, you should `#define NANOARROW_NAMESPACE SomeNamespace` to avoid symbol collisions (i.e., if somebody else loads your library into the same process or vendors it into another library)
- Maybe bits I'm missing?

### DIFF
--- a/examples/vendored-minimal/README.md
+++ b/examples/vendored-minimal/README.md
@@ -1,0 +1,40 @@
+<!---
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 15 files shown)


examples/vendored-minimal/README.md
```diff
@@ -0,0 +1,76 @@
...
+parses command line arguments into an int32 array and prints out the
+resulting length (or any error encountered whilst building the array).
+
+```bash
+./example_vendored_minimal_app 8 2 4 9
+#> Parsed array with length 4
+./example_vendored_minimal_app asd
+#> Can't parse argument 1 ('asd') to long int
+./example_vendored_minimal_app 999999999999999
+#> Error appending argument 1 ('999999999999999') to array
```
lidavidm, line 75: I might be missing something, but it doesn't look like the code actually implements this?

lidavidm (approved):
LGTM!

examples/vendored-minimal/README.md
paleolimbot: Thanks! That was a good catch (it was the output of the version before I simplified it a bit).

PR #41, opened 2022-09-06
[C] Add layout for Map type

lidavidm:


lidavidm:
Oops, this is what I get for not trying it locally. I'll fix that!

lidavidm:
Ah, we need a local build of Arrow to use testing…I wonder if we could avoid ArrayFromJSON somehow and then we could build against the Arrow conda package or something like that

lidavidm:
Fixed.

(Incidentally, it looks like we have a .clang-format, but no .pre-commit-config or CI pipeline?)

lidavidm:
Filed #42 for whenever somebody can get to it (though it isn't really a big deal)

paleolimbot (approved):

PR #44, opened 2022-09-12
[C] Fix some compiler warnings

lidavidm:


lidavidm:
Except for the switch-case fallthrough, the warnings I found were all mostly nits. So maybe we don't care about them. If we think it's useful, I'll also include the CMake config I used and fix the remainder of the clang lints (there are a bunch of places where it wants explicit casts), and add a test for the fallthrough, and we can iterate from there?

lidavidm:
Updated, though maybe we should consider whether some of these casts should bounds-check instead

src/nanoarrow/schema.c
```diff
@@ -591,7 +591,7 @@ static ArrowErrorCode ArrowSchemaViewParse(struct ArrowSchemaView* schema_view,
         return EINVAL;
       }
 
-      schema_view->fixed_size = strtol(format + 2, (char**)format_end_out, 10);
+      schema_view->fixed_size = (int32_t)strtol(format + 2, (char**)format_end_out, 10);
```
paleolimbot, line 594: I'm not too worried about this, but you could also change `struct SchemaView::fixed_size` to be an `int64_t`.

src/nanoarrow/schema.c
```diff
@@ -1026,7 +1032,7 @@ ArrowErrorCode ArrowSchemaViewInit(struct ArrowSchemaView* schema_view,
     return EINVAL;
   }
 
-  int format_len = strlen(format);
+  unsigned long format_len = strlen(format);
```
paleolimbot, line 1035: Maybe `size_t`? I'm sure `unsigned long` is fine, we just don't use it anywhere else.

CMakeLists.txt
```diff
@@ -89,6 +89,29 @@ else()
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
         $<INSTALL_INTERFACE:include>)
     target_include_directories(nanoarrow PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/generated>)
+    if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
```
paleolimbot, line 92: Should this be fenced in a `if(CMAKE_BUILD_TYPE EQUAL "Debug")`? If somebody is building this as a dependency of a dependency or something it seems a bit rude to foist warnings on them (or maybe this doesn't do that or maybe that's common?)

CMakeLists.txt
```diff
@@ -89,6 +89,29 @@ else()
...
+        PRIVATE
+        -Wall
+        -Werror
+        -Wextra
+        -Wno-type-limits
+        -Wno-unused-parameter
+        -Wpedantic
+        -Wunused-result)
+    elseif(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR
+        CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
paleolimbot, line 104: Above this is `CMAKE_C_COMPILER_ID` (as opposed to `CXX`)...does that matter?

paleolimbot (approved):
Thanks! It's great as is...just a few questions.

CMakeLists.txt
lidavidm: Derp, I copy-pasted from Arrow…this should be C indeed

PR #56, opened 2022-09-29
[C] Don't use ArrowTesting

### LINKED ISSUE #42: [C] Replace ArrayFromJSON/ArrowTesting to simplify build setup
Re: https://github.com/apache/arrow-nanoarrow/issues/41#issuecomment-1238459772

If we could avoid ArrayFromJSON and libarrow_testing, then we could build the tests using a conda-forge (or other package manager) build of Arrow, avoiding a local build.

paleolimbot:
Fixes #42 by replacing `ArrayFromJSON` with `ArrayBuilder`. This has the benefit of being able to use pre-built binaries and simplifies the development build considerably (otherwise, we have to link to the same GTest and deal with conflicts between a system Arrow and a dev arrow). This also simplifies the CI quite a bit!


lidavidm (approved):
A helper function in the style of Arrow's ArrayFromVector might be good to consider to cut down on the boilerplate a bit, but this looks good!

PR #57, opened 2022-09-30
[C][R] Stringify schemas

### LINKED ISSUE #46: [C] Add ArrowSchemaToString()
In order to write comprehensible error messages (for example, https://github.com/paleolimbot/minigpkg/blob/main/src/minigpkg/nanoarrow_sqlite3.c#L272-L277), it would be nice to be able to stringify an `ArrowSchema`.

I think this would be most useful if it printed recursively since that's mostly what a user is going to want to see in an error message, but could be convinced otherwise. Maybe:

```c
int ArrowSchemaToString(struct ArrowSchema* schema, char* buf, int64_t buf_size);
```

(i.e., it would have `snprintf()` ish semantics and doesn't allocate)

paleolimbot:
Fixes #46. These strings are mostly the C++ representation with a few generalizations to make it easier to do this from C. In general, everything looks like `name(parameters)<children>`, with some special cases for dictionaries and extensions (or both!).

The entrypoint in C is intended to be like `snprintf()`: `int64_t ArrowSchemaFormat(struct ArrowSchema* schema, char* out, int64_t n, char recursive)`. Like snprintf, you can pass `nullptr, 0` to get the length of the output if you want to make sure not to truncate the output. The implementation is a bit of a mess because that interface is hard to generalize...perhaps there's a pattern for doing this kind of thing that I don't know about.

 The actual output is best illustrated from the R bindings:

``` r
library(nanoarrow)
library(arrow, warn.conflicts = FALSE)
#> Some features are not enabled in this build of Arrow. Run `arrow_info()` for more information.

print_schema <- function(x, recursive = TRUE) {
  x <- as_nanoarrow_schema(x)
  cat(nanoarrow:::nanoarrow_schema_formatted(x, recursive))
  cat("\n")
}

print_schema(timestamp(timezone = "America/Halifax"))
#> timestamp('s', 'America/Halifax')
print_schema(struct(col1 = int32(), col2 = float64()))
#> struct<col1: int32, col2: double>
print_schema(map_of(int32(), float64()))
#> map<entries: struct<key: int32, value: double>>
print_schema(dictionary())
#> dictionary(int32)<string>
print_schema(new_extension_type(int32(), "some.ext.name"))
#> some.ext.name<int32>
print_schema(new_extension_type(dictionary(), "some.ext.name"))
#> some.ext.name<dictionary(int32)<string>>
print_schema(dictionary(value_type = new_extension_type(int32(), "some.ext.name")))
#> dictionary(int32)<some.ext.name<int32>>
```

<sup>Created on 2022-09-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


src/nanoarrow/nanoarrow.h
```diff
@@ -210,6 +200,8 @@ enum ArrowTimeUnit {
 /// NANOARROW_OK is returned.
 ArrowErrorCode ArrowSchemaInit(struct ArrowSchema* schema, enum ArrowType type);
 
+int64_t ArrowSchemaFormat(struct ArrowSchema* schema, char* out, int64_t n, char recursive);
```
lidavidm, line 203: nit: docstring?

src/nanoarrow/schema.c
```diff
@@ -1088,6 +1088,149 @@ ArrowErrorCode ArrowSchemaViewInit(struct ArrowSchemaView* schema_view,
...
+  }
+
+  if (schema->release == NULL) {
+    return snprintf(out, n, "[invalid: schema is released]");
+  }
+
+  struct ArrowSchemaView schema_view;
+  struct ArrowError error;
+
+  if (ArrowSchemaViewInit(&schema_view, schema, &error) != NANOARROW_OK) {
```
lidavidm, line 1135: I wonder if the function should just take an ArrowSchemaView instead then?

wjones127:
The only thing I would consider changing is the wrapper character for physical type for extensions (which IIUC isn't a type parameter and isn't quite child data):
```r
print_schema(new_extension_type(int32(), "some.ext.name"))
#> some.ext.name{int32}
print_schema(new_extension_type(dictionary(), "some.ext.name"))
#> some.ext.name{dictionary(int32)<string>}
```

Everything else looks pretty good!

src/nanoarrow/schema.c
paleolimbot: For argument's sake, I pushed a version that leans the other direction and just returns a `char*`. I don't want to make it too hard to do this (but also don't want to allocate unnecessarily if you think this is going be a problem). The `char*` version is nice because you can also use it from the debugger console and use it a bit more easily in other error messages (i.e., `ArrowErrorSet("Can't cast array of type %s to %s, formatted1, formatted2)`).

paleolimbot: So, I changed it back 😬 ...it turns out it takes *more* code to do it the other way. Happy to update this if it turns out to be super annoying! (Or to expose a SchemaView version if that's helpful later on)

PR #61, opened 2022-10-21
Proof-of-concept IPC reader/writer

paleolimbot:
Just a public experiment for a possible IPC reader/writer. The general idea is:

- Use only the Arrow C Data and C Stream interface in the public header
- Use flatcc to handle flatbuffers IO ( https://github.com/dvidelabs/flatcc )
- Use nanoarrow to construct/read the ArrowSchema and ArrowArray objects to/from flatbuffers

I put this in an extensions/ subdirectory in this repo. I think that's a good fit...the IPC format is part of the spec, after all, and putting it in this repo would mean that the R and/or Python package could make use of it.

Any thoughts welcome!


lidavidm:
Just a quick thought, but I wonder if we might even want to go as far as vendoring the flatcc runtime in addition to checking in the generated sources

lidavidm:
That way, this would be an entirely self-contained build unless you wanted to update flatcc/the flatbuffer definitions

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.h
```diff
@@ -0,0 +1,153 @@
...
+#endif  // ARROW_C_STREAM_INTERFACE
+#endif  // ARROW_FLAG_DICTIONARY_ORDERED
+
+#ifdef __cplusplus
+}
+#endif
+
+typedef int ArrowIpcErrorCode;
+
+struct ArrowIpcIO {
```
lidavidm, line 131: An alternative might be a push-based API, like [StreamDecoder](https://github.com/apache/arrow/blob/4ce64537e5ee223bf5eea91284efc46f214d6c76/cpp/src/arrow/ipc/reader.h#L309)? (Except maybe instead of a callback, require the user to poll the reader object)

paleolimbot: That could be implemented using the pull-based stuff I have here, no? (Most file APIs I know about are pull based, which makes that a good target for the initial implementation).

paleolimbot (commented):
Agreed...it would be nice to make it self-contained (maybe even amalgamated and checked-in to `dist/nanoarrow_ipc.h+c`). Although maybe then we have to worry about symbol conflicts for flatcc which might be hard to solve?

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.h
lidavidm: It would be more natural for reading a file, but for something pulling in from (say) an HTTP stream or network socket, it would probably be more natural to push data instead. But yes, either can be built from the other.

lidavidm:
We could perhaps keep the flatcc runtime separate, so that you would just not include it if you're also using flatcc yourself, but otherwise it's right there and you don't need to fetch+build a third party repo (that said we should probably see how much work it is to do that in the first place, maybe it's not a big deal)

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.c
```diff
@@ -0,0 +1,297 @@
...
+        break;
+      }
+      case ns(Type_Utf8):
+        result = ArrowSchemaSetType(schema, NANOARROW_TYPE_STRING);
+      case ns(Type_LargeUtf8):
+        result = ArrowSchemaSetType(schema, NANOARROW_TYPE_LARGE_STRING);
+        break;
+      default:
+        ArrowIpcErrorSet(error, "Unrecognized Field type with value %d", (int)type_type);
+        return EINVAL;
```
lidavidm, line 203: nit, but is there a reason to mix early-return and single-return error handling here?

lidavidm (commented):
Just took a look through since I saw the updates

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.c
paleolimbot: Not at all! I'd accumulated too much in one function and hadn't split them yet 🙂

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  data.data = kSimpleSchema;
+  data.size_bytes = 1;
+
+  ArrowIpcReaderInit(&reader);
+
+  EXPECT_EQ(ArrowIpcReaderVerify(&reader, &data, &error), EINVAL);
+  EXPECT_STREQ(error.message,
+               "Expected data of at least 8 bytes but only 1 bytes remain");
+  EXPECT_EQ(data.data, kSimpleSchema);
+  EXPECT_EQ(data.size_bytes, 1);
```
lidavidm, line 73: Declaring the pointers as `const` in the API signatures would let you make this assurance statically

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc_test.cc
```diff
@@ -0,0 +1,171 @@
...
+    0x08, 0x00, 0x00, 0x00, 0x73, 0x6f, 0x6d, 0x65, 0x5f, 0x63, 0x6f, 0x6c, 0x00, 0x00,
+    0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x0c, 0x00, 0x00, 0x00, 0x08, 0x00, 0x0c, 0x00,
+    0x04, 0x00, 0x08, 0x00, 0x08, 0x00, 0x00, 0x00, 0x20, 0x00, 0x00, 0x00, 0x04, 0x00,
+    0x00, 0x00, 0x10, 0x00, 0x00, 0x00, 0x73, 0x6f, 0x6d, 0x65, 0x5f, 0x76, 0x61, 0x6c,
+    0x75, 0x65, 0x5f, 0x66, 0x69, 0x65, 0x6c, 0x64, 0x00, 0x00, 0x00, 0x00, 0x0e, 0x00,
+    0x00, 0x00, 0x73, 0x6f, 0x6d, 0x65, 0x5f, 0x6b, 0x65, 0x79, 0x5f, 0x66, 0x69, 0x65,
+    0x6c, 0x64, 0x00, 0x00, 0x08, 0x00, 0x0c, 0x00, 0x08, 0x00, 0x07, 0x00, 0x08, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x20, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+
+TEST(NanoarrowIpcTest, NanoarrowIpcCheckHeader) {
```
lidavidm, line 59: Sometime down the line, these are the sort of tests where a fuzzer would be useful

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc_test.cc
lidavidm: Ah, though I see down below that it does mutate the buffer in some cases, so never mind.

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.c
```diff
@@ -0,0 +1,352 @@
...
+#include "nanoarrow/nanoarrow.h"
+
+#include "File_reader.h"
+#include "Message_reader.h"
+#include "Message_verifier.h"
+#include "Schema_reader.h"
+
+#include "nanoarrow_ipc.h"
+
+#define ArrowIpcErrorSet(err, ...) ArrowErrorSet((struct ArrowError*)err, __VA_ARGS__)
```
lidavidm, line 30: Hmm, is this actually useful? (Also technically, I think you're not supposed to pun between two structs just because they happen to have the same fields...)

What do you think about having nanoarrow_ipc.h just include nanoarrow.h, and just reuse existing utilities?

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.c
```diff
@@ -0,0 +1,352 @@
...
+    reader->schema.release(&reader->schema);
+  }
+
+  ArrowIpcReaderInit(reader);
+}
+
+static inline uint32_t ArrowIpcReadUint32LE(struct ArrowIpcBufferView* data) {
+  uint32_t value;
+  memcpy(&value, data->data, sizeof(uint32_t));
+  // bswap32() if big endian
```
lidavidm, line 47: did you mean to put a TODO here?

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.c
```diff
@@ -0,0 +1,352 @@
...
+static inline int32_t ArrowIpcReadInt32LE(struct ArrowIpcBufferView* data) {
+  int32_t value;
+  memcpy(&value, data->data, sizeof(int32_t));
+  // bswap32() if big endian
+  data->data += sizeof(int32_t);
+  data->size_bytes -= sizeof(int32_t);
+  return value;
+}
+
+#undef ns
```
lidavidm, line 62: hmm, where did the original declaration come from?

lidavidm (commented):
Generally looks good! Just a few thoughts

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc_test.cc
paleolimbot: Yeah, I was thinking one could iterate through like

```c
int result;
while (result = ArrowIpcReaderVerify(&reader, &data, &error)) {
  // ...
}
```

(but maybe that's more confusing than useful)

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.c
paleolimbot: It's a good question...probably the advantage of having it live in `arrow-nanoarrow/extensions` is that it can be distributed alongside nanoarrow.c/nanoarrow.h (i.e., all extensions get checked in to arrow-nanoarrow/dist just like the core C library). I did it this way to start off because that's how I would reccomend that another developer build a library on top of nanoarrow (vendor nanoarrow and don't expose nanoarrow.h to your user because they might be vendoring another version of it themselves).

lidavidm: Ah, I guess the question is, is this a separately library using nanoarrow internally, or is it just an optional part of nanoarrow?

I don't mind this not exposing nanoarrow. But this is technically undefined behavior here, as I understand.

paleolimbot: I copied over nanoarrow's set method for now but the next PR should decide whether or not "extensions" can/should `#include "nanoarrow.h"` in their public header. It's a lot easier if they can; however, I think there was some email thread where somebody said it was a bad idea.

extensions/nanoarrow_ipc/src/nanoarrow_ipc/nanoarrow_ipc.c
paleolimbot: I think I copied it from the documentation (but it's confusing and gone now!)

lidavidm:
Sorry, forgot to reply here. I think let's merge this and we can continue work in follow up PRs. I think the main question to resolve is whether this should `#include <nanoarrow/nanoarrow.h>` or not.

Issue #53, opened 2022-09-28
[Python] Useful scope of the python bindings?


jorisvandenbossche:
I opened https://github.com/apache/arrow-nanoarrow/pull/52 as a start for Python bindings for nanoarrow, currently just a package scaffold. 

I am planning to subsequently add basic introspection/consumption of ArrowArray structs and basic conversions of buffers/arrays to numpy. 
Some more elaborate notes / thoughts can be found https://docs.google.com/document/d/1119poLwF0r4AN19dGt9U8vLM07zEVdlRpP2omL1PXdc/edit?usp=sharing

More in general, it would be interesting to discuss or get feedback on what functionality would be useful / needed for other potential use cases.

paleolimbot:
This is awesome!

The scope of the Python bindings certainly doesn't have to be the same as the scope of the R bindings; however I'll offer up what I had in mind for the scope of the R bindings in case that helps with the conversation. Basically:

- Data structure: provide an R class that holds an Array/Schema/ArrayStream and releases it when it goes out of scope. In Python you have duck typing so it matters less, but in R it helps to be able to write `array <- as_nanoarrow_array(some_object)` and have the `as_nanoarrow_array()` S3 generic live in the nanoarrow package.
- Introspection: for debugging the R bindings themselves it's really nice to be able to "see" the Array/Schema/ArrayStream fields. This helps write tests, too, particularly when constructing Arrays from buffers.
- Conversion from Array to R vector: ADBC makes database results available via a `struct ArrowArrayStream` (and so does the next GDAL release): including conversions in nanoarrow means that these results are accessible without an Arrow C++ dependency.
- Creating ArrowArray objects from buffers: This makes it so that you can make an array for any arrow type, regardless if R arrow or R nanoarrow has implemented a conversion for it. I used this (and still use this) to prototype GeoArrow extension type representations since nested list support in R/arrow is not great and extension type support was non-existent when I started.
- Creating ArrowArray objects from R vectors: In R this is pretty straightforward since we only have a few vector types, most of which we can zero-copy convert into Arrow arrays. ADBC lets you do database inserts using a `struct ArrowArrayStream`, so being able to zero-copy R vectors into that form allows database connectivity without an Arrow C++ dependency.
- Reading ArrayStreams: This is pretty easy as long as there are data structures for schemas and arrays
- Creating ArrowArrayStreams: At the very least, from a list of arrays to facilitate testing APIs that use them

[...]

PR #62, opened 2022-10-24
[Python] Basic Array class wrapping C struct (with conversion to numpy)

jorisvandenbossche:
xref https://github.com/apache/arrow-nanoarrow/issues/53

This further explores possible Python bindings in cython. I added here an `Array` class that wraps an `ArrowArray*` and `ArrowSchema*` struct with some basic features: 1) ensuring to `release` the pointers when Array objects get deallocated, 2) method to construct it from raw pointers or from a pyarrow array (that last one is probably mostly useful for testing), 3) basic conversion to numpy which keeps this Array object as its base to keep it alive.

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -19,7 +19,10 @@
 
 """Low-level nanoarrow Python bindings."""
 
-from libc.stdint cimport uint8_t, uintptr_t
+from libc.stdlib cimport malloc, free
+from libc.stdint cimport int64_t, uint8_t, uintptr_t
+from cython.operator cimport dereference as deref
+from cpython cimport PyObject, Py_INCREF
 
 from nanoarrow_c cimport *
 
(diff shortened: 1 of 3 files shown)


python/src/nanoarrow/_lib.pyx
```diff
@@ -44,14 +47,82 @@ cdef dict _numpy_type_map = {
 }
 
 
-def as_numpy_array(arr):
-    cdef ArrowSchema schema
-    cdef ArrowArray array
+cdef class Array:
```
paleolimbot, line 50: Semantics, but what you have here is more like a wrapper around the `ArrowArrayView` (that keeps the objects it points to alive). The R package has one of these too (and similarly it backs all the conversions to R vectors).

python/src/nanoarrow/_lib.pyx
```diff
@@ -44,14 +47,82 @@ cdef dict _numpy_type_map = {
...
+                free(self.array_ptr)
+                self.array_ptr = NULL
+            if self.schema_ptr is not NULL:
+                free(self.schema_ptr)
+                self.schema_ptr = NULL
+        self.parent = None
+
+    @property
+    def format(self):
+        cdef const char* format_string = deref(self.schema_ptr).format
```
paleolimbot, line 83: With the latest nanoarrow you can do:

https://github.com/apache/arrow-nanoarrow/blob/7395bf25275679f6a4fbc35c7cf9bc8fbd6b6150/r/src/schema.c#L146-L149

python/tests/test_nanoarrow.py
```diff
@@ -6,22 +8,64 @@
...
+    ptr_array = int(ffi.cast("uintptr_t", c_array))
+
+    typ = pa.int32()
+    parr = pa.array([1, 2, 3], type=typ)
+    parr._export_to_c(ptr_array, ptr_schema)
+
+    arr = nanoarrow.Array.from_pointers(ptr_array, ptr_schema)
+    assert arr.to_numpy().tolist() == [1, 2, 3]
+
+    # trying to import second time should not cause a segfault? To enable
```
paleolimbot, line 69: My reading of cython code is not awesome, but I think importing twice causes a double free when the `release()` method to be called twice?

python/src/nanoarrow/_lib.pyx
```diff
@@ -44,14 +47,82 @@ cdef dict _numpy_type_map = {
...
+            ArrowSchema* c_schema_ptr = <ArrowSchema*> <uintptr_t > schema_ptr
+        return Array._from_ptrs(c_array_ptr, c_schema_ptr)
+
+    @classmethod
+    def from_pyarrow(cls, arr):
+        cdef ArrowSchema *schema = <ArrowSchema *>malloc(sizeof(ArrowSchema))
+        cdef ArrowArray *array = <ArrowArray *>malloc(sizeof(ArrowArray))
+        arr._export_to_c(<uintptr_t> array, <uintptr_t> schema)
+        self = Array._from_ptrs(array, schema, own_ptrs=True)
+        self.parent = arr
```
paleolimbot, line 109: You probably don't need this? (I think the release callback takes care of keeping the `shared_ptr` alive for the lifecycle of the schema/array)

paleolimbot (commented):
This reminds me how much I wish R had cython. This is some of the first cython code I've read so I might not be much help.

Just spitballing, maybe what you could have is:

```cython
cdef class Array:
    cdef:
        ArrowArray array
        object parent
    # make sure to set array.release = nullptr on __new__ and
    # call it if non-null on __dealloc__

cdef class Schema:
    cdef:
        ArrowSchema schema
        object parent

cdef class ArrayView:
    cdef:
        Array array # can you do this in cython?
        Schema schema
        ArrowArrayView array_view
        # call ArrowArrayViewInit() on __new__ and ArrowArrayViewReset() on __dealloc__
```

...that might simplify the `own_data`/`own_ptrs`/malloc lifecycle stuff you have here.

jorisvandenbossche:
Thanks for the comments!

> Just spitballing, maybe what you could have is: ... that might simplify the `own_data`/`own_ptrs`/malloc lifecycle stuff you have here.

The question is also to what extent we want to expose the Array / ArrayView difference to Python users. For nanoarrow C, this distinction makes total sense, as we want something on top of the raw ArrowArray struct, which we can't change. But in Python, the Array class is already something that wraps the raw struct, so I was thinking that therefore the Python API doesn't need to directly expose this distinction to the user.


The `own_data` isn't actually used yet. I only added it (prematurely) because I was thinking that at some point the struct might be passed to something else (eg to pyarrow), and the nanoarrow.Array class shouldn't call the release callback. Although that can (should?) probably be done by copying the struct before passing it to another consumer and setting `array.release = NULL` (https://arrow.apache.org/docs/dev/format/CDataInterface.html#moving-an-array) and in `__dealloc__` check for the release callback not being NULL whether to call it or not.

python/tests/test_nanoarrow.py
jorisvandenbossche: Yes, I think that's correct. Right now, with the raw pointers, it's the responsibility of the user passing around those pointers that they don't consume the pointers twice (as the consumer cannot know that there is another consumer of the same struct). 

I think using PyCapsules instead of raw pointers could make this more robust. 
But in practice to avoid this issue, I suppose we should "move" the array struct when constructing from external raw pointers, so the original pointer can be marked as released (release callback set to NULL)

PR #74, opened 2022-11-25
[C] Fix some MAP type support issues

lidavidm:
Generally, make sure MAP is consistent with LIST:
- MAP has 2 buffers (validity + offset)
- MAP must be finished like LIST (we could go further and have FinishElement also finish the child in the case of MAP, since we know it must always be STRUCT)

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -648,10 +648,12 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
       }
       break;
     case NANOARROW_TYPE_LIST:
-    case NANOARROW_TYPE_MAP:
+    case NANOARROW_TYPE_MAP: {
+      const char* type_name =
+          array_view->storage_type == NANOARROW_TYPE_LIST ? "list" : "map";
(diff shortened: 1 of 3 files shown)


lidavidm:
Side note: how do you run clang-format? I don't see it in CI and I don't see a pre-commit

paleolimbot (approved):
A few comments...it may be worth following this up with a section on the Map in https://arrow.apache.org/docs/format/Columnar.html ...other than a footnote in the C Data interface spec, I can't find any reference to the memory layout for a Map (or for the datetime interval types, which I remember having to look up in the C++ implementation)

lidavidm:
Map is a "logical type" so that doc defers an explanation to Schema.fbs which has 

```
/// A Map is a logical nested type that is represented as
///
/// List<entries: Struct<key: K, value: V>>
```

Maybe we should explicitly list out the logical types so you can control+F easily though?

lidavidm:
No, the list sizes are variable. Each list item represents a dictionary, where the Struct values of each list entry are key-value pairs.

lidavidm:
Trying this new issue flow out: https://github.com/apache/arrow/issues/14752

PR #77, opened 2022-11-28
[C] Fix AppendDouble to accept all values on float arrays

### LINKED ISSUE #75: [C] AppendDouble doesn't work with most floats
AppendDouble checks that the value is in the range `[FLT_MIN, FLT_MAX]` before appending. However, that's not actually the right check: `FLT_MIN` isn't the "minimum float value" (that doesn't make sense; floats contain infinity), but rather the smallest positive float. Hence, 0 and any negative float will get wrongly rejected.

IMO, it might make sense to just accept any double and cast it to float without thinking. Otherwise, maybe the right check is `(double)((float) v) == v` or something ("is this double exactly representable as a float", perhaps there's an idiom for this though I can't get the right search query to figure this out), though this will not work for NaN.

lidavidm:
Fixes #75.

### DIFF
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -303,10 +303,9 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
     case NANOARROW_TYPE_DOUBLE:
       NANOARROW_RETURN_NOT_OK(ArrowBufferAppend(data_buffer, &value, sizeof(double)));
       break;
-    case NANOARROW_TYPE_FLOAT: {
+    case NANOARROW_TYPE_FLOAT:
       NANOARROW_RETURN_NOT_OK(ArrowBufferAppendFloat(data_buffer, (float)value));
       break;
-    }
     default:
       return EINVAL;
   }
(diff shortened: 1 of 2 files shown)


src/nanoarrow/array_inline.h
```diff
@@ -303,10 +303,10 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
     case NANOARROW_TYPE_DOUBLE:
       NANOARROW_RETURN_NOT_OK(ArrowBufferAppend(data_buffer, &value, sizeof(double)));
       break;
-    case NANOARROW_TYPE_FLOAT:
-      _NANOARROW_CHECK_RANGE(value, FLT_MIN, FLT_MAX);
-      NANOARROW_RETURN_NOT_OK(ArrowBufferAppendFloat(data_buffer, value));
+    case NANOARROW_TYPE_FLOAT: {
+      NANOARROW_RETURN_NOT_OK(ArrowBufferAppendFloat(data_buffer, (float)value));
       break;
+    }
```
paleolimbot, line 309:
```suggestion
    case NANOARROW_TYPE_FLOAT:
      NANOARROW_RETURN_NOT_OK(ArrowBufferAppendFloat(data_buffer, (float)value));
      break;
```

(Feel free to revert, just a bit out of place with the other cases here)

paleolimbot (approved):
Thank you! Feel free to take or leave the switch case thing.

src/nanoarrow/array_inline.h
lidavidm: Oh my bad, I added it because I added a declaration then changed my mind.

Issue #73, opened 2022-11-25
[C] Support for unions


lidavidm:
While working through the ADBC SQLite driver, I noticed dense union support isn't available. I've also found that writing out the code for maps and lists by hand is a little tedious; would helper functions be useful? (For example, `ArrowSchemaInitList(struct ArrowSchema*, enum ArrowType list_type, enum ArrowType element_type, char* list_field_name)`, where `list_field_name == NULL` means to insert the default field name. Or maybe even omit that and require you to build it by hand if you need to override that.) It would be especially helpful for maps, though we'd have to be careful to make sure this keeps composing well if you have something like `dense_union<..., 4: map<int32, list<int32>>>` (which is in the ADBC spec at one point)

lidavidm:
This could also be split into stages (e.g. just barely enough to initialize a schema by hand at first, then helper functions, appender support, etc.)

paleolimbot:
Definitely! Unions aren't supported in the R package so I'm not familiar with the details. Something that would help me with that is the output of some Python code that constructs some real-world union types and/or arrays.

lidavidm:
Ok, I'll see if I can send a PR your way soon. Thanks!

(I don't think they come up very often to be honest, except in the case of databases trying to handle ambiguously typed data, or trying to shove multiply-typed data into a single column. IIRC DuckDB represents it as "just a struct" with some conventions which I think makes it a bit easier to handle without special cases...)

paleolimbot:
A PR would be great! If you don't get there, an example of a real-world union you'd like to create would do and I can figure something out. This is about to come up in GeoArrow because we need it to represent multiple geometry types.

One of the things I haven't understood in the union is the purpose of the `type_ids` in the schema: wouldn't either the child names or the child indices suffice?

lidavidm:
Child names aren't necessarily unique.

Child indices would generally work. I don't know the original justification off the top of my head, but type IDs as separate from indices would let you subset or rearrange a union without rewriting the type IDs array. In practice I don't think I've ever run into a case where type IDs != indices.

lidavidm:
https://issues.apache.org/jira/browse/ARROW-257 is the actual reason. It was to resolve an early Java/C++ incompatibility: Java used the implementation-defined type ID as the union index, C++ used the child index (as you suggest), and the indirection would allow the two to be compatible. Also you could define a union with more cases than physically present.

lidavidm:
[...]

PR #81, opened 2022-12-05
[C] Adapt ArrowSchema initialization to better support recursive types

paleolimbot:
Will address part of #73 (simplifies creating map, union, and list schemas).

This PR updates the syntax for initializing `ArrowSchema`s to better support lists, unions, and maps. Before this PR, you had to create these schemas by hand (i.e., setting the `format` string and and allocating children yourself). After this PR, the syntax for a list would be:

```c
struct ArrowSchema schema;
ArrowSchemaInit(&schema);
ArrowSchemaSetType(&schema, NANOARROW_TYPE_LIST);
ArrowSchemaSetType(schema.children[0], NANOARROW_TYPE_INT32);
```

...for a `map <int32, string>` would be:

```c
struct ArrowSchema schema;
ArrowSchemaInit(&schema);
ArrowSchemaSetType(&schema, NANOARROW_TYPE_MAP);
ArrowSchemaSetType(schema.children[0].children[0], NANOARROW_TYPE_INT32);
ArrowSchemaSetType(schema.children[0].children[1], NANOARROW_TYPE_STRING);
```

...and for a `union<field1: int32, field2: string>` would be:

```c
struct ArrowSchema schema;
ArrowSchemaInit(&schema);
ArrowSchemaSetTypeUnion(&schema, NANOARROW_TYPE_SPARSE_UNION);
ArrowSchemaSetName(schema.children[0], "field1");
ArrowSchemaSetType(schema.children[0], NANOARROW_TYPE_INT32);
ArrowSchemaSetName(schema.children[1], "field2");
ArrowSchemaSetType(schema.children[1], NANOARROW_TYPE_STRING);
```

This pattern required some breaking changes about how initialization happens...because you can't call `ArrowSchemaInit()` on something that has already been initialized, that pattern was incompatible with an implementation that filled in some information about the children, I had to separate `ArrowSchemaInit()` from `ArrowSchemaSetType()`. To prevent literally all the test from breaking there is also `ArrowSchemaInitFromType()` (i.e., the previous `ArrowSchemaInit()`); however, it's a little confusing to mix them because it's pretty easy to mix up whether or not the children have been initialized. (That could probably be fixed by adding `ArrowSchemaInitChildren()` and promoting its use over `ArrowSchemaAllocateChildren()`).

### DIFF
--- a/src/nanoarrow/schema.c
+++ b/src/nanoarrow/schema.c
@@ -137,8 +137,7 @@ int ArrowSchemaInitChildrenIfNeeded(struct ArrowSchema* schema,
     case NANOARROW_TYPE_LARGE_LIST:
     case NANOARROW_TYPE_FIXED_SIZE_LIST:
       NANOARROW_RETURN_NOT_OK(ArrowSchemaAllocateChildren(schema, 1));
-      NANOARROW_RETURN_NOT_OK(
-          ArrowSchemaInitType(schema->children[0], NANOARROW_TYPE_UNINITIALIZED));
+      ArrowSchemaInit(schema->children[0]);
       NANOARROW_RETURN_NOT_OK(ArrowSchemaSetName(schema->children[0], "item"));
       break;
     case NANOARROW_TYPE_MAP:
@@ -147,6 +146,8 @@ int ArrowSchemaInitChildrenIfNeeded(struct ArrowSchema* schema,
           ArrowSchemaInitType(schema->children[0], NANOARROW_TYPE_STRUCT));
       NANOARROW_RETURN_NOT_OK(ArrowSchemaSetName(schema->children[0], "entries"));
       NANOARROW_RETURN_NOT_OK(ArrowSchemaAllocateChildren(schema->children[0], 2));
+      ArrowSchemaInit(schema->children[0]->children[0]);
+      ArrowSchemaInit(schema->children[0]->children[1]);
       NANOARROW_RETURN_NOT_OK(
           ArrowSchemaSetName(schema->children[0]->children[0], "key"));
       NANOARROW_RETURN_NOT_OK(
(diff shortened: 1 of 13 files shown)


src/nanoarrow/schema.c
```diff
@@ -281,27 +301,67 @@ ArrowErrorCode ArrowSchemaInitDateTime(struct ArrowSchema* schema,
...
-  result = ArrowSchemaSetFormat(schema, buffer);
-  if (result != NANOARROW_OK) {
-    schema->release(schema);
-    return result;
+  return ArrowSchemaSetFormat(schema, buffer);
+}
+
+ArrowErrorCode ArrowSchemaSetTypeUnion(struct ArrowSchema* schema,
+                                       enum ArrowType data_type, int64_t n_children) {
+  if (n_children < 1 || n_children > 127) {
```
lidavidm, line 323: A 0-case union should be valid (and PyArrow happily accepts it).

lidavidm (approved):
Thanks! Just one small note.

Honestly, I think this is everything needed for union support, save maybe making sure that there's a way to append a type code via the appenders.

src/nanoarrow/schema.c
paleolimbot: The zero-case union is apparently not supported by Arrow C++'s C Data interface. I get:

```
'arrow_type' failed with Invalid: Invalid or unsupported format string: '+us:'
```

Reproducer for Python:

```python
import pyarrow as pa
from pyarrow.cffi import ffi
empty_union = pa.sparse_union([])
ptr = ffi.new("struct ArrowSchema*")
empty_union._export_to_c(int(ffi.cast("uintptr_t", ptr)))
pa.DataType._import_from_c(int(ffi.cast("uintptr_t", ptr)))
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
#   File "pyarrow/types.pxi", line 248, in pyarrow.lib.DataType._import_from_c
#   File "pyarrow/error.pxi", line 144, in pyarrow.lib.pyarrow_internal_check_status
#   File "pyarrow/error.pxi", line 100, in pyarrow.lib.check_status
# pyarrow.lib.ArrowInvalid: Invalid or unsupported format string: '+us:'
```

lidavidm: https://github.com/apache/arrow/issues/14855

PR #82, opened 2022-12-06
[C] Union Appender support

paleolimbot:
Adds support for dense and sparse unions in `ArrowArrayAppendNull()` and `ArrowArrayFinishUnionElement()` for #73. Some things I'm not sure about:

- Do we need to support non-standard union IDs in append mode? I don't have a good example of a Union with non-standard TypeIDs on hand and I'm still fuzy on the correct behaviour there.
- Am I constructing the Sparse Union from C++ properly? It seems like I have to manually append to each child which I didn't expect to have to do.
- Is there anything else about Unions that I'm missing?

### DIFF
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -71,7 +71,7 @@ static inline int8_t _ArrowParseUnionTypeIds(const char* type_ids, int8_t* out) 
   char* end_ptr;
   do {
     type_id = strtol(type_ids, &end_ptr, 10);
-    if (end_ptr == type_ids) {
+    if (end_ptr == type_ids || type_id < 0 || type_id > 127) {
       return -1;
     }
 
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -48,6 +48,19 @@ static inline struct ArrowBuffer* ArrowArrayBuffer(struct ArrowArray* array, int
   }
 }
 
+// We don't currently support the case of unions where type_id != child_index;
+// however, these two functions are used to keep track of where that assumption
+// is made.
+static inline int8_t _ArrowArrayUnionChildIndex(struct ArrowArray* array,
+                                                int8_t type_id) {
+  return type_id;
+}
+
+static inline int8_t _ArrowArrayUnionTypeId(struct ArrowArray* array,
+                                            int8_t child_index) {
+  return child_index;
+}
+
 static inline ArrowErrorCode ArrowArrayStartAppending(struct ArrowArray* array) {
   struct ArrowArrayPrivateData* private_data =
       (struct ArrowArrayPrivateData*)array->private_data;
@@ -123,32 +136,40 @@ static inline ArrowErrorCode _ArrowArrayAppendEmptyInternal(struct ArrowArray* a
       array->null_count += n;
       array->length += n;
       return NANOARROW_OK;
-    case NANOARROW_TYPE_DENSE_UNION:
+
+    case NANOARROW_TYPE_DENSE_UNION: {
       // Add one null to the first child and append n references to that child
-      // (Currently assumes type_id == child_index)
+      int8_t type_id = _ArrowArrayUnionTypeId(array, 0);
       NANOARROW_RETURN_NOT_OK(
           _ArrowArrayAppendEmptyInternal(array->children[0], 1, is_valid));
-      NANOARROW_RETURN_NOT_OK(ArrowBufferAppendFill(ArrowArrayBuffer(array, 0), 0, n));
+      NANOARROW_RETURN_NOT_OK(
+          ArrowBufferAppendFill(ArrowArrayBuffer(array, 0), type_id, n));
       for (int64_t i = 0; i < n; i++) {
         NANOARROW_RETURN_NOT_OK(ArrowBufferAppendInt32(
             ArrowArrayBuffer(array, 1), (int32_t)array->children[0]->length - 1));
       }
-      // Null count for unions?
+      // For the purposes of array->null_count, union elements are never considered "null"
+      // even if some children contain nulls.
       array->length += n;
       return NANOARROW_OK;
-    case NANOARROW_TYPE_SPARSE_UNION:
+    }
+
+    case NANOARROW_TYPE_SPARSE_UNION: {
       // Add n nulls to the first child and append n references to that child
-      // (Currently assumes type_id == child_index)
+      int8_t type_id = _ArrowArrayUnionTypeId(array, 0);
       NANOARROW_RETURN_NOT_OK(
           _ArrowArrayAppendEmptyInternal(array->children[0], n, is_valid));
       for (int64_t i = 1; i < array->n_children; i++) {
         NANOARROW_RETURN_NOT_OK(ArrowArrayAppendEmpty(array->children[i], n));
       }
 
-      NANOARROW_RETURN_NOT_OK(ArrowBufferAppendFill(ArrowArrayBuffer(array, 0), 0, n));
-      // Null count for unions?
+      NANOARROW_RETURN_NOT_OK(
+          ArrowBufferAppendFill(ArrowArrayBuffer(array, 0), type_id, n));
+      // For the purposes of array->null_count, union elements are never considered "null"
+      // even if some children contain nulls.
       array->length += n;
       return NANOARROW_OK;
+    }
 
     case NANOARROW_TYPE_FIXED_SIZE_LIST:
       NANOARROW_RETURN_NOT_OK(ArrowArrayAppendEmpty(
@@ -477,12 +498,11 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
 }
 
 static inline ArrowErrorCode ArrowArrayFinishUnionElement(struct ArrowArray* array,
-                                                          int32_t type_id) {
+                                                          int8_t type_id) {
   struct ArrowArrayPrivateData* private_data =
       (struct ArrowArrayPrivateData*)array->private_data;
 
-  // Currently this assumes type_ids == child index
-  int64_t child_index = type_id;
+  int64_t child_index = _ArrowArrayUnionChildIndex(array, type_id);
   if (child_index < 0 || child_index >= array->n_children) {
     return EINVAL;
   }
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -94,6 +94,23 @@ static inline int8_t _ArrowParseUnionTypeIds(const char* type_ids, int8_t* out) 
   return -1;
 }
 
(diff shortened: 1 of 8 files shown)


src/nanoarrow/nanoarrow.h
```diff
@@ -835,6 +838,15 @@ static inline ArrowErrorCode ArrowArrayAppendString(struct ArrowArray* array,
...
 static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array);
 
+/// \brief Finish a union array element
+///
+/// Appends an element to the union type ids buffer and increments array->length.
+/// For sparse unions, up to one element is added to non type-id children. Returns
+/// EINVAL if the underlying storage type is not a union, if type_id is not valid,
+/// or if child sizes after appending are inconsistent.
+static inline ArrowErrorCode ArrowArrayFinishUnionElement(struct ArrowArray* array,
+                                                          int32_t type_id);
```
lidavidm, line 848: int8_t type_id?

lidavidm: So, I would say there's no 'standard type ID', there's just the convention of making type ID == child index. Since we take type ID here, there's nothing more to do (in terms of this signature).

src/nanoarrow/array_test.cc
```diff
@@ -1070,6 +1070,97 @@ TEST(ArrayTest, ArrayTestAppendToStructArray) {
...
+  auto arrow_array = ImportArray(&array, &schema);
+  ARROW_EXPECT_OK(arrow_array);
+
+  auto child_builder_int = std::make_shared<Int64Builder>();
+  auto child_builder_string = std::make_shared<StringBuilder>();
+  std::vector<std::shared_ptr<ArrayBuilder>> children = {child_builder_int,
+                                                         child_builder_string};
+  auto builder = SparseUnionBuilder(default_memory_pool(), children,
+                                    arrow_array.ValueUnsafe()->type());
+  // Arrow's SparseUnionBuilder requires explicit empty value appends?
```
lidavidm, line 1147: Odd, but I guess that's fine. Might be worth filing an issue. (Though changing it would be a break...)

src/nanoarrow/nanoarrow.h
```diff
@@ -785,6 +785,9 @@ ArrowErrorCode ArrowArrayReserve(struct ArrowArray* array,
 /// \brief Append a null value to an array
 static inline ArrowErrorCode ArrowArrayAppendNull(struct ArrowArray* array, int64_t n);
 
+/// \brief Append an empty, non-null value to an array
+static inline ArrowErrorCode ArrowArrayAppendEmpty(struct ArrowArray* array, int64_t n);
```
lidavidm, line 789: Do we want to add an explicit unit test for this?

src/nanoarrow/array_inline.h
```diff
@@ -107,29 +107,69 @@ static inline ArrowErrorCode _ArrowArrayAppendBits(struct ArrowArray* array,
...
   }
 
-  if (private_data->storage_type == NANOARROW_TYPE_NA) {
-    array->null_count += n;
-    array->length += n;
-    return NANOARROW_OK;
+  // Some type-specific handling
+  switch (private_data->storage_type) {
+    case NANOARROW_TYPE_NA:
+      array->null_count += n * !is_valid;
```
lidavidm, line 122: Hmm, for NA, null count always has to equal length right?

src/nanoarrow/array_inline.h
```diff
@@ -107,29 +107,69 @@ static inline ArrowErrorCode _ArrowArrayAppendBits(struct ArrowArray* array,
...
+      array->length += n;
+      return NANOARROW_OK;
+    case NANOARROW_TYPE_DENSE_UNION:
+      // Add one null to the first child and append n references to that child
+      // (Currently assumes type_id == child_index)
```
lidavidm, line 127: Yeah, so what Arrow does is keep a 128-element lookup table in the type object that maps type ID to child index. That might be hard here...

src/nanoarrow/array_inline.h
```diff
@@ -107,29 +107,69 @@ static inline ArrowErrorCode _ArrowArrayAppendBits(struct ArrowArray* array,
...
-    array->length += n;
-    return NANOARROW_OK;
+  // Some type-specific handling
+  switch (private_data->storage_type) {
+    case NANOARROW_TYPE_NA:
+      array->null_count += n * !is_valid;
+      array->length += n;
+      return NANOARROW_OK;
+    case NANOARROW_TYPE_DENSE_UNION:
+      // Add one null to the first child and append n references to that child
```
lidavidm, line 126: How would this interact with children declared non-nullable?

src/nanoarrow/array_inline.h
```diff
@@ -440,6 +473,51 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+      for (int64_t i = 0; i < array->n_children; i++) {
+        if (i == child_index || array->children[i]->length == (array->length + 1)) {
+          continue;
+        }
+
+        if (array->children[i]->length != array->length) {
+          return EINVAL;
+        }
+
+        NANOARROW_RETURN_NOT_OK(ArrowArrayAppendEmpty(array->children[i], 1));
```
lidavidm, line 506: Hmm, why not just append a null here?

src/nanoarrow/array_inline.h
lidavidm: Maybe the appender's signature should require the caller to maintain this mapping (and if not given, assume type ID == child index for simplicity?)

src/nanoarrow/array_inline.h
lidavidm: (That said, not sure if libarrow handles this well either off the top of my head...)

src/nanoarrow/array_test.cc
paleolimbot: It's technically documented (although the API is marked experimental): https://github.com/apache/arrow/blob/master/cpp/src/arrow/array/builder_union.h#L245

src/nanoarrow/array_inline.h
paleolimbot: Yes, that's a good catch. Appending an empty to a null *is* a null, I suppose.

src/nanoarrow/array_inline.h
paleolimbot: Maybe because an empty will result in fewer validity buffers getting allocated? The flip side of that is that reinterpreting the dense union as a struct is easier if the irrelevant values are already flagged as null. I'm happy to defer to whatever you think is best...the SparseUnionBuilder doesn't give any hints as to which is better.

lidavidm: Empty sounds fine then!

src/nanoarrow/array_inline.h
paleolimbot: The way the append behaviour is currently written, the builder doesn't have access to a schema against which to check for nullability. That check should exist, though (probably `ArrowArrayViewValidateFull()` or something).

lidavidm:
The null count is always 0 - unions have no validity bitmap. (This was changed right before 1.0.0, because it was deemed redundant. Honestly, I wonder if we should've accepted the redundancy, because it makes unions an odd duck out in every aspect of the codebase.)

src/nanoarrow/array_inline.h
paleolimbot: Ok...huge rabbit hole, but it's pretty common to do `ArrowArrayInitFromSchema()`, and if *that* schema has the non-sequential type ids, then `ArrowArrayStartAppending()` will error. I think that is enough of a guardrail for now.

src/nanoarrow/array_inline.h
paleolimbot: Again, huge rabbit hole; however, I think I have this assumption checked wherever possible. In the future one could allocate the forward and reverse lookup tables as part of the `ArrowArrayPrivate` struct.

src/nanoarrow/nanoarrow.h
paleolimbot: I added tests for all the types since this one shows up a lot in recursive building.

src/nanoarrow/schema_test.cc
```diff
@@ -1192,9 +1197,7 @@ TEST(SchemaViewTest, SchemaViewInitNestedUnion) {
   EXPECT_EQ(schema_view.layout.element_size_bits[0], 8);
   EXPECT_EQ(schema_view.layout.element_size_bits[1], 32);
   EXPECT_EQ(schema_view.layout.element_size_bits[2], 0);
-  EXPECT_EQ(
-      std::string(schema_view.union_type_ids.data, schema_view.union_type_ids.n_bytes),
-      std::string("0"));
+  EXPECT_EQ(std::string(schema_view.union_type_ids), std::string("0"));
```
lidavidm, line 1200: nit: if they're both null-terminated strings, you can `EXPECT_STREQ` (not a big deal here though)

lidavidm:
> I have never disliked unions more; however, I believe that nanoarrow has emerged better for having dealt with them properly here. There is still implementing the `ArrowArrayGetXXXX()` functions (one for another PR, I think)

Heh. I suppose it would at least be more consistent if they did have the validity bitmap. Or if we only supported 'sparse' unions, they could just have been a special case of structs.

src/nanoarrow/array_inline.h
```diff
@@ -48,10 +48,87 @@ static inline struct ArrowBuffer* ArrowArrayBuffer(struct ArrowArray* array, int
...
+static inline int8_t _ArrowParseUnionTypeIds(const char* type_ids, int8_t* out) {
+  if (*type_ids == '\0') {
+    return 0;
+  }
+
+  int32_t i = 0;
+  long type_id;
+  char* end_ptr;
+  do {
+    type_id = strtol(type_ids, &end_ptr, 10);
```
lidavidm, line 73: Should we check that type_id is >= 0 and <128?

src/nanoarrow/array_inline.h
```diff
@@ -48,10 +48,87 @@ static inline struct ArrowBuffer* ArrowArrayBuffer(struct ArrowArray* array, int
...
+  } while (1);
+
+  return -1;
+}
+
+static inline int8_t _ArrowUnionTypeIdsWillEqualChildIndices(const char* type_id_str,
+                                                             int64_t n_children) {
+  int8_t type_ids[128];
+  int8_t n_type_ids = _ArrowParseUnionTypeIds(type_id_str, type_ids);
+  if (n_type_ids != n_children) {
```
lidavidm, line 101: That seems more like a hard error, though I suppose it's caught elsewhere?

lidavidm (approved):

src/nanoarrow/array_inline.h
paleolimbot: I added a check in `ArrowSchemaViewInit()` (previously I just passed the entire string along without checking it apparently)

PR #83, opened 2022-12-08
[C] Add read support for unions in the `ArrowArrayView`

### LINKED ISSUE #73: [C] Support for unions

paleolimbot:
Ok, I think this has everything one would need to inspect union types without having to remember the spec. Given a valid union schema, one can:

```c
ArrowArrayViewInitFromSchema(&array_view, &schema, nullptr);
int8_t child_index = ArrowArrayViewUnionChildIndex(&array_view, i);
int64_t child_offset = ArrowArrayViewUnionChildOffset(&array_view, i);
// Figure out which accessor you need. If it's an int, you could
struct ArrowArrayView* child = array_view.children[child_index];
ArrowArrayViewGetIntUnsafe(child, child_offset);
```

Fixes #73.

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -540,10 +540,12 @@ ArrowErrorCode ArrowArrayViewInitFromSchema(struct ArrowArrayView* array_view,
       return ENOMEM;
     }
 
+    memset(array_view->union_type_id_map, -1, 256);
     int8_t n_type_ids = _ArrowParseUnionTypeIds(schema_view.union_type_ids,
-                                                array_view->union_type_id_map);
-    for (int8_t i = 0; i < n_type_ids; i++) {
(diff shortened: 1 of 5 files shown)


lidavidm (approved):
This is awesome! Thanks!

PR #93, opened 2023-01-16
Update build configuration for IPC extension

### LINKED ISSUE #88: [Extension-IPC] Refactor build system for IPC extension
The initial proof-of-concept version of the IPC extension (#61) added a minimal CMake configuration to get started; however, it is almost certainly not an ideal build configuration for the real world. There are at least three high-level changes that might be useful

First, the extension follows a pattern where "nanoarrow_ipc.h" does not `#include "nanoarrow.h"`. This follows our own documentation's advice around how libraries should use nanoarrow...nanoarrow should generally be considered an implementation detail, not the basis for an extension library. However, for the extensions we define internally like the IPC extension, it's very likely that the caller is going to want to use nanoarrow and thus following our own advice would necessitate two copies of nanoarrow for most users. I think that extensions should be allowed to `#include "nanoarrow.h"` for this reason and to reduce duplication since many basic library concepts overlap (e.g., `ArrowError`, `ArrowErrorCode`, `ArrowBufferView`).

Second, the cmake contains some `FetchContent` logic to build and link the flatcc flatbuffers implementation. There should almost certainly be a "link to shared" option (maybe via `find_package()`) and/or a vendored option since flatcc is a bit of an obscure dependency.

Third, we should figure out how to bundle the extension so that it can be added to the `dist/` folder like nanoarrow.h and nanoarrow.c. To allow for some build flexibility with flatcc, I imagine this should be a few files: nanoarrow_ipc.h, nanoarrow_ipc_generated.h, and nanoarrow_ipc_flatcc.c.

paleolimbot:
Closes #88.

Currently this PR:

- Vendors flatcc into the repo
- Allows linking to a shared install (e.g., brew)
- Bundles all the generated files into a single header (easier to manage)

In theory, you can use the "reader" headers without any runtime at all; however, then you'd be able to "read" without the "verify" step, which seems like something that shouldn't be the "easy default".

I toyed with the idea of smushing flatcc into a single bundle; however, the implementation is pretty complex and it would be hard (I think).

TODO:

- Github Actions CI tests for the various configurations
- Consider `#include "nanoarrow.h"` in the nanoarrow_ipc.h header.


lidavidm (approved):
If we're going to vendor FlatCC, we should update LICENSE.txt here too. See https://infra.apache.org/licensing-howto.html. FlatCC appears to be Apache-2 so as it says we technically don't need to do anything, but it would be good to have a brief notice.

PR #107, opened 2023-02-13
fix: set map entries/key to non-nullable

### LINKED ISSUE #106: [C] `SchemaSetType(MAP)` should set entries/keys as non-nullable
From apache/arrow-adbc#453

We should set it here (since the default is to make things nullable): https://github.com/apache/arrow-nanoarrow/blob/5771e5ec291da7d4e7abed1cdc671822c3c2e4f4/src/nanoarrow/schema.c#L143-L155

`SchemaViewValidateMap` should also validate this

lidavidm:
Fixes #106.

### DIFF
--- a/src/nanoarrow/schema_test.cc
+++ b/src/nanoarrow/schema_test.cc
@@ -139,6 +139,12 @@ TEST(SchemaTest, SchemaTestInitNestedMap) {
             NANOARROW_OK);
   ASSERT_EQ(ArrowSchemaSetType(schema.children[0]->children[1], NANOARROW_TYPE_STRING),
             NANOARROW_OK);
+  ASSERT_STREQ("entries", schema.children[0]->name);
+  ASSERT_STREQ("key", schema.children[0]->children[0]->name);
+  ASSERT_STREQ("value", schema.children[0]->children[1]->name);
+
+  ASSERT_FALSE(schema.children[0]->flags & ARROW_FLAG_NULLABLE);
+  ASSERT_FALSE(schema.children[0]->children[0]->flags & ARROW_FLAG_NULLABLE);
 
   auto arrow_type = ImportType(&schema);
   ARROW_EXPECT_OK(arrow_type);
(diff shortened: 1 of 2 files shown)


lidavidm:
Hmm, seems na_map in R does trip the new validation

lidavidm:
Not sure if this is the right way to fix the R issue - maybe we should copy the key type and set it to non-nullable in na_map?

src/nanoarrow/schema_test.cc
```diff
@@ -139,6 +139,12 @@ TEST(SchemaTest, SchemaTestInitNestedMap) {
             NANOARROW_OK);
   ASSERT_EQ(ArrowSchemaSetType(schema.children[0]->children[1], NANOARROW_TYPE_STRING),
             NANOARROW_OK);
+  ASSERT_STREQ("entries", schema.children[0]->name);
+  ASSERT_STREQ("key", schema.children[0]->children[0]->name);
+  ASSERT_STREQ("value", schema.children[0]->children[1]->name);
+
+  ASSERT_FALSE(schema.children[0]->flags & ARROW_FLAG_NULLABLE);
+  ASSERT_FALSE(schema.children[0]->children[0]->flags & ARROW_FLAG_NULLABLE);
```
paleolimbot, line 147: Is there a reason why these are `ASSERT`s and not `EXPECT`s? The other expectations I've written also use `XXX(actual, expected)` (but I never looked up which was correct or if there is a correct way to do it...that's just how R's testing works).

paleolimbot (approved):
Thank you!

The R change is the best way to solve that right now...those "type" constructors are somewhere between a field and a type and the change you made seems the safest for now.

src/nanoarrow/schema_test.cc
lidavidm: ah, no particular reason other than habit.

IIRC, in Googletest:
- Assert aborts the test if it fails, Expect lets it continue
- Googletest goes `expected, actual`

lidavidm: Hmm, though the actual docs don't list this - they just use `val1, val2`. I'll look at it tomorrow

lidavidm: Made them consistent with the surrounding code.

PR #112, opened 2023-02-21
fix: make sure bundled nanoarrow is also valid C++

paleolimbot:
Apparently it's difficult to pass different flags to C and C++ files in a Python extension. The C library was almost valid C++ anyway and because there are quite a few static inline functions, it's probably a good idea to make sure that the project compiles under a C++ compiler as well (since the nanoarrow.h header will be compiled with C++ if included in a C++ project).

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -40,9 +40,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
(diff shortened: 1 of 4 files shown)


lidavidm (approved):
LGTM, though I wonder if it'll work in general (e.g. if bundled with other C++ sources, users might have warnings enabled that forbid things like C-style casts)

PR #117, opened 2023-02-23
feat(python): Python schema, array, and array view skeleton

paleolimbot:
This PR is an attempt to add minimum usable Python bindings to the nanoarrow C library. That minimum scope is essentially just the ability to extract field values from `ArrowSchema`/`ArrowArray`/`ArrowArrayStream` objects in a way that will not crash Python. This PR also includes bindings for nanoarrow's `ArrowSchemaView` (so that the parameters of parameterized types can be extracted) and `ArrowArrayView` (so that buffer types/sizes can be exported using the Python buffer protocol).

I've updated [the README](https://github.com/paleolimbot/arrow-nanoarrow/tree/python-tidbits/python#readme) to showcase the extent of the bindings as implemented in this PR; several basic examples are also provided below.

Example schema usage:

```python
import nanoarrow as na
import pyarrow as pa
schema = na.schema(pa.decimal128(10, 3))
print(schema.format)
#> d:10,3
print(schema.view().decimal_precision)
#> 10
print(schema.view().decimal_scale)
#> 3
```

Example Array usage:

```python
array = na.array(pa.array(["one", "two", "three", None]))
print(array.length)
#> 4
print(array.null_count)
#> 1

import numpy as np
view = array.view()
[np.array(buffer) for buffer in view.buffers]
#> [array([7], dtype=uint8),
#>  array([ 0,  3,  6, 11, 11], dtype=int32),
#>  array([b'o', b'n', b'e', b't', b'w', b'o', b't', b'h', b'r', b'e', b'e'],
#>        dtype='|S1')]
```

Example ArrayStream usage:

```python
pa_array_child = pa.array([1, 2, 3], pa.int32())
pa_array = pa.record_batch([pa_array_child], names=["some_column"])
reader = pa.RecordBatchReader.from_batches(pa_array.schema, [pa_array])
array_stream = na.array_stream(reader)

print(array_stream.get_schema())
#> struct<some_column: int32>

for array in array_stream:
    print(array.length)
#> 3

print(array_stream.get_next() is None)
#> True
```


python/setup.py
```diff
@@ -17,29 +17,23 @@
...
-shutil.copy(
-    Path(__file__).parent / "../dist/nanoarrow.c", target / "nanoarrow.c"
-)
-shutil.copy(
-    Path(__file__).parent / "../dist/nanoarrow.h", target / "nanoarrow.h"
-)
+# Run bootstrap.py to run cmake generating a fresh bundle based on this
+# checkout or copy from ../dist if the caller doesn't have cmake available
+this_dir = os.path.dirname(__file__)
+subprocess.run([sys.executable, os.path.join(this_dir, 'bootstrap.py')])
```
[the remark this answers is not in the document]
jorisvandenbossche: Eventually (later), we will also want to have some better logic to determine if this step needs to be run or not, for example when distributing sdists (see eg https://github.com/apache/arrow-adbc/blob/923e0408fe5a32cc6501b997fafa8316ace25fe0/python/adbc_driver_manager/setup.py#L33-L50)

python/src/nanoarrow/_lib.pyx
```diff
@@ -84,3 +85,329 @@ def as_numpy_array(arr):
...
+    def __init__(self):
+        ArrowArrayViewInitFromType(&self.c_array_view, NANOARROW_TYPE_UNINITIALIZED)
+
+    def __del__(self):
+        ArrowArrayViewReset(&self.c_array_view)
+
+    def _addr(self):
+        return <uintptr_t>&self.c_array_view
+
+cdef class CSchema:
```
jorisvandenbossche, line 131: Is the reason you call it CSchema (with C prefix) because this represents a "C" Data Interface schema?

python/tests/test_nanoarrow.py
```diff
@@ -1,27 +1,155 @@
...
 
     arr = pa.array([[1], [2, 3]])
     with pytest.raises(TypeError, match="Cannot convert a non-primitive array"):
-        nanoarrow.as_numpy_array(arr)
+        na.as_numpy_array(arr)
+
+def test_schema_basic():
+    # Blank invalid schema
+    schema = na.CSchema.Empty()
+    assert(schema.is_valid() is False)
```
jorisvandenbossche, line 51:
```suggestion
    assert schema.is_valid() is False
```

Styling nitpick: those brackets are not needed

(we should add black as pre-commit hook like in adbc, and that should also take care of this, I think: https://github.com/apache/arrow-adbc/blob/main/.pre-commit-config.yaml#L65)

python/tests/test_nanoarrow.py
```diff
@@ -1,27 +1,155 @@
...
+    # Check on the child
+    child = schema.children[0]
+    child_info = child.parse()
+    assert(child_info['type'] == 'int32')
+    assert(child_info['storage_type'] == 'int32')
+    assert(child_info['name'] == 'col1')
+
+def test_schema_info_params():
+    schema = na.CSchema.Empty()
+    pa.binary(12)._export_to_c(schema._addr())
```
jorisvandenbossche, line 88: It might be nice to add a `from_pyarrow`-like helper to avoid this constant pattern of creating an empty and calling export_to_c (this could also be just an internal helper for the tests if we don't want to expose that publicly (yet))

python/src/nanoarrow/_lib.pyx
```diff
@@ -84,3 +85,329 @@ def as_numpy_array(arr):
     # TODO set base
 
     return result
+
+
+def version():
+    return ArrowNanoarrowVersion().decode("UTF-8")
+
+cdef class CSchemaHolder:
```
jorisvandenbossche, line 93: Those "Holder" classes are only for when we "own" the pointer, and then to ensure t is properly released?  (to serve as the "base" object in that case, while if you create the CSchema/CArray class from another object that owns the pointer, this other object is the `_base`)

python/src/nanoarrow/_lib.pyx
```diff
@@ -84,3 +85,329 @@ def as_numpy_array(arr):
...
+        cdef int result = ArrowArrayViewInitFromSchema(&holder.c_array_view,
+                                                       self._schema._ptr, &error)
+        if result != NANOARROW_OK:
+            raise ValueError(ArrowErrorMessage(&error))
+
+        result = ArrowArrayViewSetArray(&holder.c_array_view, self._ptr, &error)
+        if result != NANOARROW_OK:
+            raise ValueError(ArrowErrorMessage(&error))
+
+        return CArrayView(holder, holder._addr(), self)
```
jorisvandenbossche, line 263: It sounds a bit strange that "validate" returns a array view object.

(I would personally also add all properties of CArrayView like length, buffers, etc to the CArray class itself)

jorisvandenbossche (commented):
Nice! Just a few quick comments

python/.gitignore
```diff
@@ -18,7 +18,8 @@
 
 src/nanoarrow/nanoarrow.c
 src/nanoarrow/nanoarrow.h
-src/nanoarrow/*.cpp
+src/nanoarrow/nanoarrow_c.pxd
```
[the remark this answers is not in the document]
paleolimbot: It is a pretty useful file to have on hand...nanoarrow.c/h are built for vendoring and there's no reason another Python package couldn't use the same .pxd (provided it's generated from the correct nanoarrow.h).

Maybe it could get checked into the dist/ folder next to the bundled nanoarrow/h?

python/src/nanoarrow/_lib.pyx
paleolimbot: Yes...it could be a capsule or this object but is also usefully the parent array (so that you can do `array.children[0]` and re-use the `CSchema` class.

python/setup.py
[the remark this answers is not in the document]
paleolimbot: There's a bootstrap.R file too and I really like keep the logic separate since it's rather extensive and maybe should even be tested separately. I forgot about sdist...I made a few changes to make sure that (1) bootstrap.py wouldn't be included in sdist and (2) make sure not to error if it isn't there. Probably this will need some workshopping come 'we have to build 39348 wheels' time.

python/src/nanoarrow/_lib.pyx
jorisvandenbossche: In pyarrow we have the pattern of calling the exposed C(++) names with a "C" prefix, and then the corresponding cython classes without. 
So the equivalent here would be to use CArrowSchema (instead of ArrowSchema) for nanoarrow.c one (and so for the declarations in the pxd file), and then ArrowSchema for the python class here.

I assume that what you describe (CSchema and Schema) can be done in a single class as well?

python/src/nanoarrow/_lib.pyx
paleolimbot: This was a bit of a rabbit hole but you're totally right. Now you can do `array.view()` and `schema.view()`, which is a bit more intuitive since you get a `SchemaView` and `ArrayView` back. I also completed the element accessors on the array and schema classes.

python/tests/test_nanoarrow.py
paleolimbot: I'm going to punt on pyarrow integration just this second...for the tests it's only one extra line to `_export_to_c()`.

python/tests/test_nanoarrow.py
paleolimbot: I updated all the asserts...I might punt on setting up pre-commit-config (although it definitely should be done for more than just that!). I quite like how ADBC did it.

python/src/nanoarrow/_lib.pyx
paleolimbot: You're totally right..it's all just `Schema`, `Array`, etc. now. There probably should never be anything higher level in nanoarrrow anyway!

python/.gitignore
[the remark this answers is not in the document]
paleolimbot: Checking it into dist/ would be fine because users of that would just copy the whole folder into their python project and the .h/.pxd would stay together?

For local development and CI it's essential to have cmake generate the bundle (or else you could make changes in the C library that break the Python package and CI wouldn't catch it). Also important to include copy from ../dist as a fallback so that installing by URL works via pip. I tested this on Mac and Windows and Linux but it probably needs some flushing out.

jorisvandenbossche (commented):
Nice!

I think I mentioned it before, but personally I am not sure what the benefit is of the splits between the base class and the View class (eg Array and ArrayView). 
I understand the reason that this separate structure exists in C nanoarrow, but here we are already creating a class wrapping the ArrowArray/Schema/Stream struct anyway. And from a python user perspective, I wouldn't really be able to explain why this is needed. For example, being able to do `arr.buffers[1]` is just simpler than `arr.view().buffers[1]` (also interactively, doing the `view()` makes that tab completion doesn't work nicely)

python/nanoarrow/_lib.pyx
```diff
@@ -0,0 +1,869 @@
...
+                raise NanoarrowException(
+                    "ArrowArrayStream::get_next()",
+                    code,
+                    message.decode("UTF-8")
+                )
+            else:
+                Error.raise_error("ArrowArrayStream::get_next()", code)
+
+        if not array.is_valid():
+            return None
```
jorisvandenbossche, line 856: Another option could be to raise a StopIteration exception when the stream is finished.

.github/workflows/python.yaml
```diff
@@ -57,3 +57,31 @@ jobs:
       - name: Run tests
         run: |
           pytest python/tests -v -s
+
+      - name: Run doctests
+        if: success() && matrix.python-version == '3.10'
+        run: |
+          # Needs editable install to run --doctest-cython
+          pip install pytest-cython
+          pip install -e python
```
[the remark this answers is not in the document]
paleolimbot: I think it's installed without `-e` above (maybe I'll just add it there, and maybe also add `pytest-cython` to pyproject.toml while I'm at it).

python/nanoarrow/_lib.pyx
```diff
@@ -0,0 +1,869 @@
...
+    cdef object _base
+    cdef ArrowSchema* _ptr
+
+    @staticmethod
+    def allocate():
+        base = SchemaHolder()
+        return Schema(base, base._addr())
+
+    def __init__(self, object base, uintptr_t addr):
+        self._base = base,
```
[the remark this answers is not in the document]
paleolimbot: Definitely not!

python/nanoarrow/_lib.pyx
```diff
@@ -0,0 +1,869 @@
...
+            raise IndexError(f"{k} out of range [0, {self._length})")
+        return Array(self._parent, self._child_addr(k), self._parent.schema.children[k])
+
+    cdef _child_addr(self, int64_t i):
+        cdef ArrowArray** children = self._parent._ptr.children
+        cdef ArrowArray* child = children[i]
+        return <uintptr_t>child
+
+
+cdef class ArrayViewChildren:
```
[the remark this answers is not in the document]
paleolimbot: It's slightly different: `ArrowArrayViewSetArray()` is recursive, so `[x for x in array_view.children]` involves fewer calculations than `[x.view() for x in array.children]`. I don't know if that matters (maybe faster if you had many many columns).

python/nanoarrow/_lib.pyx
```diff
@@ -0,0 +1,869 @@
...
+        self._assert_valid()
+        return self._ptr.offset
+
+    @property
+    def null_count(self):
+        return self._ptr.null_count
+
+    @property
+    def buffers(self):
+        return tuple(<uintptr_t>self._ptr.buffers[i] for i in range(self._ptr.n_buffers))
```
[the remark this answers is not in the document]
paleolimbot: Possibly just philosophically aligned with what the `ArrowArray` actually provides. If you're trying to debug something this would tell you if one of your pointers is NULL...to get the `view().buffers` the array would have to be validated against a `Schema`.

.github/workflows/python.yaml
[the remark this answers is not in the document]
jorisvandenbossche: Does the pytest-cython need the package to be installed in editable mode? (I would expect it to work with the installed package as well)

python/nanoarrow/_lib.pyx
```diff
@@ -0,0 +1,869 @@
...
+        self._cached_schema = schema
+
+    def get_schema(self):
+        """Get the schema associated with this stream
+        """
+        # Update the cached copy of the schema as an independent object
+        self._cached_schema = Schema.allocate()
+        self._get_schema(self._cached_schema)
+
+        # Return an independent copy
```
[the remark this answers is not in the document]
paleolimbot: I think I wanted to make sure that calling `.get_schema()` on the Python object definitely called `.get_schema()` on the C object, but also that calling `.get_next()` didn't call `.get_schema()` every single time. I don't know that the first use-case is ever going to come up, or perhaps I'm overemphasizing the "test your C code from Python" use case.

python/nanoarrow/_lib.pyx
paleolimbot: I think doing whatever `read_next_batch()` does is worth doing here. The R record batch reader returns NULL which is how I got here.

.github/workflows/python.yaml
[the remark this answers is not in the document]
paleolimbot: Locally, yes. I've found it to be rather finicky (but useful enough to play along with for now 🤷 ).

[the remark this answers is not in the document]
jorisvandenbossche: FWIW, we use this package in Arrow CI with a normal install of pyarrow (non-editable), so I think it should certainly work. What you might have noticed locally is that a test didn't pass after updating it if you didn't re-install it? (that's the reason for locally using editable, so you don't have to reinstall after every tiny change. But on CI that doesn't matter)

python/nanoarrow/_lib.pyx
```diff
@@ -526,23 +546,28 @@ cdef class ArrayView:
...
-    def array(self):
-        return self._array
+        if self._ptr.dictionary == NULL:
+            return None
+        else:
+            return ArrayView(
+                self,
+                <uintptr_t>self._ptr.dictionary,
+                self._schema.dictionary,
+                None
```
jorisvandenbossche, line 556: Why is this None in this case?

python/nanoarrow/_lib.pyx
```diff
@@ -730,9 +761,13 @@ cdef class ArrayViewBuffers:
     cdef ArrayView _array_view
     cdef int64_t _length
 
-    def __init__(self, ArrayView array_view):
+    def __cinit__(self, ArrayView array_view):
         self._array_view = array_view
-        self._length = array_view._array._ptr.n_buffers
+        self._length = 3
```
jorisvandenbossche: Just out of curiosity, the `n_buffers` member of ArrowArray is not always accurate (or not what we want here)?

paleolimbot: It's just to remove the reference to the `_array`, since that no longer exists. The `ArrowArrayView` in C doesn't have an explicit `n_buffers` member (but maybe should since the shenanigan on the following lines gets used quite a bit in C as well).

python/nanoarrow/_lib.pyx
paleolimbot: The buffer shelter is None here because the child/dictionary's reference to `self` is sufficient to keep the buffers valid.

.github/workflows/python.yaml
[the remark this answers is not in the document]
paleolimbot: I'll update this...there are many things that could have been wrong with my local setup!

python/nanoarrow/_lib.pyx
jorisvandenbossche: Thanks for the explanation! And yes, seeing this code here, adding `n_buffers` to the view object might be useful.

PR #140, opened 2023-03-01
chore: Update Arch linux package list for release verification

### LINKED ISSUE #139: Minor note on release verification documentation for Arch Linux
The Arch Linux instructions in https://github.com/apache/arrow-nanoarrow/blob/main/dev/release/README.md says to run the following to install the needed deps:

```
pacman -S git g++ make cmake r-base gnupg curl arrow
```

This results in an error about g++ not being found. The package on Arch to get g++ appears to be [`gcc`](https://archlinux.org/packages/core/x86_64/gcc/).

amoeba:
Closes #139

### DIFF
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -170,7 +170,7 @@ On Arch Linux (e.g., `docker run --rm -it archlinux:latest`, you can install all
 using `pacman`):
 
 ```bash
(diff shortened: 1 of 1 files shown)


amoeba:
@paleolimbot

PR #142, opened 2023-03-02
fix: include compilers in conda instructions

jorisvandenbossche:


jorisvandenbossche:
I needed this to get the verification working locally. If this is not included, it is detecting and using my system gcc, which resulted in some compilation errors. 

```
/usr/bin/ld: /home/joris/miniconda3/envs/nanoarrow-verify-rc/lib/libre2.so.10: undefined reference to `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::reserve()@GLIBCXX_3.4.29'
/usr/bin/ld: /home/joris/miniconda3/envs/nanoarrow-verify-rc/lib/libarrow.so.1000.1.0: undefined reference to `std::__throw_bad_array_new_length()@GLIBCXX_3.4.29'
...
```

PR #143, opened 2023-03-02
feat(extensions/nanoarrow_ipc): Decode RecordBatch message to ArrowArray

### LINKED ISSUE #91: [Extension-IPC] Implement RecordBatch message decoding
The initial proof-of-concept IPC reader (#61) only decodes the Schema message. In addition, support needs to be added for the RecordBatch message type defined at https://github.com/apache/arrow/blob/master/format/Message.fbs#L81-L102 . This should be more or less a mechanical (i.e., type-agnostic) process...nanoarrow exposes logic to expose the number of buffers for each type and to check the lengths of buffers. Some new checks might have to be added to nanoarrow (e.g., checking all offset values instead of just the last offset value) to ensure that downstream readers won't access memory they aren't supposed to.

The easiest way to implement this involves a copy operation for each buffer. I will have a closer look at the C++ implementation but I am guessing that there is an option - or maybe it's the default - to make this a zero-copy operation (e.g., in case the input buffer is memory-mapped). In this case the caller would have to provide a buffer deallocator for the input and when constructing the buffers we'd have to use a shared pointer or something to make sure it stays alive for the duration of the longest-lived buffer. Enabling the zero-copy bit is probably best suited for a follow-up PR since it may involve at least one non-trivial build system update (using C++).

paleolimbot:
Closes #91.

### DIFF
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
@@ -818,32 +818,26 @@ static void ArrowIpcReaderCountBuffers(struct ArrowArray* array, int64_t* n_buff
   }
 }
 
-static void ArrowIpcReaderInitFields(struct ArrowIpcField** cursor,
+static void ArrowIpcReaderInitFields(struct ArrowIpcField* fields,
                                      struct ArrowArrayView* view,
-                                     struct ArrowSchema* schema, struct ArrowArray* array,
+                                     struct ArrowArray* array, int64_t* n_fields,
                                      int64_t* n_buffers) {
-  struct ArrowIpcField* field = *cursor;
-  field->schema = schema;
+  struct ArrowIpcField* field = fields + (*n_fields);
   field->array = array;
   field->array_view = view;
   field->buffer_offset = *n_buffers;
   *n_buffers += array->n_buffers;
-  *cursor += 1;
+  *n_fields += 1;
 
   for (int64_t i = 0; i < view->n_children; i++) {
-    ArrowIpcReaderInitFields(cursor, view->children[i], schema->children[i],
-                             array->children[i], n_buffers);
+    ArrowIpcReaderInitFields(fields, view->children[i], array->children[i], n_fields,
+                             n_buffers);
   }
 }
 
 ArrowErrorCode ArrowIpcReaderSetSchema(struct ArrowIpcReader* reader,
                                        struct ArrowSchema* schema,
                                        struct ArrowError* error) {
-  // Clear previous schema, array, array_view, buffers, and fields
-  if (reader->schema.release != NULL) {
-    reader->schema.release(&reader->schema);
-  }
-
   if (reader->array.release != NULL) {
     reader->array.release(&reader->array);
   }
@@ -890,15 +884,13 @@ ArrowErrorCode ArrowIpcReaderSetSchema(struct ArrowIpcReader* reader,
     ArrowErrorSet(error, "Failed to allocate reader->buffers");
     return ENOMEM;
   }
-  memset(reader->buffers, 0, reader->n_buffers * sizeof(struct ArrowBuffer));
-
-  // If all is well, take ownership of the schema
-  ArrowSchemaMove(schema, &reader->schema);
+  memset(reader->buffers, 0, reader->n_buffers * sizeof(struct ArrowBufferView));
 
   // Init field information and calculate starting buffer offset for each
-  struct ArrowIpcField** cursor = &reader->fields;
+  int64_t n_fields = 0;
   int64_t n_buffers = 0;
-  ArrowIpcReaderInitFields(cursor, &reader->array_view, schema, &reader->array, &n_buffers);
+  ArrowIpcReaderInitFields(reader->fields, &reader->array_view, &reader->array, &n_fields,
+                           &n_buffers);
 
   return NANOARROW_OK;
 }
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
@@ -828,8 +828,8 @@ static void ArrowIpcReaderInitFields(struct ArrowIpcField** cursor,
   *cursor += 1;
 
   for (int64_t i = 0; i < view->n_children; i++) {
-    ArrowIpcReaderInitFields(cursor, view->children[i], schema->children[i], array->children[i],
-                             n_buffers);
+    ArrowIpcReaderInitFields(cursor, view->children[i], schema->children[i],
+                             array->children[i], n_buffers);
   }
 }
 
@@ -866,6 +866,7 @@ ArrowErrorCode ArrowIpcReaderSetSchema(struct ArrowIpcReader* reader,
   // Root must be a struct
   if (reader->array_view.storage_type != NANOARROW_TYPE_STRUCT) {
     ArrowErrorSet(error, "schema must be a struct type");
+    return EINVAL;
   }
 
   // Walk tree and calculate how many fields we need to allocate
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
@@ -45,6 +45,22 @@ void ArrowIpcReaderReset(struct ArrowIpcReader* reader) {
     reader->schema.release(&reader->schema);
   }
 
+  if (reader->array.release != NULL) {
+    reader->array.release(&reader->array);
+  }
+
+  ArrowArrayViewReset(&reader->array_view);
+
+  if (reader->fields != NULL) {
+    ArrowFree(reader->fields);
+    reader->n_fields = 0;
+  }
+
+  if (reader->buffers != NULL) {
+    ArrowFree(reader->buffers);
+    reader->n_buffers = 0;
+  }
+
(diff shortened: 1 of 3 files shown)


extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
```diff
@@ -168,6 +204,151 @@ TEST(NanoarrowIpcTest, NanoarrowIpcDecodeSimpleSchema) {
...
+  data.data.as_uint8 = kSimpleRecordBatch;
+  data.size_bytes = sizeof(kSimpleRecordBatch);
+
+  ArrowIpcReaderInit(&reader);
+
+  // Attempt to get array should fail nicely here
+  EXPECT_EQ(ArrowIpcReaderGetArray(&reader, data, 0, nullptr, &error), EINVAL);
+  EXPECT_STREQ(error.message, "reader did not just decode a RecordBatch message");
+
+  ASSERT_EQ(ArrowIpcReaderSetSchema(&reader, &schema, nullptr), NANOARROW_OK);
```
lidavidm, line 235: Is the assumption that the caller is responsible for all I/O, so it has to decode the schema and feed it back in, and we can build up a higher level reader on top of this interface?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
```diff
@@ -168,6 +204,151 @@ TEST(NanoarrowIpcTest, NanoarrowIpcDecodeSimpleSchema) {
...
+  ASSERT_EQ(array.children[0]->length, 3);
+  EXPECT_EQ(array.children[0]->null_count, 0);
+  const int32_t* out = reinterpret_cast<const int32_t*>(array.children[0]->buffers[1]);
+  EXPECT_EQ(out[0], 1);
+  EXPECT_EQ(out[1], 2);
+  EXPECT_EQ(out[2], 3);
+
+  array.release(&array);
+
+  // Check field extract
```
lidavidm, line 264: Is the intent to map this down to a partial read later on? Otherwise, this feels like it just duplicates Nanoarrow's APIs

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
```diff
@@ -205,6 +386,74 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcArrowTypeRoundtrip) {
...
+  ASSERT_EQ(ArrowIpcReaderGetArray(&reader, buffer_view, -1, &array, nullptr),
+            NANOARROW_OK);
+
+  maybe_batch = arrow::ImportRecordBatch(&array, dummy_schema);
+  ASSERT_TRUE(maybe_batch.ok());
+  EXPECT_EQ(maybe_batch.ValueUnsafe()->ToString(), nulls->ToString());
+  EXPECT_TRUE(maybe_batch.ValueUnsafe()->Equals(*nulls));
+
+  if (!maybe_batch.ValueUnsafe()->Equals(*nulls)) {
+    std::cout << "something";
```
lidavidm, line 450: `FAIL() << "something descriptive goes here";`?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
lidavidm: Ah, I see in the actual implementation it's intended to save a bit of deserialization work

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
```diff
@@ -785,5 +847,252 @@ ArrowErrorCode ArrowIpcReaderDecode(struct ArrowIpcReader* reader,
...
+  if (setter->endianness != NANOARROW_IPC_ENDIANNESS_UNINITIALIZED &&
+      setter->endianness != setter->system_endianness) {
+    ArrowErrorSet(error,
+                  "The nanoarrow_ipc extension does not support non-system endianness");
+    return ENOTSUP;
+  }
+
+  int result = ArrowBufferAppendBufferView(out, view);
+  if (result != NANOARROW_OK) {
+    ArrowErrorSet(error, "Failed to copy buffer");
```
lidavidm, line 975: Ideally we wouldn't have to copy data right?

lidavidm: (I suppose you'd want it under user control: if you only want one buffer out of many, then copying might let you save memory later on. But then the interface is a little wonky, because you have to provide the full record batch up front, implying you already did all the I/O, unless you have memory mapping available.)

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
```diff
@@ -785,5 +847,252 @@ ArrowErrorCode ArrowIpcReaderDecode(struct ArrowIpcReader* reader,
...
+    }
+  } else {
+    result = ArrowIpcReaderWalkGetArray(&setter, &temp, error);
+    if (result != NANOARROW_OK) {
+      temp.release(&temp);
+      return result;
+    }
+  }
+
+  // TODO: this performs some validation but doesn't do everything we need it to do
```
lidavidm, line 1087: We might defer that to a ValidateFull-style function in core nanoarrow itself?

lidavidm (approved):
LGTM overall, though some brief docstrings might help figure out how to use the interface

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -80,8 +129,33 @@ ArrowErrorCode ArrowIpcReaderDecode(struct ArrowIpcReader* reader,
...
+ArrowErrorCode ArrowIpcReaderSetSchema(struct ArrowIpcReader* reader,
+                                       struct ArrowSchema* schema,
+                                       struct ArrowError* error);
+
+struct ArrowIpcField {
+  struct ArrowArrayView* array_view;
+  int64_t buffer_offset;
+};
+
+struct ArrowIpcReaderPrivate {
```
lidavidm, line 148: nit: maybe indicate that it's only here for testing?

(alternatively, you could keep it in the .c file and copy-paste it to the .cc file in an 'extern C' block)

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
```diff
@@ -785,5 +876,261 @@ ArrowErrorCode ArrowIpcReaderDecode(struct ArrowIpcReader* reader,
...
+                                       struct ArrowError* error) {
+  struct ArrowIpcReaderPrivate* private_data =
+      (struct ArrowIpcReaderPrivate*)reader->private_data;
+
+  if (private_data->schema.release == NULL) {
+    ArrowErrorSet(error, "reader does not contain a valid schema");
+    return EINVAL;
+  }
+
+  ArrowSchemaMove(&private_data->schema, out);
```
lidavidm, line 894: Do we really want Move here, or a copy?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
lidavidm: ping

lidavidm (commented):
Generally LGTM but I have some questions still

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
paleolimbot: Yeah, this should be extensible somehow...a `struct ArrowIpcBufferDecoder` or something. It also needs a way to inject compression support at runtime. I'll do both of those in a follow-up (this PR is mostly about ensuring type coverage and correctness of the flattening/unflattening).

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
paleolimbot: Definitely!

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
paleolimbot: I think so...what's there now might be more accurately be called an `ArrowIpcDecoder`. There are so many ways to mix, match, and parallelize the various steps here based on the tools available to the caller. This library should probably provide a reader for the simple but common "just read it all as an ArrowArrayStream" (which would at least make sure that it's possible to do).

paleolimbot (commented):
I ran out of time yesterday to finish all the things here...the ability to inject buffer reading is probably better suited to another PR, as is building a simple reader that takes care of IO.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.c
paleolimbot: Until there's a code path where the decoder needs a copy of the schema for something, probably a move? Strictly speaking, the decoder just needs to "see" the schema once and doesn't need to refer to it again.

lidavidm: It does break expectations for Get though; usually you'd expect that to be idempotent

lidavidm (approved):

PR #160, opened 2023-03-14
feat(ci): Add suite of Docker-based release verification script tests

paleolimbot:


.github/workflows/verify.yaml
```diff
@@ -0,0 +1,157 @@
...
+          path: src
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Run docker-compose verify
+        env:
+          PLATFORM: ${{ matrix.config.platform }}
+        run: |
+          cd src
+          DOCKER_DEFAULT_PLATFORM=${{ matrix.config.docker_platform }} docker-compose run verify
```
paleolimbot, line 157: @assignUser Is there any way to cache the docker image builds here? The verification itself takes almost no time at all but installing linux/R packages takes a few minutes.

assignUser: You could build the images and push them to the github container registry (which would be possible with the normal $GITHUB_TOKEN and require no additional auth). 

I see you are using arrow release version so the main concern for "stale" images would be R package version updates? With the current dockerfiles there is no way for docker to know if the packages updated so we would need something like a regular (weekly?) rebuild of the images.

paleolimbot: I'm not all that worried about the images going stale because of R package versions...that case is handled by the non-docker verification that uses the standard r-lib/actions stuff. A weekly rebuild would probably be sufficient if there's an easy way to specify that?

assignUser: I'll setup a workflow for you.

PR #162, opened 2023-03-17
fix(extensions/nanoarrow_ipc): Ensure tests pass on big endian

paleolimbot:
This PR tweaks the tests and endianness setup so that the tests pass on big endian. Buffer decoding on non-platform endian isn't supported yet...this PR is just about making sure all of the other infrastructure around the buffers works. This is also in preparation for adding the nanoarrow_ipc extension to release verification (which now runs on big endian on CI).

You can test this using the shiny new alpine-s390x image in this repo's container store!

```bash
# git clone https://github.com/paleolimbot/arrow-nanoarrow.git && cd arrow-nanoarrow && git switch ipc-improvements
# docker run --rm -it ghcr.io/apache/arrow-nanoarrow:alpine-s390x
mkdir build && cd build
cmake ..
cmake --build .
ctest
```

### DIFF
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
@@ -23,6 +23,9 @@
 #include <arrow/util/key_value_metadata.h>
 #include <gtest/gtest.h>
 
+// For bswap32()
+#include "flatcc/portable/pendian.h"
+
 #include "nanoarrow_ipc.h"
 
 using namespace arrow;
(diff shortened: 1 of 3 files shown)


extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
```diff
@@ -44,6 +49,18 @@ struct ArrowIpcDecoderPrivate {
 };
 }
 
+static enum ArrowIpcEndianness ArrowIpcSystemEndianness(void) {
+  uint32_t check = 1;
+  char first_byte;
+  enum ArrowIpcEndianness system_endianness;
```
lidavidm, line 55: unused?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_test.cc
```diff
@@ -272,9 +297,16 @@ TEST(NanoarrowIpcTest, NanoarrowIpcDecodeSimpleRecordBatch) {
...
-  EXPECT_EQ(out[0], 1);
-  EXPECT_EQ(out[1], 2);
-  EXPECT_EQ(out[2], 3);
+
+  if (ArrowIpcSystemEndianness() == NANOARROW_IPC_ENDIANNESS_LITTLE) {
+    EXPECT_EQ(out[0], 1);
+    EXPECT_EQ(out[1], 2);
+    EXPECT_EQ(out[2], 3);
+  } else {
+    EXPECT_EQ(out[0], bswap32(1));
```
lidavidm, line 306: Hmm, because the hardcoded data is in little-endian?

Wonder if it would instead make sense to compare byte slices directly instead of int32_t values then

lidavidm (approved):

lidavidm (approved):

PR #164, opened 2023-03-20
feat(extensions/nanoarrow_ipc): Add single-threaded stream reader

paleolimbot:
Higher level runtimes may be able to use the `ArrowIpcDecoder` (or more than one) and handle IO/parallelization using tools that are difficult to provide from C; however, for testing we do need the ability to read streams in their entirety. This PR provides a tool to do that based on an arbitrary bytes input.

This reduces the overhead of coordinating the various steps required to decode a stream to:

```c
struct ArrowIpcInputStream input;
ArrowIpcInputStreamInit<Buffer|File|Custom Implementation>(&input, ...);

struct ArrowArrayStream stream;
ArrowIpcArrayStreamReaderInit(&stream, &input, nullptr);

struct ArrowSchema schema;
stream.get_schema(&stream, &schema);

struct ArrowArray array;
while (1) {
  stream.get_next(&stream, &array);

  if (array.release != NULL) {
    array.release(&array);
  } else {
    break;
  }
}

schema.release(&schema);
stream.release(&stream);
```

There is also a utility to read an entire stream from file or stdin:

```bash
$ ./dump_stream big.arrows 
Read Schema <0.000122 seconds>
struct
  state: string
  geometry: geoarrow.polygon{list}
    rings: list
      vertices: geoarrow.point{fixed_size_list(2)}
        xy: double
Read 23548499 rows in 1 batch(es) <1.623532 seconds>
```


extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -219,6 +227,60 @@ ArrowErrorCode ArrowIpcDecoderDecodeArray(struct ArrowIpcDecoder* decoder,
                                           struct ArrowArray* out,
                                           struct ArrowError* error);
 
+/// \brief An user-extensible input data source
+struct ArrowIpcInputStream {
+  /// \brief Read up to buf_size_bytes from stream into buf
+  ///
+  /// The actual number of bytes read is placed in the value pointed to by
+  /// size_read_out. Returns NANOARROW_OK on success.
+  ArrowErrorCode (*read)(struct ArrowIpcInputStream* stream, void* buf,
```
lidavidm, line 236: I suppose:

- Perhaps consider a concrete type (char*, uint8_t*) for buf so it's harder to accidentally pass an arbitrary pointer to it?
- What do you think about `char** buf` (i.e., letting the stream do the allocation)? That would let you do things like read an in-memory buffer by slicing instead of copying (though I suppose lifetimes may get confusing there).

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_reader.c
```diff
@@ -0,0 +1,424 @@
...
+  struct ArrowBuffer input;
+  int64_t cursor_bytes;
+};
+
+static ArrowErrorCode ArrowIpcInputStreamBufferRead(struct ArrowIpcInputStream* stream,
+                                                    void* buf, int64_t buf_size_bytes,
+                                                    int64_t* size_read_out,
+                                                    struct ArrowError* error) {
+  if (buf_size_bytes == 0) {
+    return NANOARROW_OK;
```
lidavidm, line 41: nit: possibly assign to `size_read_out` anyways?

lidavidm (approved):

lidavidm (approved):

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc.h
paleolimbot: > Perhaps consider a concrete type (char*, uint8_t*) for buf

Totally!

> What do you think about char** buf (i.e., letting the stream do the allocation)?

I played with changing it to `struct ArrowBuffer*` but it's complicated...probably the way to go about that would be to do what Arrow C++ does and make it a separate method (I think Arrow's is called `ReadAt()` or something? If somebody really wants slicing of an in-memory buffer they can use the `ArrowIpcDecoder` directly too.

PR #165, opened 2023-03-23
feat(extensions/nanoarrow_ipc): Allow shared buffers for zero-copy buffer decode

paleolimbot:
This PR makes buffer construction more flexible, with the initial target being zero-copy shared references to a single `struct ArrowBuffer` containing the read-in message body. Where supported (i.e., C11 and up), these shared buffer references are thread safe.

I implemented this by abstracting out a `struct ArrowIpcBufferFactory`. This could be exposed to the user, too (but maybe in another PR where I also implement endian swapping).

Using the decoder API, you can request this using a new function:

```c
ArrowErrorCode ArrowIpcDecoderDecodeArrayFromOwned(struct ArrowIpcDecoder* decoder,
                                                   struct ArrowBuffer* body, int64_t i,
                                                   struct ArrowArray* out,
                                                   struct ArrowError* error);
```

Using the `ArrowArrayStream` interface this is the default *if* shared buffers are thread safe (this can be overridden).

### DIFF
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
@@ -1147,9 +1147,21 @@ static struct ArrowIpcBufferFactory ArrowIpcBufferFactoryFromView(
   return out;
 }
 
+static ArrowErrorCode ArrowIpcMakeBufferFromShared(struct ArrowIpcBufferFactory* factory,
+                                                   struct ArrowIpcBufferSource* src,
+                                                   struct ArrowBuffer* dst,
+                                                   struct ArrowError* error) {
+  struct ArrowBuffer* shared = (struct ArrowBuffer*)factory->private_data;
+  ArrowIpcSharedBufferClone(shared, dst);
+  dst->data += src->body_offset_bytes;
+  dst->size_bytes = src->buffer_length_bytes;
+  return NANOARROW_OK;
+}
+
 static struct ArrowIpcBufferFactory ArrowIpcBufferFactoryFromShared(
     struct ArrowBuffer* shared) {
   struct ArrowIpcBufferFactory out;
+  out.make_buffer = &ArrowIpcMakeBufferFromShared;
   out.private_data = shared;
   return out;
 }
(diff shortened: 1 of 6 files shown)


extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
```diff
@@ -1051,53 +1064,184 @@ ArrowErrorCode ArrowIpcDecoderSetEndianness(struct ArrowIpcDecoder* decoder,
...
+}
 
-  // Check that this buffer fits within the body
-  if (offset < 0 || (offset + length) > setter->body.size_bytes) {
-    ArrowErrorSet(error,
-                  "Buffer %ld requires body offsets [%ld..%ld) but body has size %ld",
-                  (long)setter->buffer_i - 1, (long)offset, (long)offset + (long)length,
-                  setter->body.size_bytes);
-    return EINVAL;
+static int ArrowIpcSharedBufferInit(struct ArrowBuffer* buffer) {
```
lidavidm, line 1120: Would it be possible to check if buffer is already a shared buffer, and if so just reuse the existing reference count?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
```diff
@@ -1051,53 +1064,184 @@ ArrowErrorCode ArrowIpcDecoderSetEndianness(struct ArrowIpcDecoder* decoder,
...
+}
+
+struct ArrowIpcBufferSource {
+  int64_t body_offset_bytes;
+  int64_t buffer_length_bytes;
+  enum ArrowIpcCompressionType codec;
+  int swap_endian;
+};
+
+struct ArrowIpcBufferFactory {
```
lidavidm, line 1155: This might be worth some docs? Basically, it's handing out slices/copies of some internal source buffer based on the `src` parameter?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
lidavidm: Especially if `DecodeArrayFromOwned` intends to replace `body` with a shared buffer so you can reuse it, then it sounds like we need to be careful here to avoid multiple reference counts pointing to the same source data.

paleolimbot: I updated this so that there's a `struct ArrowIpcSharedBuffer` that the caller has to initialize first. Technically it worked before and was in the tests (it created a tree of shared buffers) but was confusing an unnecessary as you noted.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
paleolimbot: I added full documentation here...the basic idea is that what we get from flatbuffers is offset + length as integers, but there are a multitude of ways to get from there to a `struct ArrowBuffer` (not all of which involve absolute pointers or a complete message body that exists in memory). Because it's internal it could also be a `switch()` + `enum` if you think the callback thing is overkill 🤷

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
lidavidm: Right...I realized that later :sweat_smile: But yeah, why have a tree of pointers when it's not needed. Thanks!

lidavidm (approved):

PR #169, opened 2023-03-27
chore: Add IPC extension and ARM64 to verification script

paleolimbot:
This PR:

- Adds the IPC extension to the release verification
- Adds a memcheck option to release verification (off by default)
- Adds the IPC extension to the source tarball now that it's being verified
- Adds arm64, memcheck, and centos7 to the CI configuration that runs the verification script (with bundling checks turned off for anything that runs on qemu)

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -179,4 +179,5 @@ jobs:
 
         run: |
           cd src
+          echo "::group::Docker Pull"
           docker compose run -e GITHUB_ACTIONS ${{ matrix.config.compose_args }} verify
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -170,4 +170,5 @@ jobs:
 
         run: |
           cd src
-          docker compose ${{ matrix.config.compose_args }} run verify
+          # Set GITHUB_WORKSPACE to set more log-friendly headers
+          docker compose -e GITHUB_WORKSPACE=1 ${{ matrix.config.compose_args }} run verify
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -145,7 +145,8 @@ jobs:
           - {platform: "fedora", arch: "amd64"}
           - {platform: "archlinux", arch: "amd64"}
           - {platform: "alpine", arch: "amd64"}
-          - {platform: "ubuntu", arch: "amd64", extra_label: "-memcheck", extra_env: "TEST_WITH_MEMCHECK=1"}
+          - {platform: "ubuntu", arch: "amd64",
+             extra_label: "-memcheck", extra_env: "TEST_WITH_MEMCHECK=1 TEST_C_BUNDLED=0"}
           - {platform: "ubuntu", arch: "arm64"}
           - {platform: "alpine", arch: "s390x"}
 
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -142,10 +142,11 @@ jobs:
       matrix:
         config:
           - {platform: "ubuntu", arch: "amd64"}
-          - {platform: "ubuntu", arch: "arm64"}
           - {platform: "fedora", arch: "amd64"}
           - {platform: "archlinux", arch: "amd64"}
           - {platform: "alpine", arch: "amd64"}
+          - {platform: "ubuntu", arch: "amd64", extra_env: "TEST_WITH_MEMCHECK=1"}
+          - {platform: "ubuntu", arch: "arm64"}
           - {platform: "alpine", arch: "s390x"}
 
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/verify.yaml
```diff
@@ -166,5 +173,5 @@ jobs:
 
         run: |
           cd src
-          # docker compose pull verify
-          docker compose run verify
+          # Set GITHUB_WORKSPACE to set more log-friendly headers
```
assignUser, line 176: I am not sure what this envvar is supposed to do/how. If you want to use a var to detect the execution within GHA I would recommend using `GITHUB_ACTIONS` as this is a var that is set by GHA for this purpose vs `GITHUB_WORKSPACE` which usually contains the [workspace path](https://docs.github.com/en/actions/learn-github-actions/variables),

extensions/nanoarrow_ipc/CMakeLists.txt
```diff
@@ -16,9 +16,13 @@
 # under the License.
 
 message(STATUS "Building using CMake version: ${CMAKE_VERSION}")
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.14)
 include(FetchContent)
 
+if(NOT DEFINED CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 11)
```
assignUser, line 23: Fun fact, this actually does not enforce the usage of C++11 but allows a compiler to decay to an earlier version.
See https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html

extensions/nanoarrow_ipc/CMakeLists.txt
```diff
@@ -181,11 +185,24 @@ if (NANOARROW_IPC_BUILD_TESTS)
...
+  if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.23")
+      cmake_policy(SET CMP0135 NEW)
+  endif()
+
+  # Use an old version of googletest if we have to to support gcc 4.8
+  if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR
+    CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "5.0.0")
+    FetchContent_Declare(
+        googletest
+        URL https://github.com/google/googletest/archive/release-1.11.0.zip
```
assignUser, line 197: I'd extract the version into a var and set that in the if to avoid duplicating the whole FC_declare call.

assignUser: also checksums maybe?

extensions/nanoarrow_ipc/CMakeLists.txt
```diff
@@ -199,8 +216,20 @@ if (NANOARROW_IPC_BUILD_TESTS)
...
 
-  target_link_libraries(nanoarrow_ipc_decoder_test nanoarrow_ipc nanoarrow arrow_shared gtest_main)
-  target_link_libraries(nanoarrow_ipc_reader_test nanoarrow_ipc nanoarrow gtest_main)
+  # On Windows, dynamically linking Arrow is difficult to get right,
+  # so use static linking by default.
+  if (MSVC)
+      set(NANOARROW_IPC_ARROW_TARGET arrow_static)
+  else()
+      set(NANOARROW_IPC_ARROW_TARGET arrow_shared)
+  endif()
```
assignUser, line 225: This could also be done with a generator expression something like (untested) `set(NANOARROW_IPC_ARROW_TARGET $<$<${MSVC}>:arrow_static>$<$<NOT:$<${MSVC}>>:arrow_shared>)` but I am not really convinced that it is more readable :D

assignUser (commented):
Some minor comments :)

extensions/nanoarrow_ipc/CMakeLists.txt
paleolimbot: We don't actually need C11 here but it's preferred since it gets us thread-safe shared buffers. We do need to specify a C standard if unset, though, because some compilers default to pre C99 and that won't fly. I should probably add an explicit test for what happens when C99 is set, though (maybe a future PR though).

extensions/nanoarrow_ipc/CMakeLists.txt
paleolimbot: But it does fit on one line! I left this one as is because I don't understand it yet (but I am glad to know it exists!)

extensions/nanoarrow_ipc/CMakeLists.txt
paleolimbot: Good call!

.github/workflows/verify.yaml
paleolimbot: Good call! I'm using it to make the logs look nicer 🙂

PR #174, opened 2023-03-28
feat: Implement `ArrowArrayViewValidateFull()`

paleolimbot:
This is required for IPC reading because corrupted offset and/or union type ID buffers could result in consumers accessing out-of-bounds elements. `ArrowArrayViewSetArray()` already checked the *last* element of offset buffers against lengths but didn't check the first element and didn't check for negative sequential offsets.

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -677,7 +677,7 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
       if (array_view->buffer_views[1].size_bytes != 0) {
         first_offset = array_view->buffer_views[1].data.as_int32[0];
         if (first_offset < 0) {
-          ArrowErrorSet(error, "Expected first offset >=0 but found %ld",
+          ArrowErrorSet(error, "Expected first offset >= 0 but found %ld",
                         (long)first_offset);
           return EINVAL;
         }
@@ -692,7 +692,7 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
       if (array_view->buffer_views[1].size_bytes != 0) {
         first_offset = array_view->buffer_views[1].data.as_int64[0];
         if (first_offset < 0) {
-          ArrowErrorSet(error, "Expected first offset >=0 but found %ld",
+          ArrowErrorSet(error, "Expected first offset >= 0 but found %ld",
                         (long)first_offset);
           return EINVAL;
         }
@@ -728,7 +728,7 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
       if (array_view->buffer_views[1].size_bytes != 0) {
         first_offset = array_view->buffer_views[1].data.as_int32[0];
         if (first_offset < 0) {
-          ArrowErrorSet(error, "Expected first offset >0 but found %ld",
+          ArrowErrorSet(error, "Expected first offset >= 0 but found %ld",
                         (long)first_offset);
           return EINVAL;
         }
@@ -757,7 +757,7 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
       if (array_view->buffer_views[1].size_bytes != 0) {
         first_offset = array_view->buffer_views[1].data.as_int64[0];
         if (first_offset < 0) {
-          ArrowErrorSet(error, "Expected first offset >0 but found %ld",
+          ArrowErrorSet(error, "Expected first offset >= 0 but found %ld",
                         (long)first_offset);
           return EINVAL;
         }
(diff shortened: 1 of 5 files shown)


src/nanoarrow/array.c
```diff
@@ -758,3 +804,138 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
...
+  if (view.size_bytes <= (int64_t)sizeof(int32_t)) {
+    return NANOARROW_OK;
+  }
+
+  for (int64_t i = 1; i < view.size_bytes / (int64_t)sizeof(int32_t); i++) {
+    int32_t diff = view.data.as_int32[i] - view.data.as_int32[i - 1];
+    if (diff < 0) {
+      ArrowErrorSet(error,
+                    "Expected element size >0 but found element size %ld at position %ld",
+                    (long)diff, (long)i);
```
lidavidm, line 819: nit: the message is kind of a non-sequitur for the error case here

lidavidm (commented):
LGTM, just a couple questions

src/nanoarrow/nanoarrow_types.h
```diff
@@ -503,7 +503,8 @@ struct ArrowArrayView {
   ///
   /// If storage_type is a union type, a 256-byte ArrowMalloc()ed buffer
   /// such that child_index == union_type_id_map[type_id] and
-  /// type_id == union_type_id_map[128 + child_index]
+  /// type_id == union_type_id_map[128 + child_index]. This value may be
+  /// NULL in the case where child_id == type_id.
```
wjones127, line 507: Is that part of the Arrow format? I couldn't find such a detail.

src/nanoarrow/array.c
```diff
@@ -648,15 +663,25 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
...
   // Check child sizes and calculate sizes that depend on data in the array buffers
+  int64_t first_offset;
   int64_t last_offset;
   switch (array_view->storage_type) {
     case NANOARROW_TYPE_STRING:
     case NANOARROW_TYPE_BINARY:
       if (array_view->buffer_views[1].size_bytes != 0) {
+        first_offset = array_view->buffer_views[1].data.as_int32[0];
+        if (first_offset < 0) {
+          ArrowErrorSet(error, "Expected first offset >0 but found %ld",
```
wjones127, line 680:
```suggestion
          ArrowErrorSet(error, "Expected first offset >=0 but found %ld",
```

src/nanoarrow/array.c
```diff
@@ -665,6 +690,13 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
     case NANOARROW_TYPE_LARGE_STRING:
     case NANOARROW_TYPE_LARGE_BINARY:
       if (array_view->buffer_views[1].size_bytes != 0) {
+        first_offset = array_view->buffer_views[1].data.as_int64[0];
+        if (first_offset < 0) {
+          ArrowErrorSet(error, "Expected first offset >0 but found %ld",
```
wjones127, line 695:
```suggestion
          ArrowErrorSet(error, "Expected first offset >=0 but found %ld",
```

wjones127: And similarly in other places

src/nanoarrow/array.c
```diff
@@ -758,3 +804,138 @@ ArrowErrorCode ArrowArrayViewSetArray(struct ArrowArrayView* array_view,
...
+    }
+  }
+
+  return NANOARROW_OK;
+}
+
+static int ArrowAssertInt8In(struct ArrowBufferView view, const int8_t* values,
+                             int64_t n_values, struct ArrowError* error) {
+  for (int64_t i = 0; i < view.size_bytes; i++) {
+    int item_found = 0;
```
wjones127, line 864: Question: why `int` over `bool`?

wjones127 (changes requested):
Just a few small nits and questions.

src/nanoarrow/array.c
lidavidm: There's not really bool in C (there is _Bool, I suppose with C99 we can use that and I guess we're assuming C99 here?)

wjones127: 😮

src/nanoarrow/nanoarrow_types.h
paleolimbot: This is an implementation detail of the `ArrowArrayView`...in the spec the type ids are always there. Arguably they should always be here, too, but there's at least one test that exploits that behaviour that I discovered so I figured I should probably document it. I think the reason it ended up that way was because you can `ArrowArrayViewInitFromType()` and `ArrowArrayViewAllocateChildren()` and letting the type id map stay NULL saves some special casing of unions there.

src/nanoarrow/array.c
paleolimbot: I took another pass at making them all a bit better!

src/nanoarrow/array.c
paleolimbot: There's `#include <stdbool.h>` too, which defines macros for `bool`, `true`, and `false`. I didn't use it initially because there are some functions that return true/false in public headers and I wanted as few includes there as possible. I should probably go through and make it consistent (I may have used `char` in some places).

wjones127 (approved):
Looks good!

PR #175, opened 2023-03-29
feat!: Allow explicit validation level in `ArrowArrayFinishBuilding()`

paleolimbot:
This is a breaking change! The signature of `ArrowArrayFinishBuilding()` gains a `validation_level` argument, and the original function was renamed to `ArrowArrayFinishBuildingDefault()`. I could do this in a non-breaking way but there are no non-awkward names for `ArrowArrayFinishBuildingWithExplicitValidationLevel()` and it seems more important to cultivate a quality API at this point rather than maintain full backward compatibility (plus it's a relatively easy rename fix).

This adds two important features. First, it lets you preform more validation if you want to (via `NANOARROW_VALIDATION_LEVEL_FULL`, implemented with the just-added `ArrowArrayViewValidateFull()`). Second, it lets you perform less validation if you want to (e.g., if your buffers are sitting on a GPU and dereferencing a buffer value will crash).

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -435,13 +435,19 @@ static ArrowErrorCode ArrowArrayCheckInternalBufferSizes(
 }
 
 ArrowErrorCode ArrowArrayFinishBuilding(struct ArrowArray* array,
+                                        enum ArrowValidationLevel validation_level,
                                         struct ArrowError* error) {
-  // Even if the data buffer is size zero, the value needs to be non-null
+  // Even if the data buffer is size zero, the value needed to be non-null
+  // in Arrow C++ < 11.0.0.
   NANOARROW_RETURN_NOT_OK(ArrowArrayFinalizeBuffers(array));
 
   // Make sure the value we get with array->buffers[i] is set to the actual
   // pointer (which may have changed from the original due to reallocation)
   ArrowArrayFlushInternalPointers(array);
+
+  if (validation_level == NANOARROW_VALIDATION_LEVEL_NONE) {
+    return NANOARROW_OK;
+  }
 
   // Check buffer sizes to make sure we are not sending an ArrowArray
   // into the wild that is going to segfault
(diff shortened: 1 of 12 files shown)


src/nanoarrow/nanoarrow_types.h
```diff
@@ -319,6 +319,25 @@ enum ArrowTimeUnit {
...
 };
 
+/// \brief Validation level enumerator
+/// \ingroup nanoarrow-array
+enum ArrowValidationLevel {
+  /// \brief Do not validate buffer sizes or content.
+  NANOARROW_VALIDATION_LEVEL_NONE = 0,
+
+  /// \brief Validate buffer sizes that depend on array length but do not validate buffer
+  /// sizes that depend on buffer data access.
```
lidavidm, line 329: In other words, we never actually access any buffers right?

src/nanoarrow/array.c
```diff
@@ -458,17 +467,38 @@ ArrowErrorCode ArrowArrayFinishBuilding(struct ArrowArray* array,
...
+    ArrowArrayViewReset(&array_view);
+    return result;
+  }
+
+  if (validation_level == NANOARROW_VALIDATION_LEVEL_DEFAULT) {
+    ArrowArrayViewReset(&array_view);
+    return NANOARROW_OK;
+  }
+
+  result = ArrowArrayViewValidateFull(&array_view, error);
```
lidavidm, line 492: Hmm, if validation fails, should we free the array too?

lidavidm: Either way, the behavior should be documented

lidavidm (approved):

src/nanoarrow/array.c
paleolimbot: I added a bit to the group documentation for arrays...I don't think freeing the array in that method is a good idea because no other ArrowArrayXXX functions do this either. If surfacing this to a higher level runtime (e.g., Python), the error message could theoretically be used to fix the error and try again, too.

src/nanoarrow/nanoarrow_types.h
paleolimbot: Yes, we don't access buffers with "minimal" validation. I added a basic test for this, too.

PR #176, opened 2023-03-29
chore(extensions/nanoarrow_ipc): Test files from the arrow-testing repo

paleolimbot:
TODO:

- CI
- Fuzz-tested files


src/nanoarrow/array.c
```diff
@@ -910,10 +910,6 @@ ArrowErrorCode ArrowArrayViewValidateFull(struct ArrowArrayView* array_view,
                                           struct ArrowError* error) {
   for (int i = 0; i < 3; i++) {
     switch (array_view->layout.buffer_type[i]) {
-      case NANOARROW_BUFFER_TYPE_UNION_OFFSET:
```
lidavidm, line 913: Hmm, dense union offsets don't have to be increasing?

lidavidm: Ah, "The respective offsets for each child value array must be in order / increasing." So if we want to validate this it'll be trickier.

lidavidm (approved):

src/nanoarrow/array.c
paleolimbot: Yes, having this included resulted in an incorrect validation failure for the integration test file. I added this to the TODO list for #178.

PR #180, opened 2023-04-13
feat: Add Decimal Get/Set utilities

### LINKED ISSUE #170: Decimal creation utilities
Arrow specifies a particular format for decimals to be laid out, it would be nice to have a struct like Arrow's Decimal128/Decimal256 that allows us to construct them from integers and have a method `ArrowArrayAppendDecimal`.

paleolimbot:
Closes #170.

This is not designed to be a fully-featured performant decimal math library, but is designed to help with byte shuffling and provide enough support that users can write tests for decimal arrays (e.g., by providing integer getters and setters for numbers that fit within the int64 range).

Adds a `struct ArrowDecimal` and utility functions:

```c
static inline void ArrowDecimalInit(struct ArrowDecimal* decimal, int32_t bitwidth,
                                    int32_t precision, int32_t scale);
static inline int64_t ArrowDecimalGetIntUnsafe(struct ArrowDecimal* decimal);
static inline void ArrowDecimalGetBytes(struct ArrowDecimal* decimal, uint8_t* out);
static inline int64_t ArrowDecimalSign(struct ArrowDecimal* decimal);
static inline void ArrowDecimalSetInt(struct ArrowDecimal* decimal, int64_t value);
static inline void ArrowDecimalSetBytes(struct ArrowDecimal* decimal,
                                        const uint8_t* value);
```

...and adds a "getter" and an "appender":

```c
static inline ArrowErrorCode ArrowArrayAppendDecimal(struct ArrowArray* array,
                                                     struct ArrowDecimal* value);
static inline void ArrowArrayViewGetDecimalUnsafe(struct ArrowArrayView* array_view,
                                                  int64_t i, struct ArrowDecimal* out);
```

### DIFF
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -530,17 +530,28 @@ static inline ArrowErrorCode ArrowArrayAppendDecimal(struct ArrowArray* array,
       if (value->n_words != 2) {
         return EINVAL;
       } else {
-        return ArrowBufferAppend(data_buffer, value->words, 2 * sizeof(uint64_t));
+        NANOARROW_RETURN_NOT_OK(
+            ArrowBufferAppend(data_buffer, value->words, 2 * sizeof(uint64_t)));
+        break;
       }
     case NANOARROW_TYPE_DECIMAL256:
       if (value->n_words != 4) {
         return EINVAL;
       } else {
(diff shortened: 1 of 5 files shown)


wjones127 (approved):
This looks good. Very clean implementation :)

PR #189, opened 2023-05-08
Add more C++ and error handling helpers

paleolimbot:
This PR is motivated by reading some early usage of nanoarrow (including some of my own!). Basically, I just want to make it really easy to communicate failed calls to the nanoarrow API, particularly for those not used to C-style return code error handling. This PR:

- Adds `NANOARROW_THROW_NOT_OK()` in the C++ helper
- Adds `NANOARROW_ASSERT_OK()` in the C header
- Adds `NANOARROW_DEBUG` mode and inserts some more information when compiled with this define
- Adds C++ helpers for the IPC extension to mirror the C API helpers

In writing and documenting the helpers, I solidifed some guarantees...notably, that `ArrowError` has a nul-terminated error message whenever an error code is returned.


CMakeLists.txt
```diff
@@ -41,6 +41,12 @@ else()
     set(NANOARROW_NAMESPACE_DEFINE "// #define NANOARROW_NAMESPACE YourNamespaceHere")
 endif()
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(NANOARROW_DEBUG_DEFINE "#define NANOARROW_DEBUG 1")
+else()
+    set(NANOARROW_DEBUG_DEFINE "// #define NANOARROW_DEBUG 1")
+endif()
```
lidavidm, line 48: FWIW, CMake has an explicit feature for this: [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html)

lidavidm: Ah, though I suppose this is different since it'll propagate to anyone who includes nanoarrow_config.h.

lidavidm (approved):

CMakeLists.txt
paleolimbot: Thanks for this...I like the `target_compile_definitions()` solution better. Debug mode should probably never be hard-coded in a header and should probably always be invoked via a compiler define.

kou: We may want to use generator expression: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

```cmake
# We don't need to put this under if(CMAKE_BUILD_TYPE STREQUAL "Debug")
target_compile_definitions(nanoarrow PUBLIC "$<$<CONFIG:Debug>:NANOARROW_DEBUG>")
```

paleolimbot: Thanks! That is much cleaner.

lidavidm (approved):

PR #197, opened 2023-05-19
feat: Add `ArrowArrayViewListChildOffset()`

WillAyd:
Similar to what exists for Dense/Sparse Union. This would at least be useful in the ADBC library, which currently provides its own implementation

https://github.com/apache/arrow-adbc/pull/679#discussion_r1195786712


paleolimbot (approved):
Thank you! Seems reasonable and useful to me!

PR #199, opened 2023-05-21
fix: Improve readability of `ArrowArrayAllocateChildren()`

eddelbuettel:
In `ArrowArrayAllocateChildren()` two loops over all children follow each other directly.  The first assigns a NULL to each element, then second overwrites that with an allocation.  That makes the first loop effectively deadweight.  It may be there on purpose in which case you just disregard and close this PR.  But if it is an oversight this will make it ever-so-marginally faster as we can skip one loop over `n_children`.

(I noticed another microscopic style difference:  in the corresponding `ArrowSchemaAllocateChildren()` an `if (n_children != 0) {...}` wraps around the core code of function; there it is a simpler `if (n_children == 0) return ...`.  Of course they are also the same "in practice" and I don't have a style preference either -- just curious that they differ.  No need for change, but while at it you could....)

### DIFF
--- a/dist/nanoarrow.c
+++ b/dist/nanoarrow.c
@@ -1961,7 +1961,9 @@ ArrowErrorCode ArrowArrayAllocateChildren(struct ArrowArray* array, int64_t n_ch
     return ENOMEM;
   }
 
-  memset(array->children, 0, n_children * sizeof(struct ArrowArray*));
+  for (int64_t i = 0; i < n_children; i++) {
+    array->children[i] = NULL;
+  }
 
   for (int64_t i = 0; i < n_children; i++) {
     array->children[i] = (struct ArrowArray*)ArrowMalloc(sizeof(struct ArrowArray));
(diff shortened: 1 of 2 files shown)


paleolimbot (commented):
Thank you for filing this!

I believe that initializer is there because if one of the memory allocations fail, when the caller calls the release callback there will be wild pointers and the process will crash. There should, of course, be a comment stating that or perhaps using `memset()` to more efficiently zero out the content of the preceding malloc.

eddelbuettel:
Oh, I see now.  You are correct -- _if_ the second loop does not complete, _then_ we could get bitten and the first pass avoids that  My bad.   Maybe just add a comment?  (And again the sibbling routine has the `memset()` which I saw, didn't fully grok, and then didn't put two and two togther.  So may a `memset()` call is best.  Could amend PR if you want me to.)

eddelbuettel:
I updated the PR accordingly.  But feel free to discard and implement differently if you prefer; my key issue was that I did not clearly grok why this was happening.

eddelbuettel:
Gotcha on the generation, should have look harder at all the helper / builder scripts.  Thanks for the cleanup and prompt merge.

PR #201, opened 2023-05-23
refactor: Unify `ArrowArrayView` and `ArrowArray` validation

paleolimbot:
The initial motivation for this refactor was to make it possible to go straight from Arrow IPC to ArrowArrayView with zero heap allocations and improve IPC validation; however, the changes needed for that (basically, give every ArrowArrayView its own offset, length, and null_count) allowed unifying the validation that we do when building arrays and the validation we do when wrapping a foregin ArrowArray in an ArrowArrayView. This PR ensures that no internal ArrowArrayView functions actually refer to the `array` member since there is no guarantee that an ArrowArray backs the buffers. Along the way, error messages and documentation were improved.

### DIFF
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -654,7 +654,7 @@ static inline void ArrowArrayViewMove(struct ArrowArrayView* src,
 
 static inline int8_t ArrowArrayViewIsNull(struct ArrowArrayView* array_view, int64_t i) {
   const uint8_t* validity_buffer = array_view->buffer_views[0].data.as_uint8;
-  i += array_view->array->offset;
+  i += array_view->offset;
   switch (array_view->storage_type) {
     case NANOARROW_TYPE_NA:
       return 0x01;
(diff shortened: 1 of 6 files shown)


src/nanoarrow/array_inline.h
```diff
@@ -654,7 +654,7 @@ static inline void ArrowArrayViewMove(struct ArrowArrayView* src,
 
 static inline int8_t ArrowArrayViewIsNull(struct ArrowArrayView* array_view, int64_t i) {
   const uint8_t* validity_buffer = array_view->buffer_views[0].data.as_uint8;
-  i += array_view->array->offset;
+  i += array_view->offset;
```
lidavidm, line 657: From the docstring, shouldn't we _also_ add ArrowArray->offset if the array is set?

lidavidm:
I think otherwise this LGTM, I'd just like to clarify the definition of offset here

lidavidm (approved):

lidavidm:
Ah ok - so it's to be used _instead_ of the array offset, or it's "absolute" (which is easier to consume by far!)

PR #211, opened 2023-06-07
feat(r): Add callback-based async option for calling `ArrowArrayStream::get_next()`

paleolimbot:
Total proof-of-concept of one way to go about async `stream$get_next()`, as discussed briefly with @krlmlr, @lidavidm, and @nbenn in connection with ADBC in R. When the stream is wrapping a database result set, the `get_next()` call can block for a considerable amount of time...this async version would support a world where it's easier to cancel running queries since we can loop + wait in R land (possibly calling the brand-new "cancel" function coming to ADBC 1.1).

This PR currently will leak memory and doesn't communicate any error information beyond "result code".

See also https://github.com/r-dbi/adbc/discussions/6

Reprex:

``` r
library(nanoarrow)

stream <- basic_array_stream(list(1:5))

# ...more usefully, a promise or future whose value is set from the callback
result <- NULL
nanoarrow:::nanoarrow_array_stream_get_next_async(stream, function(code, array) {
  message("done!")
  result <<- array
})

Sys.sleep(0.5)
nanoarrow:::run_callbacks()
#> done!
convert_array(result)
#> [1] 1 2 3 4 5
```

<sup>Created on 2023-06-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/r/R/array-stream.R
+++ b/r/R/array-stream.R
@@ -259,3 +259,13 @@ names.nanoarrow_array_stream <- function(x, ...) {
 `$.nanoarrow_array_stream` <- function(x, i, ...) {
   x[[i]]
 }
+
+nanoarrow_array_stream_get_next_async <- function(stream, callback,
+                                                  schema = stream$get_schema()) {
+  callback_env <- new.env(parent = emptyenv())
+  callback_env$callback <- callback
+  array <- nanoarrow_allocate_array()
+  nanoarrow_array_set_schema(array, schema, validate = FALSE)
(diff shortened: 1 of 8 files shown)


krlmlr:
🤩 🤩 🤩 

Give me some time to dig deeper into it. I might add a promises-based wrapper around that.

krlmlr:
I was wondering if the thread should be a property of the connection object. That way we don't need to create and destroy threads, only wake them for each operation (limitless producer-consumer design?).

krlmlr:
Do you plan to further enhance this?

PR #212, opened 2023-06-07
fix(r): Fix `convert_array_stream()` for non-record batch stream with zero batches

krlmlr:
Looks like a copy-pastie.


paleolimbot (approved):
Thank you for catching this!

PR #214, opened 2023-06-08
feat(extensions/nanoarrow_ipc): Add endian swapping to IPC reader

paleolimbot:
Ensures IPC reader compatibility reading little endian streams from big endian (99% for the purposes of not scaring away potential dependencies that run tests on big endian). It's also not that hard to do since flatcc did the hard work of defining `bswap16()`, `bswap32()`, and `bswap64()` based on compiler defines.

### DIFF
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
@@ -1529,10 +1529,6 @@ static int ArrowIpcDecoderWalkSetArrayView(struct ArrowIpcArraySetter* setter,
     setter->src.data_type = array_view->layout.buffer_data_type[i];
     setter->src.element_size_bits = array_view->layout.element_size_bits[i];
 
-    if (setter->src.data_type == NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO) {
-      printf("fish!\n");
-    }
-
     NANOARROW_RETURN_NOT_OK(
         ArrowIpcDecoderMakeBuffer(setter, buffer_offset, buffer_length,
                                   &array_view->buffer_views[i], buffer_dst, error));
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
@@ -1371,7 +1371,7 @@ static int ArrowIpcDecoderSwapEndian(struct ArrowIpcBufferSource* src,
         item.months = bswap32(item.months);
         item.days = bswap32(item.days);
         item.ns = bswap64(item.ns);
-        memcpy(dst + i * item_size_bytes, &item, item_size_bytes);
+        memcpy(ptr_dst + i * item_size_bytes, &item, item_size_bytes);
       }
       break;
     }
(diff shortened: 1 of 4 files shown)


extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
```diff
@@ -1297,6 +1299,143 @@ static struct ArrowIpcBufferFactory ArrowIpcBufferFactoryFromShared(
...
+      struct ArrowIpcDecimal128* ptr_src =
+          (struct ArrowIpcDecimal128*)out_view->data.data;
+      struct ArrowIpcDecimal128* ptr = (struct ArrowIpcDecimal128*)dst->data;
+      uint64_t words[2];
+      for (int64_t i = 0; i < (dst->size_bytes / 16); i++) {
+        words[0] = bswap64(ptr_src[i].words[0]);
+        words[1] = bswap64(ptr_src[i].words[1]);
+        ptr[i].words[0] = words[1];
+        ptr[i].words[1] = words[0];
+      }
```
paleolimbot, line 1364: I'm unsure of my casting-to-struct technique here for Decimal128, Decimal256, and MonthDayNano...they pass all the tests and are consistent with the C spec for alignment/padding of structs but I don't know if this is considered undefined behaviour.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
```diff
@@ -1297,6 +1299,143 @@ static struct ArrowIpcBufferFactory ArrowIpcBufferFactoryFromShared(
...
+
+static int ArrowIpcDecoderSwapEndian(struct ArrowIpcBufferSource* src,
+                                     struct ArrowBufferView* out_view,
+                                     struct ArrowBuffer* dst, struct ArrowError* error) {
+  // Some buffer data types don't need any endian swapping
+  switch (src->data_type) {
+    case NANOARROW_TYPE_BOOL:
+    case NANOARROW_TYPE_INT8:
+    case NANOARROW_TYPE_UINT8:
+    case NANOARROW_TYPE_STRING:
```
lidavidm, line 1332: Hmm, we don't need to swap the offsets?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
lidavidm: I think the safe way to do this is to memcpy from the byte array to the struct and back. That said we already do casting to and from other types...

lidavidm (approved):

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
lidavidm: Ah, sounds good.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
paleolimbot: I updated this to not add even more types that we reinterpret cast to...there are a number of places that (perhaps inadvisably) cast to `int64_t*` or `uint64_t*`, but adding more seems like a bad idea for marginally better readability.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
```diff
@@ -1554,6 +1529,10 @@ static int ArrowIpcDecoderWalkSetArrayView(struct ArrowIpcArraySetter* setter,
     setter->src.data_type = array_view->layout.buffer_data_type[i];
     setter->src.element_size_bits = array_view->layout.element_size_bits[i];
 
+    if (setter->src.data_type == NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO) {
+      printf("fish!\n");
```
lidavidm, line 1533: Did you mean to leave this here?

lidavidm (approved):

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
paleolimbot: Definitely not! 😬

PR #221, opened 2023-06-08
feat: Include dictionary member in `ArrowArrayView` struct

paleolimbot:
This doesn't add any features really, but ensures that one can walk `ArrowSchema`, `ArrowArray`, and `ArrowArrayView` recursively using the same pattern. I'd like to include this in the 0.2 release because it is very difficult to work around this limitation: for example, if you have a deeply nested dictionary field, you currently have to walk your whole tree of arrays twice (once to validate the non-dictionary bits, once looking for dictionary bits that may or may not exist to validate). The R package worked around this in some creative but error-prone ways and I'd like to avoid anybody else attempting creative workarounds when this is a feature that will almost certainly get added soon.

Another application of this is device array support, since that PR essentially constructs an `ArrowArrayView` and recursively copies the buffer view from the array view to a standalone buffer using some device-specific logic.

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -168,21 +168,40 @@ ArrowErrorCode ArrowArrayInitFromType(struct ArrowArray* array,
 ArrowErrorCode ArrowArrayInitFromArrayView(struct ArrowArray* array,
                                            struct ArrowArrayView* array_view,
                                            struct ArrowError* error) {
-  ArrowArrayInitFromType(array, array_view->storage_type);
+  NANOARROW_RETURN_NOT_OK_WITH_ERROR(
+      ArrowArrayInitFromType(array, array_view->storage_type), error);
+  int result;
+
   struct ArrowArrayPrivateData* private_data =
       (struct ArrowArrayPrivateData*)array->private_data;
-
-  int result = ArrowArrayAllocateChildren(array, array_view->n_children);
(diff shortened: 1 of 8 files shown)


src/nanoarrow/array.c
```diff
@@ -168,21 +168,40 @@ ArrowErrorCode ArrowArrayInitFromType(struct ArrowArray* array,
 ArrowErrorCode ArrowArrayInitFromArrayView(struct ArrowArray* array,
                                            struct ArrowArrayView* array_view,
                                            struct ArrowError* error) {
-  ArrowArrayInitFromType(array, array_view->storage_type);
+  NANOARROW_RETURN_NOT_OK_WITH_ERROR(
+      ArrowArrayInitFromType(array, array_view->storage_type), error);
```
lidavidm, line 172: Seems like you could guard against forgetting with `-Wunused-result` and manually annotating functions with the (compiler specific) `__attribute__((warn_unused_result))` though that is rather a hassle

lidavidm (approved):
C code looks good to me, didn't take too close a look at R.

src/nanoarrow/array.c
paleolimbot: Good idea! I will target that for 0.3 (mostly to avoid wreaking havoc on any existing usage): https://github.com/apache/arrow-nanoarrow/issues/224

PR #234, opened 2023-06-15
build(python): use src/ layout for python package

jorisvandenbossche:
Follow-up on https://github.com/apache/arrow-nanoarrow/pull/117, moving back to a src layout (https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/)


jorisvandenbossche:
@paleolimbot I didn't do anything special (apart from adding src in `python -m pytest --cov ./src/nanoarrow`), but coverage seems to be working out of the box?

```
Name                        Stmts   Miss  Cover
-----------------------------------------------
src/nanoarrow/__init__.py       2      0   100%
src/nanoarrow/_lib.pyx        314     39    88%
src/nanoarrow/lib.py           25      3    88%
-----------------------------------------------
TOTAL                         341     42    88%
```

PR #255, opened 2023-07-11
Various fixes/edits in getting-started.md

amoeba:
Hey @paleolimbot, I went through every bit of the Getting Started guide and found a few things. Overall I found the guide to be very well-written and super helpful so all I have here are small changes. See each commit message body for details on each fix and feel free to take or leave any.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -364,7 +364,7 @@ static int linesplitter_read_internal(const std::string& src, ArrowArray* out,
     src_view.size_bytes -= next_newline + 1;
   }
 
-  NANOARROW_RETURN_NOT_OK(ArrowArrayFinishBuildingDefault(tmp.get(), error));
+  NANOARROW_RETURN_NOT_OK(ArrowArrayFinishBuilding(tmp.get(), error));
 
   ArrowArrayMove(tmp.get(), out);
   return NANOARROW_OK;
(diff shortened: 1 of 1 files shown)


docs/source/getting-started.md
```diff
@@ -364,7 +364,7 @@ static int linesplitter_read_internal(const std::string& src, ArrowArray* out,
     src_view.size_bytes -= next_newline + 1;
   }
 
-  NANOARROW_RETURN_NOT_OK(ArrowArrayFinishBuildingDefault(tmp.get(), error));
+  NANOARROW_RETURN_NOT_OK(ArrowArrayFinishBuilding(tmp.get(), error));
```
paleolimbot, line 367: This one should definitely be `ArrowArrayFinishBuildingDefault()` (although this changed between 0.1 and 0.2, so perhaps there's a reference to 0.1 above that needs fixing?)

amoeba: Yep, that was it. See https://github.com/apache/arrow-nanoarrow/pull/255/commits/8ac330371afdac5a327ee93055c746e46dd15341.

amoeba:
The issue with `ArrowArrayFinishBuildingDefault` not being found was indeed because the example FetchContent declaration referenced 0.1.0. Fixed and reverted back to `ArrowArrayFinishBuildingDefault` in https://github.com/apache/arrow-nanoarrow/pull/255/commits/8ac330371afdac5a327ee93055c746e46dd15341. So this should be good to go.

PR #258, opened 2023-07-17
Add Support for Intervals

WillAyd:


src/nanoarrow/nanoarrow_types.h
```diff
@@ -656,6 +672,15 @@ struct ArrowArrayPrivateData {
   int8_t union_type_id_is_child_index;
 };
 
+/// \brief A representation of an interval.
+struct ArrowInterval {
+  /// \brief The type of interval being used
+  enum ArrowIntervalUnit unit;
+
+  /// \brief Pointer containing the underlying data for the interval
+  void* data;
```
lidavidm, line 681: Possibly a `union` is preferable here to avoid an extra pointer indirection?

src/nanoarrow/array_test.cc
```diff
@@ -29,6 +29,7 @@
 #include <arrow/util/decimal.h>
 
 #include "nanoarrow/nanoarrow.h"
+#include "nanoarrow/nanoarrow_types.h"
```
paleolimbot, line 32: Is this include needed?

src/nanoarrow/nanoarrow.h
```diff
@@ -870,6 +870,14 @@ static inline ArrowErrorCode ArrowArrayAppendBytes(struct ArrowArray* array,
 static inline ArrowErrorCode ArrowArrayAppendString(struct ArrowArray* array,
                                                     struct ArrowStringView value);
 
+/// \brief Append a Interval to an array
+///
+/// Returns NANOARROW_OK if value can be exactly represented by
+/// the underlying storage type or EINVAL otherwise (e.g., value
+/// is outside the valid array range).
```
paleolimbot, line 877: This probably doesn't apply to the interval?

src/nanoarrow/nanoarrow_types.h
```diff
@@ -656,6 +672,19 @@ struct ArrowArrayPrivateData {
...
+  /// \brief The type of interval being used
+  enum ArrowIntervalUnit unit;
+
+  /// \brief Pointer containing the underlying data for the interval
+  union {
+    int32_t year_month;
+    struct ArrowIntervalDayTime day_time;
+    struct ArrowIntervalMonthDayNano month_day_nano;
+  } data;
+};
```
paleolimbot, line 686: For `ArrowDecimal` we do something more along the lines of "declare more fields than needed since the caller probably stack-allocated one of these for whatever loop they're in". The definition here is OK too, but for the sake of argument:

```
struct ArrowInterval {
  enum ArrowType type;
  int32_t month;
  int32_t day;
  int64_t time;
  int64_t nano;
};
```

...might be simpler?

paleolimbot (commented):
This is awesome! I have one suggestion about simplifying the `ArrowInterval` but am excited to have support for this!

src/nanoarrow/nanoarrow_types.h
WillAyd: `time` is supposed to represent milliseconds in that struct right? Not a strong preference but I think using the structs is a little easier as an end user. With the current design I was thinking an anonymous union would be the best, though that would require C11 instead of just C99

WillAyd:
Any idea what I might be overlooking with the tests and how they allocate memory? Locally I get this valgrind traceback:

```sh
==710574==    at 0x4843828: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==710574==    by 0x1C6BFD: ArrowMalloc (utils.c:194)
==710574==    by 0x1BF55A: ArrowArrayInitFromType (array.c:139)
==710574==    by 0x143D36: ArrayTest_ArrayTestAppendToIntervalArrayMonthDayNano_Test::TestBody() (array_test.cc:934)
==710574==    by 0x2048CA: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2607)
==710574==    by 0x1FC940: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2643)
==710574==    by 0x1D002F: testing::Test::Run() (gtest.cc:2682)
==710574==    by 0x1D0AFD: testing::TestInfo::Run() (gtest.cc:2861)
==710574==    by 0x1D1426: testing::TestSuite::Run() (gtest.cc:3015)
==710574==    by 0x1E13F4: testing::internal::UnitTestImpl::RunAllTests() (gtest.cc:5855)
==710574==    by 0x2058C9: bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (gtest.cc:2607)
==710574==    by 0x1FDAD2: bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (gtest.cc:2643)
```

Not immediately clear what is different from the other tests in terms of building the array

WillAyd:
Ignore comment above - I see that `ImportArray` takes ownership of the memory so probably have to implement that or explicitly call array.release

src/nanoarrow/nanoarrow_types.h
paleolimbot: For the time being can we parallel the `ArrowDecimal` pattern (i.e., one flat `struct` + an `ArrowIntervalInit()` helper) for consistency? There are almost certainly better names than `month`/`day`/`type`/`nano` that perhaps give some hint as to how the spec defines them.

WillAyd: Yea makes sense. Actually looking in CI I don't think Cython supports using a union within a struct, so this is probably the only way

WillAyd: Even with flattening to one struct I think we will still need to keep structs like `ArrowIntervalDayTime` and `ArrowIntervalMonthDayNano` to unpack the data buffer - is that in line with your expectation or am I misinterpreting?

paleolimbot: Ah, now I see `reinterpret_cast<const struct ArrowIntervalMonthDayNano*>` and friends. In #214 we opted to `memcpy()` instead of `reinterpret_cast<>` ( https://github.com/apache/arrow-nanoarrow/pull/214#discussion_r1223410268 ), although I'm aware we `reinterpret_cast<>` and make it easy for others to do so with the int/float/double primitives. I think the idea was to not continue that for anything more exotic than an `int64`.

paleolimbot: I suppose we still need something to `memcpy()` into though...I'm much more interested in having interval support than having the struct be exactly like `ArrowDecimal` (although `ArrowIntervalUnit` can probably just be `ArrowType` unless there's another distinction I'm missing?).

WillAyd: Nice catch - simplified that

WillAyd:
I don't think the Python error is related?

src/nanoarrow/nanoarrow_types.h
```diff
@@ -656,6 +656,28 @@ struct ArrowArrayPrivateData {
...
+/// \ingroup nanoarrow-utils
+static inline void ArrowIntervalInit(struct ArrowInterval* interval, enum ArrowType unit,
+                                     int32_t months, int32_t days, int32_t ms,
+                                     int64_t ns) {
+  interval->unit = unit;
+  interval->months = months;
+  interval->days = days;
+  interval->ms = ms;
+  interval->ns = ns;
+}
```
paleolimbot, line 679: I think this can just zero-initialize and set `unit` and encourage direct field access like `ArrowDecimal`? Most users won't use all the fields at once.

src/nanoarrow/nanoarrow_types.h
```diff
@@ -656,6 +656,28 @@ struct ArrowArrayPrivateData {
...
 
+/// \brief A representation of an interval.
+struct ArrowInterval {
+  /// \brief The type of interval being used
+  enum ArrowType unit;
+  int32_t months;
+  int32_t days;
+  int32_t ms;
+  int64_t ns;
+};
```
paleolimbot, line 667: Awesome!

I think this will need the fields documented for them to show up in the generated documentation (and `\ingroup nanoarrow-utils` for it to show up in the right place).

We have a lot of `ArrowType` variables that are named `type`...would that be appropriate here?

dist/nanoarrow.h
paleolimbot, line 1: Can you revert the change to this file? I should add the hook to modify it to pre-commit but right now it's updated via a nightly job (for better or worse).

paleolimbot (commented):
Yes, the Python failure is happening on main too (maybe Cython 3 releated).

Just a few nits!

paleolimbot (approved):
Thank you!

(I should have merged this before the pre-commit one...I'll take care of that 😬 )

PR #262, opened 2023-07-19
fix(python): Ensure generator does not raise `StopIteration`

paleolimbot:
As noted in #258, the Python tests are failing! I'm not sure the details of why this started now, but `StopIteration` was being raised in a generator which should just return instead.

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -894,8 +894,11 @@ cdef class ArrayStream:
             return array
 
     def __iter__(self):
-        while True:
-            yield self.get_next()
+        try:
+            while True:
+                yield self.get_next()
+        except StopIteration:
+            return
 
     @staticmethod
     def allocate():


python/src/nanoarrow/_lib.pyx
```diff
@@ -894,8 +894,11 @@ cdef class ArrayStream:
             return array
 
     def __iter__(self):
-        while True:
-            yield self.get_next()
+        try:
```
WillAyd, line 897: I _think_ this is also the use case for `yield from` in Python

paleolimbot: Thanks! I just looked it up and my head is about to explode but it does sound cleaner/the thing we should be doing 🤯

jorisvandenbossche: I was also refreshing my Python knowledge on this, and so it's actually in the `__next__` method that a StopIteration needs to be raised, not in `__iter__`

So another way to implement this would be:

```
def __iter__(self):
    return self

def __next__(self):
    return self.get_next()
```

(but the current one works fine as well)

PR #269, opened 2023-07-24
fix(r): Don't link to arrow package R6 class pages

paleolimbot:
These pages were renamed as part of the 13.0.0 release and will cause submission problems for arrow's reverse dependency checks (since the page removal causes the CMD check to fail).


thisisnic (approved):
Thanks!

PR #270, opened 2023-07-28
Add ArrowArrayViewGetIntervalUnsafe

WillAyd:
Follow on to https://github.com/apache/arrow-nanoarrow/pull/258

### DIFF
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -951,7 +951,6 @@ static inline void ArrowArrayViewGetIntervalUnsafe(struct ArrowArrayView* array_
       break;
     }
     default:
-      out = NULL;
       break;
   }
 }
--- a/src/nanoarrow/array_inline.h
+++ b/src/nanoarrow/array_inline.h
@@ -928,6 +928,34 @@ static inline struct ArrowBufferView ArrowArrayViewGetBytesUnsafe(
   return view;
 }
 
+static inline void ArrowArrayViewGetIntervalUnsafe(struct ArrowArrayView* array_view,
+                                                   int64_t i, struct ArrowInterval* out) {
+  const uint8_t* data_view = array_view->buffer_views[1].data.as_uint8;
+  switch (array_view->storage_type) {
+    case NANOARROW_TYPE_INTERVAL_MONTHS: {
+      const size_t size = 4;
+      memcpy(&out->months, data_view + i * size, 4);
+      break;
+    }
+    case NANOARROW_TYPE_INTERVAL_DAY_TIME: {
+      const size_t size = 8;
+      memcpy(&out->days, data_view + i * size, 4);
+      memcpy(&out->ms, data_view + i * size + 4, 4);
+      break;
+    }
+    case NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO: {
+      const size_t size = 16;
+      memcpy(&out->months, data_view + i * size, 4);
+      memcpy(&out->days, data_view + i * size + 4, 4);
+      memcpy(&out->ns, data_view + i * size + 8, 8);
+      break;
+    }
+    default:
+      out = NULL;
+      break;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


src/nanoarrow/array_inline.h
```diff
@@ -928,6 +928,34 @@ static inline struct ArrowBufferView ArrowArrayViewGetBytesUnsafe(
   return view;
 }
 
+static inline void ArrowArrayViewGetIntervalUnsafe(struct ArrowArrayView* array_view,
+                                                   int64_t i, struct ArrowInterval* out) {
+  const uint8_t* data_view = array_view->buffer_views[1].data.as_uint8;
+  switch (array_view->storage_type) {
+    case NANOARROW_TYPE_INTERVAL_MONTHS: {
+      const size_t size = 4;
+      memcpy(&out->months, data_view + i * size, 4);
```
paleolimbot, line 937: Could the `4`s and `8`s here be `sizeof(int32_t)` and `sizeof(int64_t)` for readability?

```suggestion
      memcpy(&out->months, data_view + i * size, sizeof(int32_t));
```

src/nanoarrow/array_inline.h
```diff
@@ -928,6 +928,34 @@ static inline struct ArrowBufferView ArrowArrayViewGetBytesUnsafe(
...
+    }
+    case NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO: {
+      const size_t size = 16;
+      memcpy(&out->months, data_view + i * size, 4);
+      memcpy(&out->days, data_view + i * size + 4, 4);
+      memcpy(&out->ns, data_view + i * size + 8, 8);
+      break;
+    }
+    default:
+      out = NULL;
```
paleolimbot, line 954: maybe `memset(out, 0, sizeof(struct ArrowInterval)`?

src/nanoarrow/array_test.cc
```diff
@@ -2453,6 +2453,84 @@ TEST(ArrayViewTest, ArrayViewTestGetString) {
   TestGetFromBinary<FixedSizeBinaryBuilder>(fixed_size_builder);
 }
 
+TEST(ArrayViewTest, ArrayViewTestGetIntervalYearMonth) {
+  GTEST_SKIP() << "MonthIntervalBuilder not implemented in Arrow";
```
paleolimbot, line 2457: Would it be possible to copy some code from the previous test you added to build up one of these arrays?

paleolimbot (commented):
Thank you! (And thank you for fixing the CI error). Just a few notes!

src/nanoarrow/array_inline.h
paleolimbot: Or setting all components to 0? Or nothing? It's undefined behaviour but I'm not sure what `out = NULL` would actually do.

WillAyd: I think the downside to setting to 0 is that it is ambiguous if you've had an error or an interval is legitimately 0

paleolimbot: Agreed... perhaps nothing then so that nobody relies on that behavior?

WillAyd: Sure removed that for now. Another thought I had was returning an `ArrowInterval`; that would break the usage pattern with `GetDecimalUnsafe` but arguably be more consistent with `GetStringUnsafe`, `GetBytesUnsafe`, etc... 

Those methods do assign some members to NULL so the caller has a way of checking if they've made a mistake

paleolimbot: I think the motivation there was to make it more likely that some test would fail...perhaps some future documentation could make it more clear that it's the caller's responsibility to type check first (in theory that's what the `Unsafe` bit is trying to communicate).

Returning the `ArrowInterval` is definitely an option...my hunch is that it would be inefficient since it you'd have to potentially copy a few of the fields for every item in the loop; however, I have no evidence for that. I think what you have here is a good compromise to support all three interval types. If there's overwhelming feedback suggesting that it's confusing or slow it can always be changed pre 1.0.

paleolimbot (approved):
Thank you!

PR #273, opened 2023-08-14
chore: Typo Fixup

WillAyd:
I am attempting to vendor nanoarrow into pandas (see https://github.com/pandas-dev/pandas/pull/54506) 

We use codespell in our pre-commit and it is failing due to these. https://github.com/codespell-project/codespell

Could also add to pre-commit here if there is interest


paleolimbot (approved):
Thank you! I would love to have that as part of pre-commit (feel free to PR or just open an issue and I can tackle it later on)

PR #274, opened 2023-08-14
ci: Add codespell to pre-commit

WillAyd:


.pre-commit-config.yaml
```diff
@@ -57,5 +57,16 @@ repos:
...
+    rev: v2.2.5
+    hooks:
+    -   id: codespell
+        types_or: [rst, markdown, c, c++]
+        additional_dependencies: [tomli]
+        exclude: |
+            (?x)
+            ^extensions/nanoarrow_ipc/thirdparty
+            |nanoarrow_ipc_flatcc_generated.h
+            |nanoarrow_device_metal.cc
```
WillAyd, line 70: Unfortunate as only the `pEvent` code in one of the comments is producing a false positive. We could alternately create a `.codespellignorelines` file to the repo. but figured this is fine for now

paleolimbot: That's perfect...I'm planning to do some work in the device extension later on and will deal with it then.

paleolimbot (approved):
Thank you!

PR #276, opened 2023-08-16
feat: Implement ArrowBitmapUnpackInt8Unsafe

WillAyd:


src/nanoarrow/nanoarrow.h
```diff
@@ -664,6 +664,10 @@ static inline ArrowErrorCode ArrowBufferAppendBufferView(struct ArrowBuffer* buf
 /// \brief Extract a boolean value from a bitmap
 static inline int8_t ArrowBitGet(const uint8_t* bits, int64_t i);
 
+/// \brief Extract boolean values from a range in a bitmap
+static inline void ArrowBitsGet(const uint8_t* bits, int64_t start_offset, int64_t length,
```
WillAyd, line 668: What is perhaps a little confusing is that `start_offset` is number of bits for this function, whereas the set functions refer to it as number of bytes. I think that makes sense given the inputs/outputs, but maybe we should change the identifers?

paleolimbot: I'm not sure I understand...I think that `offset` and `length` are always "number of items" (bits for a `uint8_t* bits`; bytes for `uint8_t*`, etc) but perhaps there are functions for which this isn't the case?

src/nanoarrow/buffer_inline.h
```diff
@@ -236,6 +242,41 @@ static inline int8_t ArrowBitGet(const uint8_t* bits, int64_t i) {
   return (bits[i >> 3] >> (i & 0x07)) & 1;
 }
 
+static inline void ArrowBitsGet(const uint8_t* bits, int64_t start_offset, int64_t length,
+                                int8_t* out) {
```
paleolimbot, line 246: Maybe `ArrowBitUnpackInt8()`? (As a follow-up I'll add the `Int32()` version)

src/nanoarrow/nanoarrow.h
WillAyd: That's a much simpler way of expressing this than I was putting out there

src/nanoarrow/buffer_inline.h
```diff
@@ -236,6 +242,41 @@ static inline int8_t ArrowBitGet(const uint8_t* bits, int64_t i) {
...
+  if (length == 0) {
+    return;
+  }
+
+  const uint8_t* bits_cursor = bits;
+  int64_t n_remaining = length;
+  int8_t* out_cursor = out;
```
paleolimbot, line 253: Do you mind using the same terminology/approach as `ArrowBitCountSet()`? ( https://github.com/apache/arrow-nanoarrow/blob/main/src/nanoarrow/buffer_inline.h#L297-L302 ). I recently had to fix a segfault resulting from index math on the first/middle/last byte and while I'm not sure the approach there is better than what you have here, it will probably help fix the next issue with either function to have the code look similar for both functions.

src/nanoarrow/buffer_inline.h
```diff
@@ -222,6 +222,12 @@ static inline int64_t _ArrowBytesForBits(int64_t bits) {
   return (bits >> 3) + ((bits & 7) != 0);
 }
 
+static inline void _ArrowBitmapUnpackInt8(const uint8_t* bits, int8_t* out) {
+  for (int i = 0; i < 8; i++) {
+    out[i] = (*bits >> i) & 1;
+  }
```
paleolimbot, line 228: I don't know the details well enough to know if it matters, but Arrow C++ writes out all N lines explicitly ( https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/bpacking_default.h#L35-L105 )

src/nanoarrow/buffer_test.cc
```diff
@@ -271,6 +271,65 @@ TEST(BitmapTest, BitmapTestElement) {
...
+
+  memset(bitmap, 0xff, sizeof(bitmap));
+  ArrowBitsGet(bitmap, 0, sizeof(result), result);
+  for (int i = 0; i < sizeof(result); i++) {
+    EXPECT_EQ(result[i], 1);
+  }
+
+  bitmap[2] = 0xfd;
+  ArrowBitsGet(bitmap, 0, sizeof(result), result);
+  EXPECT_EQ(result[16 + 0], 1);
```
paleolimbot, line 286: Would an expectation in the form `EXPECT_EQ(std::basic_string<uint8_t>(result, n), std::basic_string<uint8_t>({0, 1, 1, 0, 0, 0}))` be any more readable? It might scale better to include more cases (e.g., offset != 0, all bits within the same byte)

src/nanoarrow/buffer_test.cc
```diff
@@ -271,6 +271,65 @@ TEST(BitmapTest, BitmapTestElement) {
   EXPECT_EQ(ArrowBitGet(bitmap, 16 + 7), 0);
 }
 
+TEST(BitmapTest, BitmapTestBitsGet) {
+  uint8_t bitmap[10];
+  int8_t result[sizeof(bitmap) * 8];
```
paleolimbot, line 276: You can probably test all the cases with just 3 bytes (which might make the expectations a little less verbose)? I recently fixed a segfault in the first byte/middle byte/last byte logic in `ArrowBitCountSet()` and it may be worth testing that here, too (i.e., offset == 0, offset > 0 that does not align on a byte boundary, offset + length ends on a byte boundary, offset + length does *not* end on a byte boundary, offset + length refer to bits in the same byte).

paleolimbot (commented):
This is awesome! I'm excited to follow this up with a 32-bit version for R's logical vectors. A few notes to start the discussion!

src/nanoarrow/buffer_test.cc
WillAyd: Not sure if the newer test design is exactly what you are looking for from a readability perspective, but should definitely give better coverage

src/nanoarrow/buffer_inline.h
```diff
@@ -222,6 +222,18 @@ static inline int64_t _ArrowBytesForBits(int64_t bits) {
   return (bits >> 3) + ((bits & 7) != 0);
 }
 
+static inline void _ArrowBitmapUnpackInt8(const uint8_t* bits, int8_t* out) {
+  const uint8_t word = *bits;
```
paleolimbot, line 226:
```suggestion
static inline void _ArrowBitmapUnpackInt8(uint8_t word, int8_t* out) {
```

...may simplify its usage below?

src/nanoarrow/buffer_inline.h
```diff
@@ -236,6 +248,46 @@ static inline int8_t ArrowBitGet(const uint8_t* bits, int64_t i) {
...
+  }
+
+  // first byte
+  for (int i = 0; i < 8 - (i_begin % 8); i++) {
+    *out++ = ArrowBitGet(&bits[bytes_begin], i + i_begin % 8);
+  }
+
+  // middle bytes
+  for (int64_t i = bytes_begin + 1; i < bytes_last_valid; i++) {
+    _ArrowBitmapUnpackInt8(&bits[i], out);
```
paleolimbot, line 280:
```suggestion
    _ArrowBitmapUnpackInt8(bits[i], out);
```

src/nanoarrow/buffer_test.cc
paleolimbot: It's perfect! Thanks! Does it make sense to remove this first bit that doesn't use the helper? (I am a little confused as to what it's testing that the tests below it don't cover).

paleolimbot (approved):
Thank you!

PR #280, opened 2023-08-17
perf: Improved Bit (Un)packing Performance

WillAyd:
I was very surprised by this but getting rid of the shifting yielded a huge performance boost for me locally. I was benchmarking some pandas code that took ~500us to unpack 1 million boolean values - with this simple change that time fell to ~30us

Not an expert in assembly but here is what godbolt produces to set the index of 1 before:

```asm
        movzx   eax, BYTE PTR [rbp-1]
        shr     al
        mov     edx, eax
        .loc 1 6 6
        mov     rax, QWORD PTR [rbp-32]
        add     rax, 1
        .loc 1 6 24
        and     edx, 1
        .loc 1 6 10
        mov     BYTE PTR [rax], dl
```

and after:

```asm
        movzx   eax, BYTE PTR [rbp-1]
        and     eax, 2
        .loc 1 6 25
        test    eax, eax
        setne   dl
        .loc 1 6 6
        mov     rax, QWORD PTR [rbp-32]
        add     rax, 1
        .loc 1 6 10
        mov     BYTE PTR [rax], dl
```

Assuming the `shr` instruction is inefficient compared to the `test` / `setne` approach taken in the latter


WillAyd:
Haven't profiled the packing methods yet but I assume we stand to gain a bit removing the left shift operators there as well

WillAyd:
Hmm very strange. I also pieced together this benchmark in C:

```c
#include <stdlib.h>

#include "nanoarrow.h"
#include "buffer_inline.h"

int main() {
  int N = 1000000;
  int8_t* values = malloc(N);
  memset(values, 0, N);
  
  struct ArrowBitmap bitmap;
  ArrowBitmapInit(&bitmap);
  ArrowBitmapReserve(&bitmap, N);
  ArrowBitmapAppendInt8Unsafe(&bitmap, values, N);

  for (int i = 0; i < 10000; i++) {
    ArrowBitsUnpackInt8(bitmap.buffer.data, 0, N, values);
  }
  free(values);
}
```

which I think generally showed improvement, but not to the degree and consistency with what I was seeing in the Python/Cython library I originally noticed this in. Will see if I can test some more

WillAyd:
For reference here is a "shiftless" packing implementation too, though I haven't seen as much speed up using this:

```c
static inline void _ArrowBitmapPackInt8(const int8_t* values, uint8_t* out) {
  *out = (values[0]
          | ((values[1] + 0x1) & 0x2)
          | ((values[2] + 0x3) & 0x4)
          | ((values[3] + 0x7) & 0x8)
          | ((values[4] + 0xf) & 0x10)
          | ((values[5] + 0x1f) & 0x20)
          | ((values[6] + 0x3f) & 0x40)
          | ((values[7] + 0x7f) & 0x80)
          );
}
```

I am still testing these downstream so no rush to merge this - let me see if anything becomes clearer as I work through this in pandas and will report back

WillAyd:
I set up a repo to compare this from Cython:

https://github.com/WillAyd/cython-nanoarrow

The benchmarks for me look something like this:

```python
In [1]: from comparisons import ComparisonManager
   ...: mgr = ComparisonManager()
   ...: %timeit mgr.unpack()
   ...: %timeit mgr.unpack_no_shift()
547 µs ± 21 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
40.5 µs ± 321 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Not sure what makes Python here different than the R / C benchmarks we've tried before...

WillAyd:
On the writing side of things my benchmarks shows ~10% improvement not using shift. Not as drastic as read, but still maybe worthwhile

```python
In [2]: from comparisons import ComparisonManager
   ...: mgr = ComparisonManager()
   ...: %timeit mgr.pack()
   ...: %timeit mgr.pack_no_shift()
80.1 µs ± 984 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
72.8 µs ± 390 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [3]: from comparisons import ComparisonManager
   ...: mgr = ComparisonManager()
   ...: %timeit mgr.pack()
   ...: %timeit mgr.pack_no_shift()
78.9 µs ± 671 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
71.5 µs ± 195 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [4]: from comparisons import ComparisonManager
   ...: mgr = ComparisonManager()
   ...: %timeit mgr.pack()
   ...: %timeit mgr.pack_no_shift()
80.4 µs ± 324 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
72.2 µs ± 272 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

jorisvandenbossche:
FWIW I quickly cloned your repo and followed the instructions there, and I see similar results as you (big difference between `unpack` and `unpack_no_shift`, 1.36ms vs 140µs, so also around 10x difference)

WillAyd:
Hmm strange. So for you the packing without shifts is slower? I see 142 us going up to 203?

WillAyd:
Are you on an intel or m1 chip? Guessing that could explain some of the differences. Interesting to see all this variation...

jorisvandenbossche:
(I am on ubuntu / intel cpu)

WillAyd:
That is an interesting find! Though I wonder since the macro makes a difference - if you compiled with `-Winline` did you see any errors about the arrow functions not getting inline'd? You'd have to sift through some Cython warnings/errors in the project I provided, but that may be part of the reason why a macro helps

paleolimbot (approved):
Thank you!

PR #288, opened 2023-08-24
feat(r): Implement extension type registration/conversion 

paleolimbot:


r/R/convert-array.R
```diff
@@ -85,14 +85,28 @@ convert_array <- function(array, to = NULL, ...) {
 #' @export
 convert_array.default <- function(array, to = NULL, ..., .from_c = FALSE) {
   if (.from_c) {
+    # Handle extension conversion
+    # We don't need the user-friendly versions and this is performance-sensitive
+    schema <- .Call(nanoarrow_c_infer_schema_array, array)
+    parsed <- .Call(nanoarrow_c_schema_parse, schema)
```
krlmlr, line 91: Since this seems to throw away `schema`, perhaps offer a C entry point that combines both functions?

r/R/convert-array.R
```diff
@@ -85,14 +85,28 @@ convert_array <- function(array, to = NULL, ...) {
...
       values <- .Call(nanoarrow_c_convert_array, dictionary, to)
       array$dictionary <- NULL
       indices <- .Call(nanoarrow_c_convert_array, array, integer())
-      return(values[indices + 1L])
+
+      if (is.data.frame(values)) {
+        return(values[indices + 1L, , drop = FALSE])
+      } else {
+        return(values[indices + 1L])
+      }
```
krlmlr, line 109: This is `vec_slice()` :

```suggestion
      return(vec_slice(values, indices + 1L))
```

r/R/convert-array.R
```diff
@@ -85,14 +85,28 @@ convert_array <- function(array, to = NULL, ...) {
...
+      return(convert_array_extension(spec, array, to, ...))
+    }
+
     # Handle default dictionary conversion since it's the same for all types
     dictionary <- array$dictionary
 
     if (!is.null(dictionary)) {
       values <- .Call(nanoarrow_c_convert_array, dictionary, to)
       array$dictionary <- NULL
       indices <- .Call(nanoarrow_c_convert_array, array, integer())
```
krlmlr, line 103: Perhaps enable this function to add an offset directly, to avoid the `+ 1L` below?

r/R/extension-vctrs.R
```diff
@@ -0,0 +1,92 @@
...
+#'
+#' @param ptype A vctrs prototype as returned by [vctrs::vec_ptype()].
+#' @inheritParams na_type
+#'
+#' @return A [nanoarrow_schema][as_nanoarrow_schema].
+#' @export
+#'
+#' @examples
+#' vctr <- as.POSIXlt("2000-01-02 03:45", tz = "UTC")
+#' array <- as_nanoarrow_array(vctr, schema = na_vctrs(vctrs::vec_ptype(vctr)))
```
krlmlr, line 34: Would `na_vctrs` also accept a non-length-0 vector?

```suggestion
#' array <- as_nanoarrow_array(vctr, schema = na_vctrs(vctr))
```

r/R/extension-vctrs.R
```diff
@@ -0,0 +1,92 @@
...
+
+#' Vctrs extension type
+#'
+#' The Arrow format provides a rich type system that can handle most R
+#' vector types; however, many R vector types do not roundtrip perfectly
+#' through Arrow memory. The vctrs extension type uses [vctrs::vec_data()],
+#' [vctrs::vec_restore()], and [vctrs::vec_ptype()] in calls to
+#' [as_nanoarrow_array()] and [convert_array()] to ensure roundtrip fidelity.
+#'
+#' @param ptype A vctrs prototype as returned by [vctrs::vec_ptype()].
```
krlmlr, line 26: Is this correct?

```suggestion
#' @param ptype A vctrs prototype as returned by [vctrs::vec_ptype()].
#'   The prototype can be of arbitrary size, but a zero-size vector
#'   is sufficient here.
```

r/R/extension.R
```diff
@@ -0,0 +1,187 @@
...
+as_nanoarrow_array_extension <- function(extension_spec, x, ..., schema = NULL) {
+  UseMethod("as_nanoarrow_array_extension")
+}
+
+#' @export
+infer_nanoarrow_ptype_extension.default <- function(extension_spec, x, ...,
+                                                    warn_unregistered = TRUE) {
+  if (warn_unregistered) {
+    warn_unregistered_extension_type(x)
+  }
```
krlmlr, line 107: Should this be part of the generic?

r/src/convert_array.c
```diff
@@ -100,7 +100,26 @@ static SEXP convert_array_default(SEXP array_xptr, enum VectorType vector_type,
...
+  if (schema_view.extension_name.size_bytes > 0) {
+    if (ptype_sexp == R_NilValue) {
+      ptype_sexp = PROTECT(nanoarrow_c_infer_ptype(array_xptr_get_schema(array_xptr)));
+      SEXP default_result =
+          PROTECT(convert_array_default(array_xptr, VECTOR_TYPE_OTHER, ptype_sexp));
+      UNPROTECT(2);
+      return default_result;
+    } else {
+      return convert_array_default(array_xptr, VECTOR_TYPE_OTHER, ptype_sexp);
+    }
```
krlmlr, line 120: This gave me pause. Perhaps:

```suggestion
    if (ptype_sexp == R_NilValue) {
      ptype_sexp = PROTECT(nanoarrow_c_infer_ptype(array_xptr_get_schema(array_xptr)));
    } else {
      ptype_sexp = PROTECT(ptype_sexp);
    }
    SEXP default_result =
        convert_array_default(array_xptr, VECTOR_TYPE_OTHER, ptype_sexp);
    UNPROTECT(1);
    return default_result;
```

r/src/convert_array.c
```diff
@@ -117,19 +136,26 @@ static SEXP convert_array_chr(SEXP array_xptr) {
...
-        convert_array_default(array_xptr, VECTOR_TYPE_DATA_FRAME, ptype_sexp);
-    UNPROTECT(1);
-    return default_result;
+  // If array_xptr is an extension, union, or the ptype isn't a data.frame
+  // use convert/materialize convert behaviour.
+  if (schema_view.storage_type != NANOARROW_TYPE_STRUCT ||
+      schema_view.extension_name.size_bytes > 0 ||
+      (ptype_sexp != R_NilValue && !Rf_inherits(ptype_sexp, "data.frame"))) {
+    if (ptype_sexp == R_NilValue) {
+      ptype_sexp = PROTECT(nanoarrow_c_infer_ptype(array_xptr_get_schema(array_xptr)));
```
krlmlr, line 151: Same pattern as above.

r/src/materialize.c
```diff
@@ -190,6 +190,67 @@ static void fill_vec_with_nulls(SEXP x, R_xlen_t offset, R_xlen_t len) {
...
+    Rf_error("Unexpected SEXP type in result copy_vec_into()");
+  }
+
+  if (Rf_length(x) != len) {
+    Rf_error("Unexpected length of result in copy_vec_into()");
+  }
+
+  switch (TYPEOF(dst)) {
+    case RAWSXP:
+      memcpy(RAW(dst) + offset, RAW(x), len * sizeof(double));
```
krlmlr, line 229:
```suggestion
      memcpy(RAW(dst) + offset, RAW(x), len * sizeof(*RAW(x)));
```

or simply `char` .

r/src/materialize.c
```diff
@@ -190,6 +190,67 @@ static void fill_vec_with_nulls(SEXP x, R_xlen_t offset, R_xlen_t len) {
...
+        SET_STRING_ELT(dst, offset + i, STRING_ELT(x, i));
+      }
+      break;
+    case VECSXP:
+      for (R_xlen_t i = 0; i < len; i++) {
+        SET_VECTOR_ELT(dst, offset + i, VECTOR_ELT(x, i));
+      }
+      break;
+    default:
+      Rf_error("Unhandled SEXP type in copy_vec_into()");
```
krlmlr, line 249: `CPLXSXP` ?

r/tests/testthat/test-extension-vctrs.R
```diff
@@ -0,0 +1,72 @@
...
+  vectors <- list(
+    lgl = c(FALSE, TRUE, NA),
+    int = c(0L, 1L, NA_integer_),
+    dbl = c(0, 1, NA_real_),
+    chr = c("a", NA_character_),
+    posixct = as.POSIXct("2000-01-01 12:23", tz = "UTC"),
+    posixlt = as.POSIXlt("2000-01-01 12:23", tz = "UTC"),
+    date = as.Date("2000-01-01"),
+    difftime = as.difftime(123, units = "secs"),
+    data_frame_simple = tibble::tibble(x = 1:5),
```
krlmlr, line 33: Do you support packed columns (structs) here too?

r/tests/testthat/test-extension-vctrs.R
```diff
@@ -0,0 +1,72 @@
...
+    data_frame_nested = tibble::tibble(x = 1:5, y = tibble::tibble(z = letters[1:5]))
+  )
+
+  for (nm in names(vectors)) {
```
krlmlr, line 37: Experience from DBItest has taught me that you'll be much better off writing multiple `test_that(...)` blocks, one per element. Why not create a small script that generates this code for you?

Canonical reference: https://mtlynch.io/good-developers-bad-tests/ .

r/tests/testthat/test-extension-vctrs.R
```diff
@@ -0,0 +1,72 @@
...
+
+    # Roundtrip with an empty array stream
+    stream <- basic_array_stream(list(), schema = schema)
+    expect_identical(convert_array_stream(stream), ptype)
+
+    # Roundtrip with multiple chunks
+    stream <- basic_array_stream(list(array, array))
+    expect_identical(convert_array_stream(stream), vctrs::vec_rep(vctr, 2))
+
+    if (requireNamespace("arrow", quietly = TRUE)) {
```
krlmlr, line 58: This will also allow you to use `skip_if_not_installed("arrow")` here.

r/tests/testthat/test-extension.R
```diff
@@ -0,0 +1,116 @@
...
+})
+
+test_that("infer_nanoarrow_ptype() dispatches on registered extension spec", {
+  register_nanoarrow_extension(
+    "some_ext",
+    nanoarrow_extension_spec(subclass = "some_spec_class")
+  )
+  on.exit(unregister_nanoarrow_extension("some_ext"))
+
+  infer_nanoarrow_ptype_extension.some_spec_class <- function(spec, x, ...) {
```
krlmlr, line 33: There's `rlang::local_bindings()` .

krlmlr (commented):
Thanks, very exciting!

I only skimmed over it. I wonder if it's worth splitting out some of the refactorings and merging them individually.

r/R/convert-array.R
paleolimbot: This is a good observation that exposes a larger design flaw in the current system: at the point that `convert_array()` is called, we pretty much need the parsed version of a schema (i.e., `ArrowSchemaView`) and the parsed version of the array (i.e., `ArrowArrayView`). I wrote the Python bindings to handle this from the ground up but here there are a number of hacks like this one (and more than a few in C). I'm going to try to fix all of those in a more systemic way (but don't quite have the time to do it yet).

r/R/convert-array.R
paleolimbot: This also exposes a larger design question...currently, nanoarrow does not require vctrs for most data types (it does require it for converting lists-of-things into R vectors, and I think this is a good thing). I will extract a utility function so that it is more obvious what this does/easier to switch if we take on a vctrs dependency.

r/R/convert-array.R
paleolimbot: That is definitely a good optimization...ideally there would be an offset + scale for all numerics (scale is used for timestamp and difftime conversion). I would like to move the conversion process such to something like `get_converter(array, to)` returning something that could be constructed like `int_vector_converter(offset = 1)`. I'm currently trying to (for better or possibly worse) provide full type coverage (and associated test coverage), at the expense of speed for the less frequently used types.

r/R/extension.R
paleolimbot: I *think* so. I suppose the other option is to hard-code the warning somewhere before it gets here?

r/tests/testthat/test-extension-vctrs.R
paleolimbot: Yes (I think what you're referring to is tested on the next line?)

paleolimbot (commented):
Thank you!

PR #289, opened 2023-08-25
fix(extensions/nanoarrow_ipc): Fix crash and mixleading error messages resulting from corrupted streams

### LINKED ISSUE #287: Investigate crash when processing mangled streams
While testing nanoarrow's behavior with respect to error codes, a crasher was found. It can be reproduced with a short Python script that does the mangling and then runs the mangled stream through the dump_stream IPC helper.

Steps to reproduce:

1. Put https://gist.github.com/amoeba/b64fc94ba5224bafcb3734bd261181d5 in your working directory
2. Build nanoarrow_ipc with `NANOARROW_IPC_BUILD_APPS=ON` to get `dump_stream` and put it in your working directory
3. Run the attached script repeatedly until you get something like
    > dump_stream(53000,0x1f671a080) malloc: *** error for object 0x6000027b8050: pointer being freed was not allocated
    > dump_stream(53000,0x1f671a080) malloc: *** set a breakpoint in malloc_error_break to debug

In my early testing, it doesn't appear to be deterministic so you may have to run `test.py` repeatedly.

<details>
<summary><code>test.py</code></summary>

```python
import os
import base64
import subprocess

with open(os.path.join(os.path.dirname(__file__), "test_arrow_data")) as f:
    b64data = f.read()

decode_bytes = base64.b64decode(b64data)

for i in range(8):
    try:
        print(subprocess.check_output(["./dump_stream", "-"], input=decode_bytes[:i] + decode_bytes[i + 1 :]))
    except Exception as e:
        print(e)

```
</details>

<details>
<summary>Example output of running <code>test.py</code></summary>

```
❯ python test.py
stream.get_schema() returned 22 with error 'Expected 0xFFFFFFFF at start of message but found 0xE0FFFFFF'
Command '['./dump_stream', '-']' returned non-zero exit status 1.
stream.get_schema() returned 22 with error 'Expected 0xFFFFFFFF at start of message but found 0xE0FFFFFF'
dump_stream(53000,0x1f671a080) malloc: *** error for object 0x6000027b8050: pointer being freed was not allocated
dump_stream(53000,0x1f671a080) malloc: *** set a breakpoint in malloc_error_break to debug
Command '['./dump_stream', '-']' died with <Signals.SIGABRT: 6>.
stream.get_schema() returned 22 with error 'Expected 0xFFFFFFFF at start of message but found 0xE0FFFFFF'
Command '['./dump_stream', '-']' returned non-zero exit status 1.
stream.get_schema() returned 22 with error 'Expected 0xFFFFFFFF at start of message but found 0xE0FFFFFF'
Command '['./dump_stream', '-']' returned non-zero exit status 1.
stream.get_schema() returned 29 with error 'Expected 0 <= message body size <= 32943 bytes but found message body size of 268435500 bytes'
Command '['./dump_stream', '-']' returned non-zero exit status 1.
stream.get_schema() returned 29 with error 'Expected 0 <= message body size <= 32943 bytes but found message body size of 268435680 bytes'
Command '['./dump_stream', '-']' returned non-zero exit status 1.
stream.get_schema() returned 29 with error 'Expected 0 <= message body size <= 32943 bytes but found message body size of 268446944 bytes'
Command '['./dump_stream', '-']' returned non-zero exit status 1.
stream.get_schema() returned 29 with error 'Expected 0 <= message body size <= 32943 bytes but found message body size of 268446944 bytes'
Command '['./dump_stream', '-']' returned non-zero exit status 1.
```
</details>

paleolimbot:
Closes #287.

The reported crash occurred because the internal `ArrowIpcDecoderVerifyHeader()` incorrectly interpreted the return value of `ArrowIpcDecoderCheckHeader()`. Because the header checker returned an error code sometimes even if it succeeded, we had been ignoring the error in `ArrowIpcDecoderVerifyHeader()` which resulted in issuing commands like `memcpy(dst, src, -8)`. I believe that's undefined behaviour, leading to the intermittent nature of the crash.

In adding tests for this kind of error, I also made some improvements to error messages along the way.

```bash
# docker run --rm -it ghcr.io/apache/arrow-nanoarrow:ubuntu
# git clone https://github.com/apache/arrow-nanoarrow.git /arrow-nanoarrow
# or
# docker run --rm -it -v$(pwd):/arrow-nanoarrow ghcr.io/apache/arrow-nanoarrow:ubuntu

cd /arrow-nanoarrow/extensions/nanoarrow_ipc
mkdir build && cd build
cmake .. -DNANOARROW_IPC_BUILD_APPS=ON

curl https://gist.githubusercontent.com/amoeba/b64fc94ba5224bafcb3734bd261181d5/raw/af4c93da7ce6affba74a80e1ba94ed9573e91be8/test_arrow_data | \
    base64 -d > test_binary

with_byte_removed() {
    BYTE_PLUS_ONE=$(($2 + 2))
    cat $1 | head -c $2
    cat $1 | tail -c "+$BYTE_PLUS_ONE"
}

cmake --build .

echo "Errors:" > out.txt
for i in {1..32951}; do
  echo "$i/32951"
  echo "$i/32951" >> out.txt
  with_byte_removed test_binary $i | ./dump_stream - 2>> out.txt
done
```

### DIFF
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
@@ -46,6 +46,10 @@
 #include "nanoarrow_ipc.h"
 #include "nanoarrow_ipc_flatcc_generated.h"
 
+// A more readable expression way to refer to the fact that there are 8 bytes
+// at the beginning of every message header.
+const static int64_t kMessageHeaderPrefixSize = 8;
+
 // Internal representation of a parsed "Field" from flatbuffers. This
 // represents a field in a depth-first walk of column arrays and their
 // children.
@@ -929,9 +933,9 @@ static inline void ArrowIpcDecoderResetHeaderInfo(struct ArrowIpcDecoder* decode
   private_data->last_message = NULL;
 }
 
-// Returns NANOARROW_OK if data is large enough to read the message header,
-// ESPIPE if reading more data might help, or EINVAL if the content is not valid.
-// Advances the input ArrowBufferView by 8 bytes.
+// Returns NANOARROW_OK if data is large enough to read the first 8 bytes
+// of the message header, ESPIPE if reading more data might help, or EINVAL if the content
+// is not valid. Advances the input ArrowBufferView by 8 bytes.
 static inline int ArrowIpcDecoderCheckHeader(struct ArrowIpcDecoder* decoder,
                                              struct ArrowBufferView* data_mut,
                                              int32_t* message_size_bytes,
@@ -939,7 +943,7 @@ static inline int ArrowIpcDecoderCheckHeader(struct ArrowIpcDecoder* decoder,
   struct ArrowIpcDecoderPrivate* private_data =
       (struct ArrowIpcDecoderPrivate*)decoder->private_data;
 
-  if (data_mut->size_bytes < 8) {
+  if (data_mut->size_bytes < kMessageHeaderPrefixSize) {
     ArrowErrorSet(error, "Expected data of at least 8 bytes but only %ld bytes remain",
                   (long)data_mut->size_bytes);
     return ESPIPE;
@@ -954,7 +958,7 @@ static inline int ArrowIpcDecoderCheckHeader(struct ArrowIpcDecoder* decoder,
 
   int swap_endian = private_data->system_endianness == NANOARROW_IPC_ENDIANNESS_BIG;
   int32_t header_body_size_bytes = ArrowIpcReadInt32LE(data_mut, swap_endian);
-  *message_size_bytes = header_body_size_bytes + (2 * sizeof(int32_t));
+  *message_size_bytes = header_body_size_bytes + kMessageHeaderPrefixSize;
   if (header_body_size_bytes < 0) {
     ArrowErrorSet(
         error, "Expected message body size > 0 but found message body size of %ld bytes",
@@ -994,11 +998,12 @@ ArrowErrorCode ArrowIpcDecoderVerifyHeader(struct ArrowIpcDecoder* decoder,
 
   // Check that data contains at least the entire header (return ESPIPE to signal
   // that reading more data may help).
-  int64_t message_body_size = decoder->header_size_bytes - 8;
+  int64_t message_body_size = decoder->header_size_bytes - kMessageHeaderPrefixSize;
   if (data.size_bytes < message_body_size) {
     ArrowErrorSet(error,
                   "Expected >= %ld bytes of remaining data but found %ld bytes in buffer",
-                  (long)message_body_size + 8, (long)data.size_bytes + 8);
+                  (long)message_body_size + kMessageHeaderPrefixSize,
+                  (long)data.size_bytes + kMessageHeaderPrefixSize);
     return ESPIPE;
   }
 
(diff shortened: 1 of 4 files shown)


extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
```diff
@@ -930,7 +930,8 @@ static inline void ArrowIpcDecoderResetHeaderInfo(struct ArrowIpcDecoder* decode
 }
 
 // Returns NANOARROW_OK if data is large enough to read the message header,
```
lidavidm, line 932: I guess this docstring is slightly misleading then because it may return 0 even if the buffer technically does not contain the entire header; it just has to contain the continuation token and size?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
```diff
@@ -1025,6 +1029,16 @@ ArrowErrorCode ArrowIpcDecoderDecodeHeader(struct ArrowIpcDecoder* decoder,
   NANOARROW_RETURN_NOT_OK(
       ArrowIpcDecoderCheckHeader(decoder, &data, &decoder->header_size_bytes, error));
 
+  // Check that data contains at least the entire header (return ESPIPE to signal
+  // that reading more data may help).
+  int64_t message_body_size = decoder->header_size_bytes - 8;
```
lidavidm, line 1034: nit: above we use `2 * sizeof(int32_t)` but here we hardcode 8; maybe just pull it out into a constant?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_reader_test.cc
```diff
@@ -291,6 +291,62 @@ TEST(NanoarrowIpcReader, StreamReaderExpectedSchema) {
...
+    memcpy(simple_stream_invalid + i, kSimpleSchema + (i + 1),
+           (sizeof(kSimpleSchema) - i));
+
+    ArrowBufferInit(&input_buffer);
+    ASSERT_EQ(ArrowBufferAppendBufferView(&input_buffer, data), NANOARROW_OK);
+    ASSERT_EQ(ArrowIpcInputStreamInitBuffer(&input, &input_buffer), NANOARROW_OK);
+    ASSERT_EQ(ArrowIpcArrayStreamReaderInit(&stream, &input, nullptr), NANOARROW_OK);
+
+    ASSERT_NE(stream.get_schema(&stream, &schema), NANOARROW_OK)
+        << "After removing Schema message byte " << i;
```
lidavidm, line 320: FWIW, you can `SCOPED_TRACE` inside the loop and have it apply to all the assertions (though building up the message is a little annoying)

lidavidm (approved):

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_reader_test.cc
paleolimbot: Nice!

```
/Users/deweydunnington/Desktop/rscratch/arrow-nanoarrow/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_reader_test.cc:345: Failure
Expected: (stream.get_next(&stream, &array)) != (0), actual: 0 vs 0
Google Test trace:
/Users/deweydunnington/Desktop/rscratch/arrow-nanoarrow/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_reader_test.cc:332: 144
```

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_decoder.c
paleolimbot: Good catch! I also renamed the function since it's more reading than checking now.

lidavidm (approved):

PR #291, opened 2023-08-28
fix: Ensure that test for increasing offsets is not affected by overflow

### LINKED ISSUE #290: IPC input with out-of-range (but correctly increasing) last offset causes segfault when element is accessed
```bash
# docker run --rm -it ghcr.io/apache/arrow-nanoarrow:ubuntu
# git clone https://github.com/apache/arrow-nanoarrow.git /arrow-nanoarrow
# or
# docker run --rm -it -v$(pwd):/arrow-nanoarrow ghcr.io/apache/arrow-nanoarrow:ubuntu

cd /arrow-nanoarrow/extensions/nanoarrow_ipc
mkdir build && cd build
cmake .. -DNANOARROW_IPC_BUILD_APPS=ON

curl https://gist.githubusercontent.com/amoeba/b64fc94ba5224bafcb3734bd261181d5/raw/af4c93da7ce6affba74a80e1ba94ed9573e91be8/test_arrow_data | \
    base64 -d > test_binary

with_byte_removed() {
    BYTE_PLUS_ONE=$(($2 + 2))
    cat $1 | head -c $2
    cat $1 | tail -c "+$BYTE_PLUS_ONE"
    cat $1 | head -c 1
}

cmake --build .

with_byte_removed test_binary 22818 > mangled
R -e 'arrow::read_ipc_stream("mangled", as_data_frame = FALSE)$ValidateFull()'
#> Error: Invalid: Column 25: In chunk 0: Invalid: First or last binary offset out of bounds
./dump_stream mangled
#> Passes
```

It looks like nanoarrow correctly handles removing the *previous* byte:

```bash
with_byte_removed test_binary 22817 > mangled
R -e 'arrow::read_ipc_stream("mangled", as_data_frame = FALSE)$ValidateFull()'
#> Error: Invalid: Column 25: In chunk 0: Invalid: First or last binary offset out of bounds
./dump_stream mangled
#> stream.get_next() returned 22 with error '[100] Expected element size >= 0 but found element size -1013'
```

paleolimbot:
Closes #290.

This PR fixes the check for increasing offsets: previously it calculated a diff between pairs of elements and ensured that the value was > 0. This failed for expressions like `int32_t diff = (int32_t)-2147483648 - (int32_t)2080374784`, where because of overflow, `diff` evalutes to `> 0`.

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -1041,10 +1041,8 @@ static int ArrowAssertIncreasingInt32(struct ArrowBufferView view,
   }
 
   for (int64_t i = 1; i < view.size_bytes / (int64_t)sizeof(int32_t); i++) {
-    int32_t diff = view.data.as_int32[i] - view.data.as_int32[i - 1];
-    if (diff < 0) {
-      ArrowErrorSet(error, "[%ld] Expected element size >= 0 but found element size %ld",
-                    (long)i, (long)diff);
+    if (view.data.as_int32[i] < view.data.as_int32[i - 1]) {
+      ArrowErrorSet(error, "[%ld] Expected element size >= 0", (long)i);
       return EINVAL;
     }
   }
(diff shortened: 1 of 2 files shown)


lidavidm (approved):
Is it possible to add a test to check the overflow case?

lidavidm (approved):

PR #313, opened 2023-10-31
feat(python): Add ArrowDeviceArray extension to Python bindings

paleolimbot:
(super draft POC...cc @jorisvandenbossche )


python/src/nanoarrow/_lib_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+    for attr in ("format", "name", "flags"):
+        attr_repr = repr(getattr(schema, attr))
+        lines.append(f"{indent_str}- {attr}: {attr_repr}")
+
+    metadata = schema.metadata
+    if schema.metadata is None:
+        lines.append(f"{indent_str}- metadata: NULL")
+    else:
+        lines.append(f"{indent_str}- metadata: NULL")
```
jorisvandenbossche, line 40:
```suggestion
        lines.append(f"{indent_str}- metadata:")
```

python/src/nanoarrow/_lib.pyx
```diff
@@ -474,20 +481,8 @@ cdef class Array:
         else:
             return None
 
-    def view(self):
```
jorisvandenbossche, line 477: Is there a reason you are making this into a function instead of method? (for Schema it's still a method, so this gets a bit inconsistent)

python/src/nanoarrow/_lib.pyx
```diff
@@ -903,3 +943,112 @@ cdef class ArrayStream:
...
+
+    def __dealloc__(self):
+        if self.c_array.array.release != NULL:
+          self.c_array.array.release(&self.c_array.array)
+
+    def _addr(self):
+        return <uintptr_t>&self.c_array
+
+cdef class DeviceHolder:
+    """Memory holder for an ArrowDevice
```
jorisvandenbossche, line 968: Is this ArrowDevice concept something specific from nanoarrow? (I don't find it in https://arrow.apache.org/docs/dev/format/CDeviceDataInterface.html)

jorisvandenbossche (commented):
That's a nice improvement for the repr!

python/src/nanoarrow/_lib.pyx
paleolimbot: Yes...the `Array` as defined here doesn't know if it's a CPU or GPU array, and the "view" does. I don't know that the `view()` method is a particularly good idea for the `Schema`, either...a cached property might be a better fit.

Early on you made the suggestion that the `Array` and `ArrayView` were a little confusing, name wise. I wonder if renaming the `Array` to `CArray` (+ Schema and ArrayStream) would fix some of that. Then `ArrayView` could be renamed to `Array`, which would probably make more sense to everybody.

python/src/nanoarrow/_lib.pyx
paleolimbot: Yes, it's a nanoarrow invention. It contains methods for things like "copy this buffer to another device" but it's not ABI stable, so we'd have to workshop how to register a device from another package such that `Device.resolve()` does the right thing.

python/src/nanoarrow/_lib.pyx
paleolimbot: ...the CPU or GPUness matters because the `ArrayView` will let you access buffer content, which will crash for the GPU.

jorisvandenbossche: > Early on you made the suggestion that the `Array` and `ArrayView` were a little confusing, name wise. I wonder if renaming the `Array` to `CArray` (+ Schema and ArrayStream) would fix some of that. Then `ArrayView` could be renamed to `Array`, which would probably make more sense to everybody.

I was just writing up some thoughts on that! ;) -> https://github.com/apache/arrow-nanoarrow/issues/319

python/src/nanoarrow/_lib.pyx
jorisvandenbossche: Can you add some docstring then clarifying that? (to the Device class below)

python/src/nanoarrow/_lib.pyx
```diff
@@ -964,3 +1003,121 @@ cdef class ArrayStream:
...
+    def _addr(self):
+        return <uintptr_t>&self.c_device
+
+
+cdef class Device:
+    cdef object _base
+    cdef ArrowDevice* _ptr
+
+    def __cinit__(self, object base, uintptr_t addr):
+        self._base = base,
```
jorisvandenbossche, line 1062: Is this `base` used anywhere?

python/src/nanoarrow/_lib.pyx
```diff
@@ -964,3 +1003,121 @@ cdef class ArrayStream:
...
+        self._ptr = <ArrowDevice*>addr
+
+    def array_init(self, uintptr_t array_addr, Schema schema):
```
jorisvandenbossche, line 1065:
```suggestion
    def _array_init(self, uintptr_t array_addr, Schema schema):
```

(doesn't need to be public I assume)

python/src/nanoarrow/_lib.pyx
paleolimbot: It's a strong reference to the `DeviceHolder` (like in `Array`, `Schema`, and `ArrayStream`). It's probably less confusing to just use the `Device` for this (unlike `Array` and `Schema`, there's never any other base than a `DeviceHolder`).

paleolimbot: Right...in making that change I remembered why this is like this: the CPU device is statically allocated in C, so there's no `base`/`DeviceHolder`. But other types of device implementations might not (it's not totally solved in the C library yet).

PR #318, opened 2023-11-16
feat(python): Support the PyCapsule protocol

jorisvandenbossche:
First commit with just importing for now. Will do exporting next.

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -42,30 +42,6 @@ def c_version():
     return ArrowNanoarrowVersion().decode("UTF-8")
 
 
-cdef inline void ArrowSchemaMove(ArrowSchema* src, ArrowSchema* dst):
-    """
-    Move the contents of src into dst and set src->release to NULL
-    """
-    memcpy(dst, src, sizeof(ArrowSchema))
-    src.release = NULL
-
-
-cdef inline void ArrowArrayMove(ArrowArray* src, ArrowArray* dst):
-    """
-    Move the contents of src into dst and set src->release to NULL
-    """
-    memcpy(dst, src, sizeof(ArrowArray))
-    src.release = NULL
-
-
-cdef inline void ArrowArrayStreamMove(ArrowArrayStream* src, ArrowArrayStream* dst):
-    """
-    Move the contents of src into dst and set src->release to NULL
-    """
-    memcpy(dst, src, sizeof(ArrowArrayStream))
-    src.release = NULL
-
-
 cdef class SchemaHolder:
     """Memory holder for an ArrowSchema
 
@@ -238,16 +214,10 @@ cdef class Schema:
             A valid PyCapsule with name 'arrow_schema' containing an
             ArrowSchema pointer.
         """
-        cdef:
-            ArrowSchema* c_schema
-            Schema out
-
-        c_schema = <ArrowSchema*> PyCapsule_GetPointer(schema_capsule, 'arrow_schema')
-
-        out = Schema.allocate()
-        ArrowSchemaMove(c_schema, out._ptr)
-
-        return out
+        return Schema(
+            schema_capsule,
+            <uintptr_t>PyCapsule_GetPointer(schema_capsule, 'arrow_schema')
+        )
 
     def _addr(self):
         return <uintptr_t>self._ptr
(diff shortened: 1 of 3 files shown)


python/src/nanoarrow/_lib.pyx
```diff
@@ -428,6 +477,33 @@ cdef class Array:
...
+    @staticmethod
+    def _import_from_c_capsule(schema_capsule, array_capsule):
+        """
+        Import from a ArrowSchema and ArrowArray PyCapsule tuple.
+
+        Parameters
+        ----------
+        schema : PyCapsule
+            A valid PyCapsule with name 'arrow_array' containing an
+            ArrowArray pointer.
```
paleolimbot, line 489: This seems like it should be updated?

python/src/nanoarrow/lib.py
```diff
@@ -22,9 +22,10 @@ def schema(obj):
     if isinstance(obj, Schema):
         return obj
 
-    # Not particularly safe because _export_to_c() could be exporting an
-    # array, schema, or array_stream. The ideal
-    # solution here would be something like __arrow_c_schema__()
+    if hasattr(obj, "__arrow_c_schema__"):
+        return Schema._import_from_c_capsule(obj.__arrow_c_schema__())
+
+    # for pyarrow < 14.0
```
paleolimbot, line 28: At some point we should maybe remove this logic (or do stricter check...maybe using `type().__name__` or something). A name-based check would support older pyarrow, anyway...I think it would be reasonable for any other package that were using `_export_to_c()` to update and use the new dunder method to be compatible here.

python/tests/test_capsules.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def __arrow_c_array__(self):
+        return self.array.__arrow_c_array__()
+
+
+class StreamWrapper:
+    def __init__(self, stream):
+        self.stream = stream
+
+    def __arrow_c_stream__(self):
```
paleolimbot, line 43:
```suggestion
    def __arrow_c_stream__(self, schema=None):
```

(right?)

python/tests/test_capsules.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def __arrow_c_schema__(self):
+        return self.schema.__arrow_c_schema__()
+
+
+class ArrayWrapper:
+    def __init__(self, array):
+        self.array = array
+
+    def __arrow_c_array__(self):
```
paleolimbot, line 35:
```suggestion
    def __arrow_c_array__(self, schema=None):
```

python/tests/test_capsules.py
```diff
@@ -0,0 +1,81 @@
...
+
+class SchemaWrapper:
+    def __init__(self, schema):
+        self.schema = schema
+
+    def __arrow_c_schema__(self):
+        return self.schema.__arrow_c_schema__()
+
+
+class ArrayWrapper:
```
paleolimbot, line 31: `Array(Stream)Wrapper` should also implement `__arrow_c_schema__`, right? Even if not used here, it might be good for future readers who end up here to see the example.

python/src/nanoarrow/_lib.pyx
```diff
@@ -200,6 +227,28 @@ cdef class Schema:
...
+            A valid PyCapsule with name 'arrow_schema' containing an
+            ArrowSchema pointer.
+        """
+        cdef:
+            ArrowSchema* c_schema
+            Schema out
+
+        c_schema = <ArrowSchema*> PyCapsule_GetPointer(schema_capsule, 'arrow_schema')
+
+        out = Schema.allocate()
```
paleolimbot, line 247: I wonder if what you want here is more like:

```
cdef Array array = Schema(schema_capsule, <uintptr_t>PyCapsule_GetPointer(schema_capsule, 'arrow_schema'))
```

(i.e., no need to "move" here...`base` can be a capsule, which does basically the same thing as the `SchemaHolder`)

paleolimbot (commented):
Thank you for working on this! I think you probably want to use the `Schema`, `Array`, and `ArrayStream` constructors with the capsule as the `base` instead of "moving" into the "holder" helpers. We might be able to get rid of the holders entirely now that the capsules are defined (once there's code for allocating them).

jorisvandenbossche:
> I think you probably want to use the `Schema`, `Array`, and `ArrayStream` constructors with the capsule as the `base` instead of "moving" into the "holder" helpers.

That's fine for me as well, no strong opinion here. I actually started that way, but thought that moving the struct as consumer so we own the memory would be considered good practice (nanoarrow is of course a bit of a peculiar consumer). But actually for nested ones that still only moves the top-level one, so it's not that this guarantees anything ..

jorisvandenbossche:
Indeed, the PyCapsule also ensures proper deallocation like the Holder class, so will try a variant with keeping the capsule around as the `base`.

python/tests/test_capsules.py
jorisvandenbossche: Yes, "Will do exporting next."

python/src/nanoarrow/_lib.pyx
```diff
@@ -30,8 +30,11 @@ be literal and stay close to the structure definitions.
 from libc.stdint cimport uintptr_t, int64_t
 from cpython.mem cimport PyMem_Malloc, PyMem_Free
 from cpython.bytes cimport PyBytes_FromStringAndSize
+from cpython.pycapsule cimport PyCapsule_GetPointer, PyCapsule_New
 from cpython cimport Py_buffer
 from nanoarrow_c cimport *
+from libc.string cimport memcpy
```
paleolimbot, line 36: I think this can be removed now?

python/src/nanoarrow/_lib.pyx
```diff
@@ -428,6 +447,33 @@ cdef class Array:
         self._ptr = <ArrowArray*>addr
         self._schema = schema
 
+    @staticmethod
+    def _import_from_c_capsule(schema_capsule, array_capsule):
```
paleolimbot, line 451: I think the ordering everywhere else is `array`, `schema`?

paleolimbot (commented):
Just a few details!

PR #321, opened 2023-11-22
feat: Add Type/Field/Schema JSON writer for integration testing JSON

paleolimbot:
Adds methods to the `TestingJSONWriter` to write `Schema`, `Field`, and `Type`. There are some unsupported types but I'd like to get some end-to-end bits working before rounding out type support!

### DIFF
--- a/src/nanoarrow/nanoarrow_testing.hpp
+++ b/src/nanoarrow/nanoarrow_testing.hpp
@@ -51,7 +51,25 @@ class TestingJSONWriter {
       return EINVAL;
     }
 
-    return ENOTSUP;
+    // Write fields
+    out << R"("fields": )";
+    if (schema->n_children == 0) {
+      out << "[]";
+    } else {
+      out << "[";
+      NANOARROW_RETURN_NOT_OK(WriteField(out, schema->children[0]));
+      for (int64_t i = 1; i < schema->n_children; i++) {
+        out << ", ";
+        NANOARROW_RETURN_NOT_OK(WriteField(out, schema->children[i]));
+      }
+      out << "]";
+    }
+
+    // Write metadata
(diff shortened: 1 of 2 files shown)


src/nanoarrow/nanoarrow_testing_test.cc
```diff
@@ -288,3 +306,300 @@ TEST(NanoarrowTestingTest, NanoarrowTestingTestBatch) {
...
+
+  TestWriteJSON(
+      [](ArrowSchema* schema) {
+        ArrowSchemaInit(schema);
+        NANOARROW_RETURN_NOT_OK(
+            ArrowSchemaSetTypeUnion(schema, NANOARROW_TYPE_DENSE_UNION, 0));
+        return NANOARROW_OK;
+      },
+      [](ArrowArray* array) { return NANOARROW_OK; }, &WriteTypeJSON,
+      R"({"name": "union", "mode": "DENSE", "typeIds": []})");
```
pitrou, line 604: Should perhaps test with some child fields and type ids?

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -42,6 +42,100 @@ namespace testing {
...
+    ArrowSchemaView view;
+    NANOARROW_RETURN_NOT_OK(ArrowSchemaViewInit(&view, (ArrowSchema*)field, nullptr));
+
+    out << "{";
+
+    // Write schema->name (may be null)
+    if (field->name == nullptr) {
+      out << R"("name": null)";
+    } else {
+      out << R"("name": ")" << field->name << R"(")";
```
pitrou, line 91: Ideally we should escape reserved characters in the field name...

PR #322, opened 2023-11-23
feat: Add integration testing JSON reader

paleolimbot:
This PR adds a `TestingJSONReader`. It uses https://github.com/nlohmann/json to do the parsing. This doesn't implement parsing a record batch or array yet, and there are a few missing types (the same missing types as the writer). I'd like to do those as a follow-up.

The testing for this is mostly roundtrip testing, which is maybe not ideal but also maybe OK because the writer is an independent implementation. When the ability to compare schemas for equality gets added that can be tacked on to the roundtrip test utility function as well.

### DIFF
--- a/src/nanoarrow/nanoarrow_testing.hpp
+++ b/src/nanoarrow/nanoarrow_testing.hpp
@@ -20,6 +20,11 @@
 
 #include "nanoarrow.hpp"
 
+#if defined(NANOARROW_TESTING_WITH_NLOHMANN_JSON)
+#include <nlohmann/json.hpp>
+using json = nlohmann::json;
+#endif
+
 #ifndef NANOARROW_TESTING_HPP_INCLUDED
 #define NANOARROW_TESTING_HPP_INCLUDED
 
@@ -612,6 +617,16 @@ class TestingJSONWriter {
     std::streamsize previous_precision_;
   };
 };
+
+#if defined(NANOARROW_TESTING_WITH_NLOHMANN_JSON)
+
+/// \brief Writer for the Arrow integration testing JSON format
+class TestingJSONReader {
+ public:
+  ArrowErrorCode ReadField(const std::string& value) { return ENOTSUP; }
+};
+
+#endif
 
 /// @}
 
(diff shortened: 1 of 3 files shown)


src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -613,6 +618,458 @@ class TestingJSONWriter {
   };
 };
 
+#if defined(NANOARROW_TESTING_WITH_NLOHMANN_JSON)
+
+/// \brief Reader for the Arrow integration testing JSON format
+class TestingJSONReader {
```
pitrou, line 624: I would recommend to only put declarations in this header file and move implementations into a .cpp.

pitrou: Also, since this is entirely guarded by a `#ifdef`, perhaps it should go into separate files?

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -613,6 +618,458 @@ class TestingJSONWriter {
...
+    for (int64_t i = 0; i < schema->n_children; i++) {
+      NANOARROW_RETURN_NOT_OK(SetField(schema->children[i], children[i], error));
+    }
+
+    NANOARROW_RETURN_NOT_OK(SetMetadata(schema, value["metadata"], error));
+
+    // Validate!
+    ArrowSchemaView schema_view;
+    NANOARROW_RETURN_NOT_OK(ArrowSchemaViewInit(&schema_view, schema, error));
+
```
felipecrv, line 737: ✂️ blank line

pitrou (commented):
I haven't checked the details but this looks good in general.

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -16,10 +16,15 @@
 // under the License.
 
 #include <iostream>
+#include <sstream>
 #include <string>
 
 #include "nanoarrow.hpp"
 
+#if defined(NANOARROW_TESTING_WITH_NLOHMANN_JSON)
```
bkietz, line 24: Is this really something we can disable? In cmake it looks like we always depend on it

felipecrv (approved):

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: Great point! (Removed!)

Issue #215, opened 2023-06-08
Const-correctness


genemerewether:
Hi, is there a policy / upcoming work on `const` versions of API functions, where reasonable?

i.e. `const` accessors for `internal::Unique`, `const` pointers in function prototypes where the underlying data is not modified, or where a const flavor of e.g. `ArrowArrayView` could be used instead?

paleolimbot:
I honestly just haven't thought about it! Most current usage/testing is in C right now where const-ness isn't quite as baked in to the language.

One first step is could be to go through the C API functions and declare pointers that are not modified as `const`...there are quite a few of them but mostly that lives in documentation at this point and not as `const` declarations.

paleolimbot:
Just a note that I'm bumping this to 0.4 because it would result in a lot of compiler warnings for any project that was using a development version between 0.2 and 0.3 (and because the 0.3 release is forthcoming!).

This has come up a few times in existing usage...declaring input pointers (e.g., `ArrowSchema*`) in the C API as `const ArrowSchema*` is also a helpful way to communicate that the function will not take ownership of the schema.

PR #325, opened 2023-11-27
feat: Add integration testing reader for column and batch

paleolimbot:


src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -651,6 +652,33 @@ class TestingJSONReader {
     }
   }
 
+  ArrowErrorCode ReadColumn(const std::string& value, const ArrowSchema* schema,
```
pitrou, line 655: What is `value`? Is it the column name? The JSON key name? Something else?

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -651,6 +652,33 @@ class TestingJSONReader {
...
+      NANOARROW_RETURN_NOT_OK(
+          ArrowArrayInitFromSchema(array.get(), const_cast<ArrowSchema*>(schema), error));
+
+      // Parse the JSON into the array
+      NANOARROW_RETURN_NOT_OK(SetArrayColumn(obj, array_view.get(), array.get(), error));
+
+      // Return the result
+      ArrowArrayMove(array.get(), out);
+      return NANOARROW_OK;
+    } catch (std::exception& e) {
```
pitrou, line 676: What is supposed to throw here? If it's only `json::parse`, can you narrow the scope of the `try` block?

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -1053,6 +1081,390 @@ class TestingJSONReader {
...
+    NANOARROW_RETURN_NOT_OK(
+        Check(value.is_object(), error, "Expected Column to be a JSON object"));
+
+    // Check + resolve name early to generate better error messages
+    NANOARROW_RETURN_NOT_OK(
+        Check(value.contains("name"), error, "Column missing key 'name'"));
+
+    const auto& name = value["name"];
+    NANOARROW_RETURN_NOT_OK(Check(name.is_null() || name.is_string(), error,
+                                  "Column name must be string or null"));
```
pitrou, line 1096: Do we expect to receive a null column name? This seems a bit weird.

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -1053,6 +1081,390 @@ class TestingJSONReader {
...
+                                    error_prefix + "children has incorrect size"));
+
+      for (int64_t i = 0; i < array_view->n_children; i++) {
+        NANOARROW_RETURN_NOT_OK(SetArrayColumn(children[i], array_view->children[i],
+                                               array->children[i], error, error_prefix));
+      }
+    }
+
+    // Build buffers
+    for (int i = 0; i < 3; i++) {
```
pitrou, line 1124: Why the hardcoded 3?

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -1053,6 +1081,390 @@ class TestingJSONReader {
...
+          case NANOARROW_TYPE_BINARY:
+            return SetBuffersBinary<int32_t>(data, ArrowArrayBuffer(array, buffer_i - 1),
+                                             buffer, error);
+          case NANOARROW_TYPE_LARGE_BINARY:
+            return SetBuffersBinary<int64_t>(data, ArrowArrayBuffer(array, buffer_i - 1),
+                                             buffer, error);
+          case NANOARROW_TYPE_FIXED_SIZE_BINARY:
+            return SetBuffersBinary<int64_t>(
+                data, nullptr, buffer, error,
+                array_view->layout.element_size_bits[buffer_i] / 8);
```
pitrou, line 1251: It would probably be more readable if there was a separate helper function for fixed-size binary?

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -651,6 +652,33 @@ class TestingJSONReader {
...
 
+  ArrowErrorCode ReadColumn(const std::string& value, const ArrowSchema* schema,
+                            ArrowArray* out, ArrowError* error = nullptr) {
+    try {
+      auto obj = json::parse(value);
+
+      // ArrowArrayView to enable validation
+      nanoarrow::UniqueArrayView array_view;
+      NANOARROW_RETURN_NOT_OK(ArrowArrayViewInitFromSchema(
+          array_view.get(), const_cast<ArrowSchema*>(schema), error));
```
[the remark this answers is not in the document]
paleolimbot: Definitely! #215 is the issue I was planning for that one (there are a lot of places where read-only input arguments should be marked `const`).

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: I think `name` MAY be null (at least in the C data spec: https://arrow.apache.org/docs/format/CDataInterface.html#c.ArrowSchema.name ), so the writer serializes `name == NULL` as a JSON null (rather than omitting or setting to `""`.

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: I think any usage of the `json` object methods *can* throw (although I try to avoid it with `Check()`). In the medium-likely event I forgot something it's probably more useful if the caller of `ReadColumn()` doesn't have to handle them?

src/nanoarrow/nanoarrow_testing.hpp
pitrou: Ok. I don't expect this in actual JSON integration files, but it's also not a problem if you accept it.

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: Done!

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: I added a define for this (`NANOARROW_MAX_FIXED_BUFFERS`). The `ArrowArrayView` and `ArrowLayout` define fixed-length arrays for buffer things and it's a pretty common pattern to loop over all three and switch on `layout.buffer_type[i]`.

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: Good call, I updated to catch just `json::exception`.

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: Good call, I updated this + added documentation stubs!

python/src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -0,0 +1,1508 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
pitrou, line 1: Did you add this file by mistake?

PR #328, opened 2023-11-29
feat: Add batch reader and data file read/write to/from ArrowArrayStream

paleolimbot:


src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -46,6 +46,53 @@ namespace testing {
...
+      NANOARROW_RETURN_NOT_OK(stream->get_next(stream, array.get()));
+      if (array->release == nullptr) {
+        break;
+      }
+
+      out << ", ";
+      NANOARROW_RETURN_NOT_OK(
+          ArrowArrayViewSetArray(array_view.get(), array.get(), nullptr));
+      NANOARROW_RETURN_NOT_OK(WriteBatch(out, schema.get(), array_view.get()));
+    }
```
bkietz, line 86: IIUC, we need to check if the first batch is released too. As a nit, I've folded this into a single loop:
```suggestion
    std::string sep;
    do {
      NANOARROW_RETURN_NOT_OK(stream->get_next(stream, array.get()));
      if (array->release == nullptr) {
        break;
      }
      NANOARROW_RETURN_NOT_OK(
          ArrowArrayViewSetArray(array_view.get(), array.get(), nullptr));
      out << sep;
      sep = ", ";
      NANOARROW_RETURN_NOT_OK(WriteBatch(out, schema.get(), array_view.get()));
      array.reset();
    } while (true);
```

bkietz: I'd recommend adding a test for empty streams too

EpsilonPrime (approved):
I'm not as familiar with this codebase so I had some questions about some of the comments.  The code itself looks on point.

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -660,6 +758,34 @@ class TestingJSONReader {
     }
   }
 
+  /// \brief Read JSON representing a RecordBatch
+  ///
+  /// Read a JSON object in the form `{"count": 123, "columns": [...]}`, propagating `out`
+  /// on success.
+  ArrowErrorCode ReadBatch(const std::string& batch_json, const ArrowSchema* schema,
```
bkietz, line 765: When would this be used? When reading a json file the parsing will all be done up front and we won't see individual batch strings, so won't SetArrayBatch be sufficient?

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: Good catch 😬 ! (Test added!)

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -46,6 +46,53 @@ namespace testing {
...
+
+      out << ", ";
+      NANOARROW_RETURN_NOT_OK(
+          ArrowArrayViewSetArray(array_view.get(), array.get(), nullptr));
+      NANOARROW_RETURN_NOT_OK(WriteBatch(out, schema.get(), array_view.get()));
+    }
+
+    out << "]";
+
+    // Dictionaries errored at the schema stage
```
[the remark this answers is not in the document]
paleolimbot: I updated this comment to hopefully be more clear that the dictionaries are only empty right now because they're not supported by the writer yet. I'm hoping to get something end-to-end before filling out dictionaries/a few other missing types!

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: It's pretty much 100% to test `SetArrayBatch()`, although it's true that there is less to test in `SetArrayBatch()` than (say) `SetArrayColumn()`.

src/nanoarrow/nanoarrow_testing.hpp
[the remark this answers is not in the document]
paleolimbot: Ben had the same comment...hopefully this is more clear now! (I just haven't implemented dictionaries yet)

src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -46,6 +46,53 @@ namespace testing {
 /// \brief Writer for the Arrow integration testing JSON format
 class TestingJSONWriter {
  public:
+  /// \brief Write an ArrowArrayStream as a data file to out
```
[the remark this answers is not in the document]
paleolimbot: Hopefully this is more clear now! The integration test JSON docs don't have a nicely named object for the JSON object that sits at the root of what they call a "data file" (whereas they do for things like RecordBatch and Schema).

src/nanoarrow/nanoarrow_testing.hpp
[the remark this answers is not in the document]
bkietz: Alright, but since there aren't any dictionaries we should probably leave the dictionaries absent instead of empty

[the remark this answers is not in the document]
paleolimbot: Right! Sorry, I missed that part of the comment 😬

src/nanoarrow/nanoarrow_testing.hpp
paleolimbot: I think you are right! I am going to defer reorganizing until I get closer to actual integration tests...it is not the only organizational change that needs to happen (e.g., it also maybe shouldn't be a 1700 line header) and I would like to get a few more pieces in place before the final shuffle 🙂

bkietz (approved):

Issue #187, opened 2023-04-28
Show examples of building Arrow C++ structures


wjones127:
For users coming from Arrow C++ / PyArrow, it might not be obvious what nanoarrow structures to create so they can export. Depending on the context, an `ArrowSchema` can represent a data type, a field, or a schema. Similarly, an `ArrowArray` can represent an array or record batch (tabular). To start, we should show the correspondence between Nanoarrow structs and Arrow C++ types.

| Nanoarrow | Arrow C++ |
|----|----|
| `ArrowArray` | `arrow::Array` |
| `ArrowArray` where type is struct | `arrow::RecordBatch` |
| `std::vector<ArrowArray>` | `arrow::ChunkedArray` |
| `std::vector<ArrowArray>` where type is struct | `arrow::Table` |
| `ArrowSchema` | `arrow::DataType` |
| `ArrowSchema` | `arrow::Field` |
| `ArrowSchema` | `arrow::Schema` |

Then we may also want recipes for:

 * How to build a struct array in nanoarrow, and export as an `arrow::RecordBatch`
 * How to build a record batch reader, and export as an `arrow::Table`.

paleolimbot:
That's a great idea! That mapping is definitely not obvious unless you've spent a lot of time with the C data interface.

eddelbuettel:
I was wondering if you we could bring this back to the fore?  I am currently loosing my mind as a simple 'call from R' wrapper for the (nice) `linesplitter` example works just fine when I start in R and pass an external pointer down:

```c++
// Plain Interface
// [[Rcpp::export]]
bool linesplit_from_R_plain(const std::string str, SEXP sxparr) {
    // We get an R-created 'nanoarrow_array', an S3 class around an external pointer
    if (!Rf_inherits(sxparr, "nanoarrow_array"))
        Rcpp::stop("Expected class 'nanoarrow_array' not found");

    // It is a straight up external pointer so we can use R_ExternalPtrAddr()
    struct ArrowArray* arr = (struct ArrowArray*)R_ExternalPtrAddr(sxparr);

    auto res = linesplitter_read(str, arr);
    return true;
}
```

But when I want to start from C++, I get lost somewhere build the `nanoarrow`-compliant external pointer up by itself.  It all works fine when I cheat (as the `nanoarrow` package also does in places) and call into R:

```c++
// res <- linesplit_from_cpp("the\nquick\nbrown\nfox");
// print(res);
// print(arrow::Array$create(res))
//
[...]

PR #332, opened 2023-12-04
feat(r): Provide LinkingTo headers for extension packages

paleolimbot:
In developing geoarrow, an R extension that imports/exports a number of Arrow C data structures wrapped by nanoarrow S3 objects, it has become apparent that the sanitize and allocate operations are non-trivial and basically have to be copied in every package that wants to import/export nanoarrow things. @eddelbuettel has run up against some valid use-cases as well! https://github.com/eddelbuettel/linesplitter/issues/1 , https://github.com/apache/arrow-nanoarrow/issues/187

This PR makes the definition of how Arrow C Data interface objects are encoded as R external pointers available as a public header (such that downstream packages can `LinkingTo: nanoarrow` and `#include <nanoarrow/r.h>`. I think the initial target will just be allocate an owning external pointer and sanitize an input SEXP.

@eddelbuettel Are there any other operations that are blocking any of your projects that would be must-haves in this header?

(I know it's missing array_stream...I forgot about the ability to supply R finalizers and so it's a slightly more complicated change)

### DIFF
--- a/r/inst/include/nanoarrow/r.h
+++ b/r/inst/include/nanoarrow/r.h
@@ -176,20 +176,45 @@ static inline SEXP nanoarow_array_stream_owning_xptr(void);
 /// \brief Ensure an input SEXP points to an initialized ArrowSchema
 ///
 /// This function will always return an ArrowSchema pointer that can be safely
-/// consumed or raise an error via Rf_error().
+/// consumed or raise an error via Rf_error(). This is intended to be used to
+/// sanitize an *input* ArrowSchema.
 static inline struct ArrowSchema* nanoarrow_schema_from_xptr(SEXP schema_xptr);
+
+/// \brief Ensure an output SEXP points to an uninitialized ArrowSchema
+///
+/// This function will always return an ArrowSchema pointer that can be safely
+/// used as an output argument or raise an error via Rf_error(). This is intended
+/// to be used to sanitize an *output* ArrowSchema allocated from R or elsewhere.
+static inline struct ArrowSchema* nanoarrow_output_schema_from_xptr(SEXP schema_xptr);
 
 /// \brief Ensure an input SEXP points to an initialized ArrowArray
 ///
 /// This function will always return an ArrowArray pointer that can be safely
-/// consumed or raise an error via Rf_error().
+/// consumed or raise an error via Rf_error(). This is intended to be used to
+/// sanitize an *input* ArrowArray.
 static inline struct ArrowArray* nanoarrow_array_from_xptr(SEXP array_xptr);
+
+/// \brief Ensure an output SEXP points to an uninitialized ArrowArray
+///
+/// This function will always return an ArrowArray pointer that can be safely
+/// used as an output argument or raise an error via Rf_error(). This is intended
+/// to be used to sanitize an *output* ArrowArray allocated from R or elsewhere.
+static inline struct ArrowArray* nanoarrow_output_array_from_xptr(SEXP array_xptr);
 
 /// \brief Ensure an input SEXP points to an initialized ArrowArrayStream
 ///
 /// This function will always return an ArrowArrayStream pointer that can be safely
-/// consumed or raise an error via Rf_error().
+/// consumed or raise an error via Rf_error(). This is intended to be used to
+/// sanitize an *input* ArrowArrayStream.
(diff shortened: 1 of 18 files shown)


eddelbuettel:
We are making use of Arrow exports from C++ using just the `void*` pointer.  What we have was written before / without `nanoarrow` but I have started to bring it on the R bindings site.  There are a number of things there I'd like to revisit / extend. From my looking around at some of the packages I see using `nanoarrow` (including `adbc*` and `duckdb`) I have the feeling that there are commonalities: those packages may benefit too. 

I quite like `nanoarrow` but (if I may) also find it really frustrating at times because it essentially is three somewhat distinct and non-overlapping parts.  First, the vendorable header / c file pair which are great.  Then the R package which is nice for some things from R but which, frustratingly does not (did not?) make any tools available for re-use in conjunction with the vendorable part.  And then there is the budding Python package which may one day cover what the R package does today, and also provide object creation help.

Now this PR looks promising/  I will look more closely at it and then release 0.4.0.  This clearly seems to move in the right direction so three cheers for that!   If you want to brainstorm / chat for a few minutes I am sure we can find some time around the two hours of timezone difference.

eddelbuettel:
FYI I just checked in on this.  Rebases nearly perfectly (one edit of a comment) and the `clang-format` call is simple too.

eddelbuettel:
No problem. My curiosity was triggered when I saw the :x: there but it is really just `clang-format` being itself on one line.

eddelbuettel:
It needs a one-liner from clang-format.

r/inst/include/nanoarrow/r.h
```diff
@@ -0,0 +1,353 @@
...
+/// pointer to a non-R library, the ownership of the structure must *not* have such SEXP
+/// dependencies. The nanoarrow R package can wrap such an SEXP dependency into a
+/// self-contained thread-safe release callback via nanoarrow_pointer_export() that
+/// manages the SEXP dependency using a preserve/release mechanism similar to
+/// R_PreserveObject()/ R_ReleaseObject().
+///
+/// The "tag" of an external pointer to an ArrowArray must be R_NilValue or an external
+/// pointer to an ArrowSchema that may be used to interpret the pointed-to ArrowArray. The
+/// "tag" of a nanoarrow external pointer to an ArrowSchema or ArrowArrayStream is
+/// reserved for future use and must be R_NilValue.
```
eddelbuettel, line 62: All the above is quite good and helpful, and I wish I would have had that earlier.  

Some of it is still a little vague (ie if this is different and distinguishable from Arrow, why is that so and where are some example use cases -- my use of nanoarrow has been to 'be like Arrow' without requiring and I do not seem alone in this).

Let me chew over it some more -- but it looks like a nice step forward.  So thanks for that!

eddelbuettel:
The (arguably _very_ minimal) test of rewriting [linesplitter](https://github.com/eddelbuettel/linesplitter) to use `nanoarrow/r.h` via `LinkingTo` in lieu of a vendored `nanoarrow.hpp` passes so that is good.

eddelbuettel:
Switched a few more local copied / variants to what `nanoarrow/r.h` now exposes and it is mostly good.  So I'd stay you can merge this and we can always refine. 

I do have one test function though where I currently do

```c++
    auto sxpsch = nanoarrow_schema_owning_xptr();      				// allocates and wraps as XP
    auto sch = nanoarrow_schema_from_xptr(sxpsch);
    ArrowSchemaInitFromType(sch, NANOARROW_TYPE_STRING);
```

that ends in error with `nanoarrow_schema() has already been released`.  Have not poked much yet.

eddelbuettel:
Hah!  These subtle 'documentation bugs' are a wee bit tricky and one of the reasons I <cough, cough> blegged you to chip in with a reference example for something as simple as the [linesplitter repo](https://github.com/eddelbuettel/linesplitter).

I still have high hopes for all this as it moves in the right direction. One goal is to be able to start from a pair of `nanoarrow::UniqueArray` and `nanoarrow::UniqueSchema`, operate on them in C++ and then towards the end return them to R via R-specific helpers ... and do the same to Python with Python-specific helpers.   I know Python is behind but some small steps would help me convince me colleagues that this is a bet we can make in a mulitlingual repo.

PR #334, opened 2023-12-11
chore(extensions/nanoarrow_ipc): Add golden file testing JSON tests to existing arrow-testing tests

paleolimbot:
This PR adds a "golden file" integration test to the existing suite of tests that read files from the arrow-testing repo. The existing test read the IPC stream file using nanoarrow's IPC reader and Arrow C++'s IPC reader and used Arrow C++ to check equality. The added test is more similar to the golden file test described in the integration testing section of the documentation (read IPC, read testing JSON, use nanoarrow testing utils to check equality). This involved some refactoring to re-use the infrastructure properly for both types of tests.

Originally I wrote a dedicated executable and some bash to do these kinds of tests ( https://gist.github.com/paleolimbot/ec2a2067198f0de1901c107c783d3b26 ); however, integrating it into the existing tests is cleaner and makes it easier to run them with valgrind.

### DIFF
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
@@ -269,7 +269,7 @@ class TestFile {
 
     nanoarrow::UniqueArrayStream json_stream;
     int result = GetArrowArrayStreamCheckJSON(dir_prefix, json_stream.get(), &error);
-    // Skip instead of faile for ENOTSUP
+    // Skip instead of fail for ENOTSUP
     if (result == ENOTSUP) {
       GTEST_SKIP() << "File contains type(s) not supported by the testing JSON reader: "
                    << error.message;
--- a/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
+++ b/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
@@ -131,26 +131,12 @@ class TestFile {
   }
 
   // Create an arrow::io::InputStream wrapper around an ArrowBuffer
-  static void BufferInputStream(ArrowBuffer* src, std::shared_ptr<io::InputStream>* dst) {
-    // Read the same data with Arrow C++.
+  static arrow::Result<std::shared_ptr<io::InputStream>> BufferInputStream(
+      ArrowBuffer* src) {
     auto content_copy_wrapped = Buffer::Wrap<uint8_t>(src->data, src->size_bytes);
     auto buffer_reader = std::make_shared<io::BufferReader>(content_copy_wrapped);
-
-    // Support Arrow 9.0.0 for Fedora and Centos7 images
-#if ARROW_VERSION_MAJOR >= 10
-    auto maybe_input_stream =
-        io::RandomAccessFile::GetStream(buffer_reader, 0, content_copy_wrapped->size());
-    if (!maybe_input_stream.ok()) {
-      GTEST_FAIL() << maybe_input_stream.status().message();
-    }
-
-    std::shared_ptr<io::InputStream> input_stream = maybe_input_stream.ValueUnsafe();
-#else
-    std::shared_ptr<io::InputStream> input_stream =
-        io::RandomAccessFile::GetStream(buffer_reader, 0, content_copy_wrapped->size());
-#endif
-
-    *dst = input_stream;
+    return io::RandomAccessFile::GetStream(buffer_reader, 0,
+                                           content_copy_wrapped->size());
   }
 
   // Decompress gzipped buffer content (currently uses Arrow C++)
@@ -161,11 +147,14 @@ class TestFile {
       return EINVAL;
     }
 
-    std::shared_ptr<io::InputStream> gz_input_stream;
-    BufferInputStream(src, &gz_input_stream);
+    auto maybe_gz_input_stream = BufferInputStream(src);
+    if (!maybe_gz_input_stream.ok()) {
+      ArrowErrorSet(error, "%s", maybe_gz_input_stream.status().message().c_str());
+      return EINVAL;
+    }
 
     auto maybe_input =
-        io::CompressedInputStream::Make(maybe_gzip->get(), gz_input_stream);
+        io::CompressedInputStream::Make(maybe_gzip->get(), *maybe_gz_input_stream);
     if (!maybe_input.ok()) {
       ArrowErrorSet(error, "%s", maybe_input.status().message().c_str());
       return EINVAL;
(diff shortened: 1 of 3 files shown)


extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
```diff
@@ -70,31 +71,124 @@ class TestFile {
...
+                            json_content->size_bytes);
+
+    // Use testing util to populate the array stream
+    nanoarrow::testing::TestingJSONReader reader;
+    NANOARROW_RETURN_NOT_OK(reader.ReadDataFile(json_string, out, error));
+    return NANOARROW_OK;
+  }
+
+  // Read a whole file into an ArrowBuffer
+  static ArrowErrorCode ReadFileBuffer(const std::string& path, ArrowBuffer* content,
```
pitrou, line 120: This looks quite complicated. Why not use `arrow::io::ReadableFile`? You could even wrap it directly in `CompressedInputStream`.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
```diff
@@ -70,31 +71,124 @@ class TestFile {
...
+    return NANOARROW_OK;
+  }
 
-    struct ArrowIpcInputStream input;
+  // Create an arrow::io::InputStream wrapper around an ArrowBuffer
+  static arrow::Result<std::shared_ptr<io::InputStream>> BufferInputStream(
+      ArrowBuffer* src) {
+    auto content_copy_wrapped = Buffer::Wrap<uint8_t>(src->data, src->size_bytes);
+    auto buffer_reader = std::make_shared<io::BufferReader>(content_copy_wrapped);
+    return io::RandomAccessFile::GetStream(buffer_reader, 0,
```
pitrou, line 138: A `BufferReader` is already an `InputStream`, so you don't need `GetStream`...

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
```diff
@@ -70,31 +71,124 @@ class TestFile {
...
+                                           content_copy_wrapped->size());
+  }
+
+  // Decompress gzipped buffer content (currently uses Arrow C++)
+  static ArrowErrorCode UnGZIP(ArrowBuffer* src, ArrowBuffer* dst, ArrowError* error) {
+    auto maybe_gzip = arrow::util::Codec::Create(arrow::Compression::GZIP);
+    if (!maybe_gzip.ok()) {
+      ArrowErrorSet(error, "%s", maybe_gzip.status().message().c_str());
+      return EINVAL;
+    }
```
pitrou, line 148: You're writing this so often that you probably want a macro for it, IMHO.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
```diff
@@ -133,26 +227,18 @@ class TestFile {
...
-    auto content_copy_wrapped =
-        Buffer::Wrap<uint8_t>(content_copy->data, content_copy->size_bytes);
-    auto buffer_reader = std::make_shared<io::BufferReader>(content_copy_wrapped);
-
-    // Support Arrow 9.0.0 for Fedora and Centos7 images
-#if ARROW_VERSION_MAJOR >= 10
-    auto maybe_input_stream =
-        io::RandomAccessFile::GetStream(buffer_reader, 0, content_copy_wrapped->size());
+    // Read the same file with Arrow C++. Use the in-memory version to avoid
+    // requiring Arrow C++ with filesystem.
```
pitrou, line 231: You don't need the filesystem layer to use `arrow::io::ReadableFile`.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
```diff
@@ -180,6 +266,41 @@ class TestFile {
...
+        << error.message;
+
+    nanoarrow::UniqueArrayStream json_stream;
+    int result = GetArrowArrayStreamCheckJSON(dir_prefix, json_stream.get(), &error);
+    // Skip instead of fail for ENOTSUP
+    if (result == ENOTSUP) {
+      GTEST_SKIP() << "File contains type(s) not supported by the testing JSON reader: "
+                   << error.message;
+    }
+    ASSERT_EQ(result, NANOARROW_OK) << error.message;
```
pitrou, line 288: Don't you want a `ASSERT_NANOARROW_OK` macro to avoid repeating this?

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
```diff
@@ -70,31 +71,124 @@ class TestFile {
...
-    // understands
-    nanoarrow::UniqueBuffer content_copy;
-    ASSERT_EQ(ArrowBufferAppend(content_copy.get(), content->data, content->size_bytes),
-              NANOARROW_OK);
+    return NANOARROW_OK;
+  }
 
-    struct ArrowIpcInputStream input;
+  // Create an arrow::io::InputStream wrapper around an ArrowBuffer
+  static arrow::Result<std::shared_ptr<io::InputStream>> BufferInputStream(
```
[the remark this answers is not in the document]
paleolimbot: I tried this, but the functions that need an `InputStream` to which the result is passed seem to require a `std::shared_ptr<InputStream>`!

felipecrv:
You can convert the unique_ptr to shared_ptr implicitly by std::move’ing it.

The proliferation of shared_ptr everywhere is so unfortunate.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
```diff
@@ -70,31 +71,124 @@ class TestFile {
...
+      ArrowBuffer* src) {
+    auto content_copy_wrapped = Buffer::Wrap<uint8_t>(src->data, src->size_bytes);
+    auto buffer_reader = std::make_shared<io::BufferReader>(content_copy_wrapped);
+    return io::RandomAccessFile::GetStream(buffer_reader, 0,
+                                           content_copy_wrapped->size());
+  }
+
+  // Decompress gzipped buffer content (currently uses Arrow C++)
+  static ArrowErrorCode UnGZIP(ArrowBuffer* src, ArrowBuffer* dst, ArrowError* error) {
```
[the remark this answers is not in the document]
paleolimbot: Eventually I need this function to not use Arrow C++ (so that tests can build/run on platforms where Arrow C++ doesn't build); however, I added a macro to make the Arrow -> nanoarrow result propagation less bad!

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
paleolimbot: Eventually I need the "read gzipped JSON" half of this to not use Arrow C++ (so that it is possible/easier to run tests on platforms where Arrow C++ doesn't build), so I'm hesitant to lean too far into the Arrow C++ API for any piece of this test.

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
paleolimbot: Probably! It's a bit of a project to make that transformation in all the tests and hasn't been a priority so far (it only saves a few characters).

extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc
pitrou: Fair enough!

PR #337, opened 2023-12-13
feat: Add C Data integration test shared library

paleolimbot:
Based on https://github.com/apache/arrow/blob/main/cpp/src/arrow/integration/c_data_integration_internal.cc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ set(NANOARROW_VERSION_MINOR "${nanoarrow_VERSION_MINOR}")
 set(NANOARROW_VERSION_PATCH "${nanoarrow_VERSION_PATCH}")
 
 option(NANOARROW_BUILD_TESTS "Build tests" OFF)
+option(NANOARROW_BUILD_INTEGRATION_TESTS "Build integration tests" OFF)
 option(NANOARROW_BUNDLE "Create bundled nanoarrow.h and nanoarrow.c" OFF)
 option(NANOARROW_BUNDLE_AS_CPP "Bundle nanoarrow source file as nanoarrow.cc" OFF)
 option(NANOARROW_NAMESPACE "A prefix for exported symbols" OFF)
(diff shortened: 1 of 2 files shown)


pitrou:
> I haven't tested this via archery because I have no idea how to do so (the implementation names and file locations seem hard-coded?)

I think you can take a look at what we do for Rust, since it's a similar case of a non-monorepo implementation:
https://github.com/apache/arrow/blob/main/dev/archery/archery/integration/tester_rust.py

(in short: locations can be influenced with an environment variable)

src/nanoarrow/integration/c_data_integration.cc
```diff
@@ -0,0 +1,200 @@
...
+    return global_error.message;
+  }
+}
+
+// TODO
+int64_t nanoarrow_BytesAllocated() { return 0; }
+
+const char* nanoarrow_CDataIntegration_ExportSchemaFromJson(const char* json_path,
+                                                            ArrowSchema* out) {
+  ArrowErrorInit(&global_error);
```
pitrou, line 177: Does this deallocate a previously-initialized error message?

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set(NANOARROW_VERSION_MINOR "${nanoarrow_VERSION_MINOR}")
 set(NANOARROW_VERSION_PATCH "${nanoarrow_VERSION_PATCH}")
 
 option(NANOARROW_BUILD_TESTS "Build tests" OFF)
+option(NANOARROW_BUILD_INTEGRATION_TESTS "Build integration tests" OFF)
```
pitrou, line 35:
```suggestion
option(NANOARROW_BUILD_INTEGRATION_TESTS "Build cross-implementation Arrow integration tests" OFF)
```

src/nanoarrow/integration/c_data_integration.cc
```diff
@@ -0,0 +1,200 @@
...
+    return EINVAL;
+  }
+
+  return NANOARROW_OK;
+}
+
+static ArrowErrorCode ExportBatchFromJson(const char* json_path, int num_batch,
+                                          ArrowArray* out, ArrowError* error) {
+  MaterializedArrayStream data;
+  NANOARROW_RETURN_NOT_OK(MaterializeJsonFilePath(json_path, &data, error));
```
pitrou, line 124: Not very important, but ideally we'd only parse JSON batch `num_batch`, instead of parsing them all.

src/nanoarrow/integration/c_data_integration.cc
paleolimbot: It doesn't (it is basically `global_error.message[0] = '\0';`). The C++ implementation seems to use `static std::string` which I think is similar?

pitrou: But does making an error message allocate its storage dynamically?

paleolimbot: It doesn't (the global error here is a `static ArrowError`, aka `static char[1024]`).

src/nanoarrow/integration/c_data_integration.cc
paleolimbot: Good call! This makes sense (particularly after adding the tracking for bytes allocated).

PR #339, opened 2023-12-14
chore(ci): Update docker images

paleolimbot:
This fixes a few problems:

- Install `which` on archlinux, since its absence was causing R to break
- Use `pip` to install into a venv instead of running `pip` as root to install into system directories (which was broken on alpine)

### DIFF
--- a/ci/docker/centos7.dockerfile
+++ b/ci/docker/centos7.dockerfile
@@ -20,7 +20,7 @@ ARG NANOARROW_ARCH
 FROM --platform=linux/${NANOARROW_ARCH} centos:7
 
 RUN yum install -y epel-release
-RUN yum install -y git gnupg curl R gcc-c++ cmake3
+RUN yum install -y git gnupg curl R gcc-c++ gcc-gfortran cmake3
 
 # For Arrow C++. Use 9.0.0 because this version works fine with the default gcc
 RUN curl -L https://github.com/apache/arrow/archive/refs/tags/apache-arrow-9.0.0.tar.gz | tar -zxf - && \
@@ -45,7 +45,7 @@ ENV LC_ALL en_US.UTF-8
 
 # For R. Note that arrow is not installed (takes too long).
 RUN mkdir ~/.R && echo "MAKEFLAGS = -j$(nproc)" > ~/.R/Makevars
-RUN R -e 'install.packages(c("blob", "hms", "tibble", "rlang", "testthat", "tibble", "vctrs", "withr"), repos = "https://cloud.r-project.org")'
+RUN R -e 'install.packages(c("blob", "hms", "tibble", "rlang", "testthat", "tibble", "vctrs", "withr", "bit64"), repos = "https://cloud.r-project.org")'
 RUN rm -f ~/.R/Makevars
 
 ENV NANOARROW_CMAKE_OPTIONS -DArrow_DIR=/arrow/lib/cmake/Arrow
--- a/ci/docker/centos7.dockerfile
+++ b/ci/docker/centos7.dockerfile
@@ -37,7 +37,7 @@ RUN curl -L https://github.com/apache/arrow/archive/refs/tags/apache-arrow-9.0.0
 RUN python3 -m venv /venv
 RUN source /venv/bin/activate && \
     pip install --upgrade pip && \
-    pip install build Cython pytest numpy
+    pip install build Cython pytest pytest-cython numpy
 
 # Locale required for R CMD check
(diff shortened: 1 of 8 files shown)


ci/docker/ubuntu.dockerfile
```diff
@@ -34,14 +32,17 @@ RUN apt-get install -y -V ca-certificates lsb-release wget && \
...
-# has been sorted: https://github.com/sphinx-doc/sphinx/issues/11605
-RUN pip3 install pydata-sphinx-theme "sphinx<7.2.0" breathe build Cython numpy pytest pyarrow
+RUN python3 -m venv --upgrade-deps /venv
+RUN echo "source /venv/bin/activate && pip install pydata-sphinx-theme sphinx breathe build Cython numpy pytest pytest-cython pytest-cov pyarrow" | bash
+ENV NANOARROW_PYTHON_VENV "/venv"
+
+# Locale required for R CMD check
+RUN locale-gen en_US.UTF-8 && update-locale en_US.UTF-8
 
 # For R. Note that we install arrow here so that the integration tests for R run
```
assignUser, line 42: I don't see arrow R being installed?

assignUser, line 42: I don't see arrow R being installed?

ci/docker/centos7.dockerfile
```diff
@@ -20,12 +20,7 @@ ARG NANOARROW_ARCH
...
 RUN yum install -y epel-release
-RUN yum install -y git gnupg curl R gcc-c++ cmake3 python3-devel
-# On centos7/arm64, cythonize/build is required and fails with cython >= 3
-RUN pip3 install build "Cython < 3.0.0" numpy pytest
-
-RUN localedef -c -f UTF-8 -i en_US en_US.UTF-8
-ENV LC_ALL en_US.UTF-8
+RUN yum install -y git gnupg curl R gcc-c++ gcc-gfortran cmake3
 
 # For Arrow C++. Use 9.0.0 because this version works fine with the default gcc
```
assignUser, line 25: iirc we also publish an .rpm for arrow c++? Why not use that?  (can't wait for centos 7 to reach eol...)

ci/docker/centos7.dockerfile
```diff
@@ -34,13 +29,24 @@ RUN curl -L https://github.com/apache/arrow/archive/refs/tags/apache-arrow-9.0.0
...
+
+# Locale required for R CMD check
+RUN localedef -c -f UTF-8 -i en_US en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
+
 # For R. Note that arrow is not installed (takes too long).
 RUN mkdir ~/.R && echo "MAKEFLAGS = -j$(nproc)" > ~/.R/Makevars
-RUN R -e 'install.packages(c("blob", "hms", "tibble", "rlang", "testthat", "tibble", "vctrs", "withr"), repos = "https://cloud.r-project.org")'
+RUN R -e 'install.packages(c("blob", "hms", "tibble", "rlang", "testthat", "tibble", "vctrs", "withr", "bit64"), repos = "https://cloud.r-project.org")'
 RUN rm -f ~/.R/Makevars
```
assignUser, line 50: This is part of every  docker file with the minor difference of installing arrow or not, should that not be moved into a script? + use ppm to install arrow for speed?

assignUser (approved):
LGTM, some minor comments.

ci/docker/ubuntu.dockerfile
paleolimbot: I moved the comment (it happens a few lines later)!

ci/docker/centos7.dockerfile
paleolimbot: It looks like the last time this was built for arm64 was Arrow 3.0.0 (!!!). (This gets used on ARM almost entirely because it's how I debug these locally and I'm on M1 Mac). It's just a minimal build so it's not too bad.

ci/docker/centos7.dockerfile
paleolimbot: Definitely! I'm going to punt that to the next time I have to update these, though, since I need to do it for Python too. I think pak can help, as can `Config/Needs/XXX`, but I haven't quite gotten around to that level of automation yet.

I should probably be using Posit Package Manager for the other dependencies, too (although it would only help on x86 I think).

paleolimbot: (It also should be said that the native build of these images happens once a week and take 10 minutes...the QEMU ones take forever, though)

PR #341, opened 2023-12-19
chore(ci): Fix verification builds for Windows and centos7

paleolimbot:
This PR contains a few updates such that the verification job passes on Windows and centos7, and fedora.

On centos7/fedora, the builds were failing because the new test that was added needs to be compiled with `-fPIC`. I updated the verification script to always set `CMAKE_POSITION_INDEPENDENT_CODE=ON` from the top-level `cmake` invocation. This was easier than trying to get `target_compile_options()` working on the versions of cmake available by default in all the runners.

On the GitHub Actions-based verifications, the builds were failing because the cached Arrow C++ build was not built with the zlib/the gzip codec. Simply adding `ARROW_WITH_ZLIB=ON` did not work on Windows (something about a missing `.lib` file), so I just switched to using zlib directly. When the IPC reader supports compression we will need it anyway and it was a fairly minimal change.

There were also a few compiler warnings that I fixed while I had a real live Windows/MSVC session going.


dev/release/verify-release-candidate.sh
```diff
@@ -217,6 +217,7 @@ test_cmake_project() {
   show_info "Configure CMake Project"
   ${CMAKE_BIN} "${NANOARROW_SOURCE_DIR}/${2}" \
     "${@:3}" \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
```
assignUser, line 220: This works of course but out of interest what was the issue with the target version? It reads like you set the flags directly but you can also apply this option on a target basis and let cmake add the flags: `set_property(TARGET fpic-test PROPERTY POSITION_INDEPENDENT_CODE ON)`

extensions/nanoarrow_ipc/CMakeLists.txt
```diff
@@ -215,6 +215,16 @@ if(NANOARROW_IPC_BUILD_TESTS)
     fetchcontent_makeavailable(nlohmann_json)
   endif()
 
+  # zlib to decode gzipped integration testing JSON files
+  # We don't use Arrow C++ for this because building Arrow C++ with zlib
+  # is not trivial on Windows.
+  find_package(ZLIB)
+  if(NOT ZLIB_FOUND)
+    # Wrapper around FetchContent that better isolates the zlib CMakeLists.txt
```
assignUser, line 223: Yeah, separate dir scope can really help prevent global options from being modified ^^

assignUser (approved):
CMake looks good :)

dev/release/verify-release-candidate.sh
paleolimbot: I looked into this a little more deeply (specifically, trying `set_target_properties(nanoarrow_c_data_integration PROPERTIES POSITION_INDEPENDENT_CODE ON)` and the version of that you suggested as well. It seems like the solution would require the `nanoarrow` target to be position-independent, which is a bad default for actual use of nanoarrow (but a totally fine default for testing). Another option would be to `if (NANOARROW_BUILD_TESTS) set(CMAKE_POSITION_INDEPENDENT_CODE ON)` but the command-line version seems less intrusive?

FWIW I was experimenting with:

```
# docker run --platform=linux/amd64  --rm -it -v $(pwd):/nanoarrow ghcr.io/apache/arrow-nanoarrow:centos7
mkdir /nanoarrow/build && cd /nanoarrow/build
cmake3 .. -DNANOARROW_BUILD_TESTS=ON
cmake3 --build .
```

PR #352, opened 2024-01-05
Export CMake config

### LINKED ISSUE #350: Export CMake config
When trying to use this library from another CMake project, I've noticed that this project does not export CMake config.

I know that this library is meant to be vendored/bundled with my project, but in my case, this is quite undesirable.

So I've created a patch to export the config here: https://github.com/aljazerzen/arrow-nanoarrow-nix/blob/main/0001-patch.patch

I'm relatively new to CMake, so this issue is mostly about "i this a good idea" and "why not?"

aljazerzen:
Closes #350

You probably want the `.in` file somewhere else.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,13 +144,35 @@ else()
     endif()
   endif()
 
-  install(TARGETS nanoarrow DESTINATION lib)
+  install(
+    EXPORT nanoarrowTargets
+    FILE nanoarrowTargets.cmake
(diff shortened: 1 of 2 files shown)


aljazerzen:
Ummm, I'm not proficient in CMake and have also forgot what I've done here. Go ahead and close this.

vyasr:
@paleolimbot feel free to close this. I'll have a PR up soon to supersede this one and close the associated issue.

PR #353, opened 2024-01-05
ci(python): Add cibuildwheel setup for Python wheels

paleolimbot:
This PR adds a weekly/workflow dispatch job for building and testing Python wheels. This required a few housekeeping items:

- Versioning the python package. I used the approach from ADBC, which is a modified 'miniver'. Basically, just set the version as a string using a regex replace when needed.
- The bootstrap.py logic was updated to use a proper temporary directory
- Tests were updated to skip instead of fail when pyarrow/numpy are not available (because I can never remember which platforms they will or won't install on and the default cibuildwheel grid is large).
- I hadn't tested install from sdist, so a few files were missing from the manifest.

Two outstanding issues that don't necessarily need to be solved in this PR:

- The linux arm self-hosted runner not is not picking up the job? (config copied from https://github.com/apache/arrow/blob/main/.github/workflows/go.yml#L46-L66 ). It's not the end of the world to not have linux/arm64 wheels but if we already have the runner set up it seems like a good time to make sure it works.
- At least one test doesn't pass on 32-bit Windows (already fixed in #340). For now I just enabled the version tests to make sure everything built/linked properly.

### DIFF
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -17,7 +17,7 @@
 
 name: Build Python Wheels
 
-# Build wheels only on a release branch or when requested since it's expensive
+# Build wheels only on a release branch, weekly, or when requested
 on:
   pull_request:
     branches:
@@ -25,12 +25,6 @@ on:
     paths:
       - '.github/workflows/python-wheels.yaml'
   workflow_dispatch:
-    inputs:
-      version:
-        description: "Version to force on wheels"
-        required: false
-        type: string
-        default: ""
   push:
     branches:
       - 'maint-**'
@@ -45,8 +39,7 @@ jobs:
         - "ubuntu-20.04"
         - "windows-2019"
         - "macOS-11"
-        - ["self-hosted", "ubuntu-20.04", "arm64"]
-        - ["self-hosted", "macOS", "arm64"]
+        - ["self-hosted", "ubuntu-20.04", "arm"]
 
     steps:
       - uses: actions/checkout@v4
@@ -61,6 +54,8 @@ jobs:
       - name: Build wheels
         run: |
           python -m cibuildwheel --output-dir wheelhouse python
+        env:
+          CIBW_ARCHS_MACOS: x86_64 arm64
 
       - uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -87,6 +87,8 @@ jobs:
 
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.16.2
+        with:
+          package-dir: python
         env:
           CIBW_ARCHS_MACOS: x86_64 arm64
           # Optional (test suite will pass if these are not available)
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -87,10 +87,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==2.15.0
+
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.2
-        with:
-          package-dir: python
+        run: |
+          python -m cibuildwheel --output-dir wheelhouse python
         env:
           # Optional (test suite will pass if these are not available)
           # Commenting this for now because not all the tests pass yet (fixes in another PR)
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -86,7 +86,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v3
 
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.16.2
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -37,7 +37,7 @@ jobs:
     runs-on: "ubuntu-20.04"
     steps:
     - uses: actions/checkout@v4
-    - uses: actions/setup-python@v3
+    - uses: actions/setup-python@v5
     - name: Check that cmake is installed
       run: |
         cmake --version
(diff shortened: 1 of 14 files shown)


assignUser (approved):
workflow looks good rest also looks sensible but my python is mid ^^

.github/workflows/python-wheels.yaml
```diff
@@ -0,0 +1,97 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Build Python Wheels
+
+# Build wheels weekly or when requested
+on:
+  pull_request:
```
jorisvandenbossche, line 22: We could also run in on every commit in main? (I agree it's certainly good to not let it run on every commit in every PR)

```
  push:
    branches:
      - main
```

.github/workflows/python-wheels.yaml
```diff
@@ -0,0 +1,97 @@
...
+  workflow_dispatch:
+  schedule:
+    - cron: '6 0 * * 0'
+
+jobs:
+  build_sdist:
+    runs-on: "ubuntu-20.04"
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v3
```
jorisvandenbossche, line 36:
```suggestion
    - uses: actions/setup-python@v5
```

.github/workflows/python-wheels.yaml
```diff
@@ -0,0 +1,97 @@
...
+        pip install build
+
+    - name: Build sdist
+      run: |
+        cd python
+        python -m build --sdist
+
+    - name: Check install from sdist
+      run: |
+        pip install python/dist/nanoarrow-*.tar.gz
```
jorisvandenbossche, line 52: As an additional check, we could also run `twine check --strict python/dist/*`

.github/workflows/python-wheels.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+    - name: Test import
+      run: |
+        python -c "import nanoarrow; print(nanoarrow.__version__)"
+
+    - uses: actions/upload-artifact@v3
```
jorisvandenbossche, line 58:
```suggestion
    - uses: actions/upload-artifact@v4
```

.github/workflows/python-wheels.yaml
```diff
@@ -0,0 +1,97 @@
...
+    name: Build wheels on ${{ matrix.config.label }}
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      matrix:
+        config:
+          - {os: "ubuntu-20.04", label: "linux"}
+          - {os: "windows-2019", label: "windows"}
+          - {os: "macOS-11", label: "macOS"}
+          # Uncomment when apache/arrow-nanoarrow is added to the arrow runner group
+          # - {os: ["self-hosted", "arm"], label: "linux/arm64"}
```
jorisvandenbossche, line 74: BTW, we can build MacOS arm64 wheels here on github actions if we want (just not test them), that's what we do in other projects I am involved with

.github/workflows/python-wheels.yaml
```diff
@@ -0,0 +1,97 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v3
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==2.15.0
+
+      - name: Build wheels
+        run: |
+          python -m cibuildwheel --output-dir wheelhouse python
```
jorisvandenbossche, line 85:
```suggestion
      - name: Build wheels
        uses: pypa/cibuildwheel@v2.16.2
```

I am not sure why the cibuildwheel README shows the manual way, but the docs recommend using the action (https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions)

.github/workflows/python-wheels.yaml
```diff
@@ -0,0 +1,97 @@
...
+          CIBW_ARCHS_MACOS: x86_64 arm64
+          # Optional (test suite will pass if these are not available)
+          # Commenting this for now because not all the tests pass yet (fixes in another PR)
+          # CIBW_BEFORE_TEST: pip install --only-binary ":all:" pyarrow numpy || pip install --only-binary ":all:" numpy || true
+          CIBW_TEST_REQUIRES: pytest
+          CIBW_TEST_COMMAND: pytest {package}/tests
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: wheels
```
jorisvandenbossche, line 96:
```suggestion
      - uses: actions/upload-artifact@v4
        with:
          name: wheels-${{ matrix.config.label }}
```

(the latest version requires unique names)

python/pyproject.toml
```diff
@@ -34,7 +34,7 @@ repository = "https://github.com/apache/arrow-nanoarrow"
 [build-system]
 requires = [
     "setuptools >= 61.0.0",
-    "setuptools-scm",
+    "versioneer[toml]==0.29",
```
jorisvandenbossche, line 37: This isn't needed if you use the copied miniver helper?

jorisvandenbossche (commented):
Looking good!

I added a whole bunch of comments, but many are also small nitpicks that are not crucial at all

.github/workflows/python-wheels.yaml
jorisvandenbossche: In fact, that's already happening (because you defined `CIBW_ARCHS_MACOS` to include both, I suppose)

jorisvandenbossche: We could also try emulation on github actions here for linux arm64 (given it's a small package, that might be possible and not too slow). Then we wouldn't need any other self-hosted runner or different CI platform.

paleolimbot: It took a second to figure it out, but it seems like the arm64 runner works!

.github/workflows/python-wheels.yaml
paleolimbot: I had to revert to the previous version because of the arm self-hosted runner (for which the stock install-python action won't work).

PR #359, opened 2024-01-11
chore(ci): Integrate Python into more CI workflows

paleolimbot:
This PR adds Python to the coverage and verify workflows (except on centos7, where the image isn't set up for a supported version of Python yet). There were also some housekeeping issues identified by various logs that I noticed in the process (e.g., fixing the Python dev version format, ensuring that CMAKE_BIN is respected in bootstrap.py).

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -36,7 +36,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.10']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
 
     steps:
       - uses: actions/checkout@v3
@@ -59,12 +59,12 @@ jobs:
           pytest python/tests -v -s
 
       - name: Run doctests
-        if: success() && matrix.python-version == '3.10'
+        if: success() && matrix.python-version == '3.12'
         run: |
           pytest --pyargs nanoarrow --doctest-modules
 
       - name: Coverage
-        if: success() && matrix.python-version == '3.10'
+        if: success() && matrix.python-version == '3.12'
         run: |
           pip uninstall --yes nanoarrow
           pip install pytest-cov Cython
@@ -79,7 +79,7 @@ jobs:
           python -m coverage xml
 
       - name: Upload coverage to codecov
-        if: success() && matrix.python-version == '3.10'
+        if: success() && matrix.python-version == '3.12'
         uses: codecov/codecov-action@v2
         with:
           files: 'python/coverage.xml'
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -62,24 +62,3 @@ jobs:
         if: success() && matrix.python-version == '3.12'
         run: |
           pytest --pyargs nanoarrow --doctest-modules
-
-      - name: Coverage
-        if: success() && matrix.python-version == '3.12'
-        run: |
-          pip uninstall --yes nanoarrow
-          pip install pytest-cov Cython
-          pushd python
-
-          # Build with Cython + gcc coverage options
-          pip install -e .
-          NANOARROW_PYTHON_COVERAGE=1 python setup.py build_ext --inplace
-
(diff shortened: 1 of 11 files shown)


ci/scripts/coverage.sh
```diff
@@ -137,6 +146,36 @@ function main() {
...
     popd
+
+    # Build + test Python package with cython/gcc coverage options
+    show_header "Build + test Python package"
+    pushd "${SANDBOX_DIR}"
+    TARGET_NANOARROW_PYTHON_DIR="${TARGET_NANOARROW_DIR}/python"
+
+    pushd "${TARGET_NANOARROW_PYTHON_DIR}"
+    python -m pip install -e .
+    NANOARROW_COVERAGE=1 python setup.py build_ext --inplace
```
jorisvandenbossche, line 157:
```suggestion
    NANOARROW_COVERAGE=1 python -m pip install -e .
```

That doesn't work? (`pip install -e .` should essentially under the hood already be calling `python setup.py build_ext --inplace`, but then through official APIs)

.github/workflows/python.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.10']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
```
WillAyd, line 39: Much of the scientific python community has already dropped Python 3.8 support per NEP 29:

https://numpy.org/neps/nep-0029-deprecation_policy.html

Up to you of course what to support, but maybe not worth going as far back as 3.8 (which is EOL this year for Python as well)

.github/workflows/verify.yaml
```diff
@@ -138,7 +138,9 @@ jobs:
           - {
               platform: "centos7",
               arch: "amd64",
-              compose_args: "-e NANOARROW_ACCEPT_IMPORT_GPG_KEYS_ERROR=1"
+              # Currently the Python on the centos7 image is 3.6, which does not support
```
WillAyd, line 141: Instead of using a base centos 7 you might also want to just use the same centos image that cibuildwheel uses for python packaging. That has quite a few versions of python installed on it

quay.io/pypa/manylinux2014_x86_64

ci/scripts/coverage.sh
paleolimbot: That is much cleaner! Thank you!

python/setup.py
```diff
@@ -65,6 +66,11 @@ def get_version(pkg_path):
     extra_link_args = []
     extra_define_macros = []
 
+# Setting the CFLAGS environment variable doesn't seem sufficient in all cases to add
```
WillAyd, line 69: Do you know where this limitation was coming into play? I've generally always had success with CFLAGS, unless activating a conda environment with a compilers package that may overwrite this. I don't _think_ that is the problem here but figured I'd mention

.github/workflows/python.yaml
paleolimbot: That's good to know! I don't envision that nanoarrow for Python will have a problem with Python 3.8 (but perhaps that's just because it doesn't implement m(any) features yet 🙂 )

python/setup.py
paleolimbot: Thanks for the heads up! I'll spend a bit more time debugging to make sure I'm not imagining this limitation. I wonder if it's also because I'm setting `extra_compile_args` to `[]` and that somehow overrides CFLAGS.

.github/workflows/verify.yaml
paleolimbot: That's good to know! I may punt on centos7 testing (or decide that it's no longer worth it) until the next version...we already test that the C library compiles on gcc 4.8 which is the main hurdle.

python/setup.py
paleolimbot: Good call...CFLAGS is totally fine 🤦

assignUser (approved):

WillAyd (approved):
Nice work!

PR #361, opened 2024-01-12
feat: Add decimal support to integration tester

paleolimbot:


WillAyd:
Very cool - this would be very helpful for arrow-adbc. Is this expected to work for Decimal256 as well?

WillAyd:
> It would probably be more useful for ADBC if it handled the decimal point and a trailing exponent

The postgres implementation in ADBC currently sends the string and decimal placement separately, so this would be helpful for that without going all the way

src/nanoarrow/utils.c
```diff
@@ -223,3 +223,186 @@ struct ArrowBufferAllocator ArrowBufferDeallocator(
...
+  int has_sign = is_negative || value.data[0] == '+';
+  value.data += has_sign;
+  value.size_bytes -= has_sign;
+
+  // Check all characters are digits that are not the negative sign
+  for (int64_t i = 0; i < value.size_bytes; i++) {
+    char c = value.data[i];
+    if (c < '0' || c > '9') {
+      return EINVAL;
+    }
```
danepitkin, line 279: this might be an unhelpful guess at micro-optimization.. but if you are trying to avoid branching, you can set a bitmask here and check for failure outside of the loop. e.g. `is_invalid &= c < '0' || c > '9';`. Maybe the compiler would optimize this for us? I wouldn't know.

paleolimbot: I'm not sure I follow...I wasn't intending to optimize anything, just to make sure we had a valid integer string before proceeding. The other option is to pass `end_ptr` instead of `NULL` into `strtoll()` and check that it parsed the entire chunk of characters. Maybe that would be less confusing?

src/nanoarrow/nanoarrow.h
```diff
@@ -280,6 +284,14 @@ void ArrowLayoutInit(struct ArrowLayout* layout, enum ArrowType storage_type);
...
 static inline struct ArrowStringView ArrowCharView(const char* value);
 
+/// \brief Sets the integer value of an ArrowDecimal from a string
+ArrowErrorCode ArrowDecimalSetIntString(struct ArrowDecimal* decimal,
+                                        struct ArrowStringView value);
+
+/// \brief Get the integer value of an ArrowDecimal as string
+ArrowErrorCode ArrowDecimalAppendIntStringToBuffer(const struct ArrowDecimal* decimal,
+                                                   struct ArrowBuffer* buffer);
+
```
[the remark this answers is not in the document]
paleolimbot: I agree that it's not great! I think the reason I didn't pick `ArrowDecimalSetFromString()` is because to me that would parse the decimal point and this function won't do that. It will also error very quickly if anybody tries to put a decimal point there and I'm game to change it.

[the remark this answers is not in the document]
paleolimbot: It floated in last night that `Digits` is probably a better term. I pushed that change but feel free to suggest something else!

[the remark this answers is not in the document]
danepitkin: Even better, I love this change!

src/nanoarrow/utils.c
```diff
@@ -223,3 +223,186 @@ struct ArrowBufferAllocator ArrowBufferDeallocator(
...
+      break;
+    }
+  }
+
+  value.data += n_leading_zeroes;
+  value.size_bytes -= n_leading_zeroes;
+
+  // Use 32-bit words for portability
+  uint32_t words32[8];
+  int n_words32 = decimal->n_words * 2;
```
pitrou, line 297: Assert that this is <= 8?

src/nanoarrow/utils.c
```diff
@@ -223,3 +223,186 @@ struct ArrowBufferAllocator ArrowBufferDeallocator(
...
+    return NANOARROW_OK;
+  }
+
+  const uint32_t k1e9 = 1000000000U;
+  int num_segments = 0;
+  uint32_t segments[9];
+  memset(segments, 0, sizeof(segments));
+  uint64_t* most_significant_elem = words_little_endian + most_significant_elem_idx;
+
+  do {
```
pitrou, line 367: You should comment on what this is doing.

src/nanoarrow/utils.c
```diff
@@ -223,3 +223,186 @@ struct ArrowBufferAllocator ArrowBufferDeallocator(
...
+
+      *elem = (quotient_hi << 32) | quotient_lo;
+    } while (elem-- != words_little_endian);
+
+    segments[num_segments++] = remainder;
+  } while (*most_significant_elem != 0 || most_significant_elem-- != words_little_endian);
+
+  // We know our output has no more than 9 digits per segment, plus a negative sign,
+  // plus any further digits between our output of 9 digits and the maximum theoretical
+  // number of digits in an unsigned long, including the null terminator.
```
pitrou, line 388: I don't understand what the maximum theoretical number of digits has to do with this, can you explain?

src/nanoarrow/utils.c
```diff
@@ -223,3 +223,186 @@ struct ArrowBufferAllocator ArrowBufferDeallocator(
...
+                         (unsigned long)segments[num_segments - 1]);
+  buffer->size_bytes += n_chars;
+
+  // Subsequent output needs to be left-padded with zeroes such that each segment
+  // takes up exactly 9 digits.
+  for (int i = num_segments - 2; i >= 0; i--) {
+    int n_chars = snprintf((char*)buffer->data + buffer->size_bytes, 21, "%09lu",
+                           (unsigned long)segments[i]);
+    buffer->size_bytes += n_chars;
+  }
```
pitrou, line 405: Can you assert that `buffer->size_bytes` is not longer than the size allocated above?

src/nanoarrow/utils_test.cc
```diff
@@ -233,6 +233,29 @@ TEST(DecimalTest, Decimal128Test) {
   EXPECT_EQ(memcmp(decimal.words, bytes_neg, sizeof(bytes_neg)), 0);
 }
 
+TEST(DecimalTest, DecimalNegateTest) {
+  struct ArrowDecimal decimal;
+
+  for (auto bitwidth : {128, 256}) {
+    ArrowDecimalInit(&decimal, bitwidth, 10, 3);
+    ArrowDecimalSetInt(&decimal, 12345);
```
pitrou, line 241: Can you test with larger numbers?

src/nanoarrow/utils_test.cc
```diff
@@ -269,3 +292,154 @@ TEST(DecimalTest, Decimal256Test) {
...
+  EXPECT_EQ(decimal.words[decimal.low_word_index], std::numeric_limits<uint64_t>::max());
+  EXPECT_EQ(decimal.words[decimal.high_word_index], 0);
+
+  // Check roundtrip to string
+  buffer.size_bytes = 0;
+  ASSERT_EQ(ArrowDecimalAppendDigitsToBuffer(&decimal, &buffer), NANOARROW_OK);
+  EXPECT_EQ(std::string(reinterpret_cast<char*>(buffer.data), buffer.size_bytes),
+            "18446744073709551615");
+
+  // Check with the maximum value of a 128-bit integer
```
pitrou, line 345: Could you also have a similar test for 256 bits?

pitrou (changes requested):

src/nanoarrow/utils.c
paleolimbot: Done!

src/nanoarrow/utils.c
paleolimbot: The idea was to make sure that the `21` in `snprintf(, 21, "%lu")` is always accurate (I added this to the comment).

danepitkin:
Overall LGTM!

PR #365, opened 2024-01-15
Suppress unused variable warnings

WillAyd:
When using as a dependency and trying to use Wextra these generate warnings

### DIFF
--- a/src/nanoarrow/array_stream.c
+++ b/src/nanoarrow/array_stream.c
@@ -19,8 +19,6 @@
 
 #include "nanoarrow.h"
 
-#define NANOARROW_UNUSED(x) (void)(x)
-
 struct BasicArrayStreamPrivate {
   struct ArrowSchema schema;
   int64_t n_arrays;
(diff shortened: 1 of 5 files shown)


paleolimbot (commented):
Thanks!

Could you add a macro for this like `NANOARROW_UNUSED_PARAMETER()` in nanoarrow_types.h and use it in the functions you identified? I believe that is what Arrow C++ does in its headers for this kind of thing.

I think we should probably also remove `-Wno-unused-parameter` from our CMakeLists since we'll no longer need it (which would test this in CI).

src/nanoarrow/array_stream.c
```diff
@@ -19,6 +19,8 @@
 
 #include "nanoarrow.h"
 
+#define NANOARROW_UNUSED(x) (void)(x)
```
paleolimbot, line 22: Would you mind moving this to nanoarrow_types.h? Then it will be accessible to all of these files (and you can use it in array_inline.h). There are a few other macro definitions along these lines there as well 🙂

paleolimbot (approved):
Thank you! Just one comment about possibly putting the macro in the nanoarrow_types.h header.

PR #366, opened 2024-01-17
feat(python): Implement user-facing Schema class

paleolimbot:
General design:

- One `Schema` class. I could do many subclasses also, but the autocomplete is sort of nice and checking `schema.type` is maybe nicer than `isinstance(x, x)`. This means some properties return `None` for types where the property is not relevant (they could error, too, or be dynamically added).
- Very fast `Schema` construction from an existing `CSchema` (e.g., imported from somewhere)
- Somewhat slower performance if you work with these objects in Python (copies of children, looping over columns in Python, etc.)

Example:

```python
import nanoarrow as na

na.struct({"col1": na.int32()})
#> Schema(STRUCT, fields=[Schema(INT32, name='col1')])
na.fixed_size_binary(123)
#> Schema(FIXED_SIZE_BINARY, byte_width=123)
schema = na.fixed_size_binary(123)
schema.byte_width
#> 123
```

Things not implemented here for future PRs:

- A few types: list, map, dictionary, extension, union
- Metadata


danepitkin:
One `Schema` class feels like the right approach to me. It might be nice to split the creation of Schema objects into a factory class. This is what PyArrow does https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+    DURATION = CArrowType.DURATION
+    LARGE_STRING = CArrowType.LARGE_STRING
+    LARGE_BINARY = CArrowType.LARGE_BINARY
+    LARGE_LIST = CArrowType.LARGE_LIST
+    INTERVAL_MONTH_DAY_NANO = CArrowType.INTERVAL_MONTH_DAY_NANO
+
+    def __arrow_c_schema__(self):
+        # This will only work for parameter-free types
+        c_schema = CSchemaBuilder.allocate().set_type(self.value).finish()
+        return c_schema._capsule
```
jorisvandenbossche, line 73: Is this needed? (given it does not always work)
Or this ensures you can pass it directly to `c_schema`? (but that is not necessarily something that a user needs to do?)

python/src/nanoarrow/_lib.pyx
```diff
@@ -478,6 +593,130 @@ cdef class CSchemaView:
...
+            Error.raise_error("ArrowSchemaSetFormat()", result)
+
+        return self
+
+    def set_name(self, name):
+        self.c_schema._assert_valid()
+
+        cdef int result
+        if name is None:
+            result = ArrowSchemaSetName(self._ptr, NULL)
```
jorisvandenbossche, line 668: FWIW, it seems that Arrow C++ sets an empty string by default for schemas (types) that don't have a name, instead of NULL (now the spec allows both, so not that this has to change, I just noticed the difference when testing with passing pyarrow objects to na.Schema(..))

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+        >>> import pyarrow as pa
+        >>> schema = na.Schema(na.Type.INT32)
+        >>> schema = na.Schema(na.Type.DURATION, unit=na.TimeUnit.SECOND)
+        >>> schema = na.Schema(pa.int32())
+        """
+
+        if isinstance(obj, Type):
+            self._c_schema = _c_schema_from_type_and_params(obj, params, nullable)
+        elif not params:
+            self._c_schema = c_schema(obj)
```
jorisvandenbossche, line 113: The `nullable` keyword is being ignored here. Should we check `and nullable is None` in addition, so we raise an error when set and passing an object that already has that information encoded:

```
In [36]: na.Schema(pa.int16()).nullable
Out[36]: True

In [37]: na.Schema(pa.int16(), nullable=False).nullable
Out[37]: True

In [38]: na.Schema(pa.field("test", pa.int16(), nullable=False), nullable=True).nullable
Out[38]: False
```

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+
+
+def time64(unit, nullable=True) -> Schema:
+    """Create an instance of a 64-bit time of day type."""
+    return Schema(Type.TIME64, unit=unit, nullable=nullable)
+
+
+def timestamp(unit, timezone=None, nullable=True) -> Schema:
+    """Create an instance of a timestamp type."""
+    return Schema(Type.TIMESTAMP, timezone=timezone, unit=unit, nullable=nullable)
```
jorisvandenbossche, line 359: Idea for a follow-up PR, but we might want to do some mapping of "ms" -> TimeUnit.MILLI etc, so one can do `na.timestamp("ms")` ?

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+def _clean_fields(fields):
+    if isinstance(fields, dict):
+        return [(str(k), c_schema(v)) for k, v in fields.items()]
+    else:
+        fields_clean = []
+        for item in fields:
+            if isinstance(item, tuple) and len(item) == 2:
+                fields_clean.append((str(item[0]), c_schema(item[1])))
+            else:
+                fields_clean.append((None, c_schema(item)))
```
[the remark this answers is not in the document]
jorisvandenbossche: Seeing the tests, I suppose this does allow you to pass a list of Schema objects which already can have a name, and then this is fine. 

But then maybe we could do something like?

```
else:
    s = c_schema(item)
    if s.name is None:
        raise ValueError("schema for child field needs to have a name")
    fields_clean.append((None, s))
```

jorisvandenbossche (commented):
This is really nice! I added a lot of small comments, but they are really that, _small_ comments (I just did a deep dive into the code and experimented with it, so that resulted in a lot of comments ;)) 
I think the only architectural comment I have is that for what is now the `Schema.__init__`, we also could expose this constructor as a `na.schema(..)` function (like pyarrow does). But I don't have a real preference either way, the current code is also fine. 

Doesn't necessarily need to happen in this PR, but we need to add a `__repr__` to the Schema class

python/src/nanoarrow/_lib.pyx
paleolimbot: Good call! I ran into this in R too but forgot...setting the default to `""` causes way fewer problems for everybody.

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+# under the License.
+
+import enum
+from typing import Union
+
+from nanoarrow._lib import CArrowTimeUnit, CArrowType, CSchemaBuilder, CSchemaView
+from nanoarrow.c_lib import c_schema
+
+
+class Type(enum.Enum):
```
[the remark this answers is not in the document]
paleolimbot: I'm not sure that `Type` is the best name, but `ArrowType` does seem a little long. But also maybe that collides with a built-in?

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+class Schema:
+    """The Schema is nanoarrow's data type representation, encompasing the role
+    of PyArrow's ``Schema``, ``Field``, and ``DataType``. This scope maps to
+    that of the ArrowSchema in the Arrow C Data interface."""
+
+    def __init__(
+        self,
+        obj,
+        *,
+        nullable=None,
```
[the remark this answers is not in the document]
paleolimbot: It is, although I left `None` here because in the future it might be sort of useful to do `Schema(some_existing_schema, nullable=FALSE)` (i.e., modify an existing schema). (So `None` would be "don't change it").

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+        **params,
```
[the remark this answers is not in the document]
paleolimbot: Done! I'm not sure it's the best here, but you have to say `name=False` to actually set the name to NULL (probably only useful for testing).

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+    def child(self, i):
+        """Extract a child Schema
+
+        >>> import nanoarrow as na
+        >>> schema = na.struct({"col1": na.int32()})
+        >>> schema.child(0).type
+        <Type.INT32: 8>
+        """
+
+        # Returning a copy to reduce interdependence between Schema instances
```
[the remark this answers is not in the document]
paleolimbot: > Or because you are afraid that this then would keep too many objects alive when you inspect this?

That's the interdependence I was thinking of...for a root struct that has a million columns, it might be unexpected to keep all of them alive if all you want is the column type of the first column or something.

python/src/nanoarrow/schema.py
paleolimbot: Done!

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+def decimal256(precision: int, scale: int) -> Schema:
+    """Create an instance of a 256-bit decimal type."""
+    return Schema(Type.DECIMAL256, precision=precision, scale=scale)
+
+
+def struct(fields, nullable=True) -> Schema:
+    """Create a type representing a named sequence of fields."""
+    return Schema(Type.STRUCT, fields=fields, nullable=nullable)
+
+
```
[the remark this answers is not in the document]
paleolimbot: bool and null were a good call! I think I'll punt the others to follow-up PRs where more attention can be paid to the type-specific details 🙂

python/src/nanoarrow/schema.py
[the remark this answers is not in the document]
paleolimbot: I think I fixed this! Basically, it should be very hard to get that error. I think Arrow C++ exports an empty field name instead of a NULL always.

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+        col1
+        """
+        for i in range(self.n_children):
+            yield self.child(i)
+
+    def __arrow_c_schema__(self):
+        return self._c_schema.__arrow_c_schema__()
+
+
+def int8(nullable=True) -> Schema:
```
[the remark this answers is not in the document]
paleolimbot: Yes, the name is never set from these. Usually a constructor like `na.struct({"col_name": na.int32()})` would override the name anyway.

python/src/nanoarrow/schema.py
paleolimbot: Done!

python/src/nanoarrow/schema.py
paleolimbot: Perhaps unneeded, but instances of the enum values are perfectly valid representations of a type (for most types).

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+from nanoarrow.c_lib import c_schema
+
+
+class Type(enum.Enum):
+    """The Type enumerator provides a means by which the various type
+    categories can be identified. Type values can be used in place of
+    :class:`Schema` instances in most places for parameter-free types."""
+
+    UNINITIALIZED = CArrowType.UNINITIALIZED
+    NA = CArrowType.NA
```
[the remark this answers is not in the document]
paleolimbot: Done!

python/tests/test_schema.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def test_schema_time():
+    schema_obj = na.time32(na.TimeUnit.SECOND)
+    assert schema_obj.type == na.Type.TIME32
+    assert schema_obj.unit == na.TimeUnit.SECOND
+
+    schema_obj = na.time64(na.TimeUnit.SECOND)
+    assert schema_obj.type == na.Type.TIME32
+    assert schema_obj.unit == na.TimeUnit.SECOND
```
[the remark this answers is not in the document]
paleolimbot: I fixed this in the C library...apparently I never checked! The error is not great (`ArrowSchemaSetTypeDateTime() returned error code 22` or something) and we can make it better over time.

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+    return Schema(Type.DECIMAL128, precision=precision, scale=scale)
+
+
+def decimal256(precision: int, scale: int) -> Schema:
+    """Create an instance of a 256-bit decimal type."""
+    return Schema(Type.DECIMAL256, precision=precision, scale=scale)
+
+
+def struct(fields, nullable=True) -> Schema:
+    """Create a type representing a named sequence of fields."""
```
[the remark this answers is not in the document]
paleolimbot: Done! (Also for all the types while I was at it)

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+    return Schema(Type.STRING, nullable=nullable)
+
+
+def large_string(nullable=True) -> Schema:
+    """Create an instance of a variable-length UTF-8 encoded string type
+    that uses 64-bit offsets."""
+    return Schema(Type.LARGE_STRING, nullable=nullable)
+
+
+def binary(byte_width=None, nullable=True) -> Schema:
```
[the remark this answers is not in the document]
paleolimbot: Done! I like the `fixed_size_binary()` as a separate constructor too.

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+        self,
+        obj,
+        *,
+        nullable=None,
+        **params,
+    ) -> None:
+        """Create Schema objects
+
+        Create Schema objects from (1) a :class:`Type` or (1) any object supported by
+        :func:`c_schema`. When creating a class from a  a :class:`Type`, some
```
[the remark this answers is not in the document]
paleolimbot: I added a few bullet points!

python/src/nanoarrow/schema.py
```diff
@@ -0,0 +1,448 @@
...
+    that of the ArrowSchema in the Arrow C Data interface."""
+
+    def __init__(
+        self,
+        obj,
+        *,
+        nullable=None,
+        **params,
+    ) -> None:
+        """Create Schema objects
```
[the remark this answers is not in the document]
paleolimbot: I moved it to `nanoarrow.schema()`, since I'm envisioning that as the entry point for most people to construct one. I wish handling repetition in docstrings were a bit easier.

python/src/nanoarrow/__init__.py
```diff
@@ -26,4 +26,40 @@
...
+    binary,
+    large_binary,
+    fixed_size_binary,
+    date32,
+    date64,
+    time32,
+    time64,
+    timestamp,
+    duration,
+    interval_months,
```
WillAyd, line 58: Probably more of a pyarrow issue than something here but I don't think there is a pyarrow exposure for any interval aside from month/day/nano

python/src/nanoarrow/_lib.pyx
```diff
@@ -220,6 +220,64 @@ cdef class Error:
...
+    EXTENSION = NANOARROW_TYPE_EXTENSION
+    FIXED_SIZE_LIST = NANOARROW_TYPE_FIXED_SIZE_LIST
+    DURATION = NANOARROW_TYPE_DURATION
+    LARGE_STRING = NANOARROW_TYPE_LARGE_STRING
+    LARGE_BINARY = NANOARROW_TYPE_LARGE_BINARY
+    LARGE_LIST = NANOARROW_TYPE_LARGE_LIST
+    INTERVAL_MONTH_DAY_NANO = NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO
+
+
+cdef class CArrowTimeUnit:
```
WillAyd, line 270: I think this would fit more naturally if you just declared a `cpdef enum` The Cython docs has an example of this:

https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#structs-unions-enums

python/src/nanoarrow/_lib.pyx
```diff
@@ -376,6 +458,10 @@ cdef class CSchemaView:
     # lifetime guarantees that the pointed-to data from ArrowStringViews remains valid
     cdef object _base
     cdef ArrowSchemaView _schema_view
+    # Not part of the ArrowSchemaView (but possibly should be)
+    cdef int _dictionary_ordered
```
WillAyd, line 462: You could declare these as `bool` instead of `int`

python/src/nanoarrow/_lib.pyx
```diff
@@ -478,6 +593,130 @@ cdef class CSchemaView:
     def __repr__(self):
         return _lib_utils.schema_view_repr(self)
 
+
+cdef class CSchemaBuilder:
+    cdef CSchema c_schema
+    cdef ArrowSchema* _ptr
+
+    def __cinit__(self, CSchema schema):
```
WillAyd, line 601: The RAII equivalent of cinit is `__dealloc__` - you might want to implement that to potentially cleanup the schema in case ownership is never transfered when this object gets destroyed

python/src/nanoarrow/_lib.pyx
```diff
@@ -478,6 +593,130 @@ cdef class CSchemaView:
...
+    def set_type(self, int type):
+        self.c_schema._assert_valid()
+
+        cdef int result = ArrowSchemaSetType(self._ptr, <ArrowType>type)
+        if result != NANOARROW_OK:
+            Error.raise_error("ArrowSchemaSetType()", result)
+
+        return self
+
+    def set_type_decimal(self, int type, int precision, int scale):
```
WillAyd, line 623: I don't think it matters for cython but as a matter of style `type` is a reserved keyword in Python, so you typically don't use it as a variable name. I would suggest `type_` instead

python/src/nanoarrow/_lib.pyx
```diff
@@ -478,6 +593,130 @@ cdef class CSchemaView:
...
+        self.c_schema._assert_valid()
+
+        cdef int result = ArrowSchemaSetTypeDecimal(self._ptr, <ArrowType>type, precision, scale)
```
WillAyd, line 626: Do you actually need the casts here for the `type` argument?

python/src/nanoarrow/__init__.py
paleolimbot: Yeah I noticed that too! I don't know the history there but it's not too difficult to expose it here so 🤷

python/src/nanoarrow/_lib.pyx
paleolimbot: Apprently! I just re-tried and I get `src/nanoarrow/_lib.pyx:626:63: Cannot assign type 'int' to 'ArrowType'` 🤷

python/src/nanoarrow/_lib.pyx
paleolimbot: Good call! I updated these to `type_id`.

python/src/nanoarrow/_lib.pyx
paleolimbot: That's handled by the `CSchema` `_base` member (which is usually a PyCapsule that implements the logic you described). Here, the strong reference to the Python object is what I'm relying on to keep the underlying `ArrowSchema` alive.

python/src/nanoarrow/_lib.pyx
paleolimbot: Good call! I used `bint` since we don't link to `libcpp` but I think that has the same conversion properties!

python/src/nanoarrow/_lib.pyx
paleolimbot: That would be way better! It is a little awkward because I auto-generate the pxd file and when I tried to insert `cpdef` in the pxd I got an error (because I think they have to get defined in a pyx file). I will probably punt this to a future improvement (since the list of types is pretty stable).

paleolimbot: Or maybe I just have to add `nanoarrow_c.pyx`?

WillAyd: If you are using in another module you typically define it in a .pxd file which is cythons equivalent of a header

paleolimbot: Ah, so maybe what I need is actually `_lib.pyx`! I'll give it a try.

paleolimbot: I gave a few things a try but couldn't get it to work (which is more about my Cython familiarity than anything). For now I'll leave it as is...there are a number of things that could be improved about the Cython (including possibly eliminating it in favour of nanobind, or at least splitting up the giant _lib.pyx).

WillAyd: Makes sense. Yea Cython can be as confusing as C++ at times (ok maybe not quite...but close).

Nanoarrow is great. I use that on a smaller project I maintain called pantab. Would highly recommend

python/src/nanoarrow/schema.py
```diff
@@ -299,160 +280,627 @@ def __arrow_c_schema__(self):
...
+    """Create a nanoarrow Schema
+
+    The Schema is nanoarrow's high-level data type representation, encompasing
+    the role of PyArrow's ``Schema``, ``Field``, and ``DataType``. This scope
+    maps to that of the ArrowSchema in the Arrow C Data interface.
+
+    Parameters
+    ----------
+    obj :
+        A :class:`Type` specifier or an schema-like object. A schema-like object
```
jorisvandenbossche, line 293:
```suggestion
        A :class:`Type` specifier or a schema-like object. A schema-like object
```

python/src/nanoarrow/schema.py
```diff
@@ -299,160 +280,627 @@ def __arrow_c_schema__(self):
...
+        * A nanoarrow :class:`Schema`, :class:`CSchema`, or :class:`Type`
+        * Any object implementing the Arrow PyCapsule interface
+          ``__arrow_c_schema__()`` protocol method.
+
+    name : str, optional
+        An optional name to bind to this field.
+
+    nullable : bool, optional
+        Explicitly specify field nullability. Fields are nullable by default.
+        Not supported of ``obj`` is anything other than a :class:`Type` object.
```
jorisvandenbossche, line 305:
```suggestion
        Only supported if ``obj`` is a :class:`Type` object (for any other input, the nullability is preserved from the passed object).
```

python/src/nanoarrow/schema.py
```diff
@@ -299,160 +280,627 @@ def __arrow_c_schema__(self):
...
+
+    **params
+        Type-specific parameters when ``obj`` os a :class:`Type`.
```
jorisvandenbossche, line 308:
```suggestion
        Type-specific parameters when ``obj`` is a :class:`Type`.
```

python/src/nanoarrow/schema.py
```diff
@@ -214,18 +255,18 @@ def scale(self) -> int:
         return self._c_schema_view.decimal_scale
 
     @property
-    def n_children(self) -> int:
+    def n_fields(self) -> int:
```
jorisvandenbossche, line 237: Is there a specific reason you switched from "children" to "fields"? (the spec speaks about "children", I think?)

paleolimbot: I changed it to maintain the 1:1 mapping between the parameter names (e.g., `struct(fields=...)`) and property names. I believe that in pyarrow the term is "fields" as well ( https://arrow.apache.org/docs/python/generated/pyarrow.ListType.html#pyarrow.ListType.field ).

danepitkin:
Nice work! I think it would be great to get this into nanoarrow v0.4 so that users can try it out. I'm +1 for merging.

PR #367, opened 2024-01-19
feat: Support for Binaryview and StringView types

jorisvandenbossche:
Very draft PR, just putting here publicly what I experimented. 

This is currently a minimal addition that just allows to inspect an array / type:

```
In [1]: builder = pa.lib.StringViewBuilder()
   ...: builder.append("test")
   ...: builder.append("very long string that is not inlined")
   ...: builder.append(None)
   ...: builder.append("test")
   ...: arr = builder.finish()

In [2]: import nanoarrow as na

In [3]: na.c_schema(arr.type)
Out[3]: 
<nanoarrow.c_lib.CSchema string_view>
- format: 'vu'
- name: ''
- flags: 2
- metadata: NULL
- dictionary: NULL
- children[0]:

In [4]: na.c_schema_view(arr.type)
Out[4]: 
<nanoarrow.c_lib.CSchemaView>
- type: 'string_view'
- storage_type: 'string_view'

In [5]: na.c_array(arr)
Out[5]: 
<nanoarrow.c_lib.CArray string_view>
- length: 4
- offset: 0
- null_count: 1
- buffers: (140250311598080, 140250311598144, 140250311598208, 94079215855408)
- dictionary: NULL
- children[0]:

In [4]: na.c_array_view(arr)
Out[4]: 
<nanoarrow.c_lib.CArrayView>
- storage_type: 'string_view'
- length: 4
- offset: 0
- null_count: 1
- buffers[2]:
  - <bool validity[1 b] 11010000>
  - <string_view data[0 b] b''>
- dictionary: NULL
- children[0]:
```

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -695,6 +695,13 @@ static int ArrowArrayViewSetArrayInternal(struct ArrowArrayView* array_view,
     }
   }
 
+  if (array_view->storage_type == NANOARROW_TYPE_STRING_VIEW) {
+    array_view->n_varidic_buffers = array->n_buffers - 3;
+    array_view->variadic_buffer_sizes = array->buffers[array->n_buffers - 1];
+    // array_view->variadic_buffers = ...
+    buffers_required += array_view->n_varidic_buffers + 1;
+  }
+
   // Check the number of buffers
   if (buffers_required != array->n_buffers) {
     ArrowErrorSet(error, "Expected array with %d buffer(s) but found %d buffer(s)",
--- a/src/nanoarrow/nanoarrow_types.h
+++ b/src/nanoarrow/nanoarrow_types.h
@@ -443,7 +443,9 @@ enum ArrowType {
   NANOARROW_TYPE_LARGE_STRING,
   NANOARROW_TYPE_LARGE_BINARY,
   NANOARROW_TYPE_LARGE_LIST,
-  NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO
+  NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO,
+  NANOARROW_TYPE_BINARY_VIEW,
+  NANOARROW_TYPE_STRING_VIEW,
 };
 
 /// \brief Get a string value of an enum ArrowType value
@@ -482,10 +484,14 @@ static inline const char* ArrowTypeString(enum ArrowType type) {
       return "double";
     case NANOARROW_TYPE_STRING:
       return "string";
+    case NANOARROW_TYPE_STRING_VIEW:
+      return "string_view";
     case NANOARROW_TYPE_BINARY:
       return "binary";
     case NANOARROW_TYPE_FIXED_SIZE_BINARY:
       return "fixed_size_binary";
+    case NANOARROW_TYPE_BINARY_VIEW:
+      return "binary_view";
     case NANOARROW_TYPE_DATE32:
       return "date32";
     case NANOARROW_TYPE_DATE64:
@@ -784,6 +790,10 @@ struct ArrowArrayView {
   /// type_id == union_type_id_map[128 + child_index]. This value may be
   /// NULL in the case where child_id == type_id.
   int8_t* union_type_id_map;
+
+  int64_t n_varidic_buffers;
+  int64_t* variadic_buffer_sizes;
+  const void** variadic_buffers;
 };
 
 // Used as the private data member for ArrowArrays allocated here and accessed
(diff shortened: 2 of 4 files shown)


src/nanoarrow/nanoarrow_types.h
```diff
@@ -443,7 +443,9 @@ enum ArrowType {
   NANOARROW_TYPE_LARGE_STRING,
   NANOARROW_TYPE_LARGE_BINARY,
   NANOARROW_TYPE_LARGE_LIST,
-  NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO
+  NANOARROW_TYPE_INTERVAL_MONTH_DAY_NANO,
+  NANOARROW_TYPE_BINARY_VIEW,
+  NANOARROW_TYPE_STRING_VIEW,
```
jorisvandenbossche, line 448: Is this ArrowType enum meant to be stable? (i.e. can I only add new types at the end, or can I put them at a more logical place in the list?)

jorisvandenbossche:
@paleolimbot the reason that the Python binding for the ArrayView currently checks the buffer types dynamically to determine the number of buffers, is that so this works on an uninitialized ArrayView (just based on the layout, without an Array already being created)? See here:

https://github.com/apache/arrow-nanoarrow/blob/b3c952a3e21c2b47df85dbede3444f852614a3e2/python/src/nanoarrow/_lib.pyx#L685-L690

i.e. why not just return `self._ptr.array.n_buffers` here?

Because if we don't consider the variadic buffers as part of the ArrayView's buffers, then accessing like the above in Python doesn't work for those

src/nanoarrow/array.c
```diff
@@ -695,6 +695,13 @@ static int ArrowArrayViewSetArrayInternal(struct ArrowArrayView* array_view,
     }
   }
 
+  if (array_view->storage_type == NANOARROW_TYPE_STRING_VIEW) {
+    array_view->n_varidic_buffers = array->n_buffers - 3;
+    array_view->variadic_buffer_sizes = array->buffers[array->n_buffers - 1];
+    // array_view->variadic_buffers = ...
```
paleolimbot, line 701:
```suggestion
    array_view->variadic_buffers = array-> buffers + 2
```

...I think?

rouault:
Sorry for hickjacking this thread but I've become aware of the introductions of those new arrow data types because GDAL compilation broke against arrow 15.0 (due to a switch() not handling the new cases).  Will be addressed in https://github.com/OSGeo/gdal/pull/9116 in a minimalistic way by erroring out on those types. Do you know if those types will be actually found in serialized formats, namely Parquet and Feather ? And I hope that I won't have to add support for them in OGRLayer::WriteArrowBatch()... Their value proposition compared to regular string or binary is unclear to me. The only thing I see is that they might be a way to reduce memory usage by pointing to the same offset in case of duplicated strings? but wasn't that the purpose of dictionaries?  (actually the same question would hold for the RUN_END_ENCODED stuff added in libarrow 12).  As a relatively new comer to the Arrow ecosystem, I should point that the proliferation of basic data types is going to be a serious obstacle to adoption by new implementations.
Perhaps there should be some "data type negociation" mechanism where a consumer could tell to the producer something like "I just understand Int32, Float64, String. Do your best to present me only those data types by possibly morphing content to that"

jorisvandenbossche:
> Basically, the `ArrowArrayView` doesn't care how the memory it's pointing to got there (i.e., there's no requirement that an `ArrowArray` ever existed). This is mostly helpful for reading IPC: the decoder first fills in an ArrowArrayView and only allocates the `ArrowArray` if requested.

OK, understood! I see the C doc comment actually mentions this: *"use it to represent a hypothetical ArrowArray that does not exist yet, or use it to validate the buffers of a future ArrowArray."*.

eitsupi:
Hey, R polars package's Series/DataFrame to/from nanoarrow_array_stream conversion have been rewritten to only using the C Stream interface (pola-rs/r-polars#1078).
Even though nanoarrow 0.4.0 does not support StringView, It seems to work fine even when StringView is included.

JayjeetAtGithub:
Hi @jorisvandenbossche @paleolimbot We are trying to support interop between `arrow::StringViewArray` / `arrow::BinaryViewArray` and `cudf::column` in `libcudf`. But since, `libcudf` uses nanoarrow which does not have string view or binary view types yet, we are blocked. If nanoarrow could support these two types (along with the mappings into ArrowArray / ArrowDeviceArray / ArrowSchema) implementation for these types, we could implement the interop to `cudf::column`.

kylebarron:
> Their value proposition compared to regular string or binary is unclear to me

See also for more reading:

- https://cedardb.com/blog/german_strings/
- https://www.influxdata.com/blog/faster-queries-with-stringview-part-one-influxdb/
- https://youtu.be/et54i_Rplik?feature=shared&t=2966

PR #368, opened 2024-01-24
feat: Check flags field of ArrowSchema on ArrowSchemaViewInit

### LINKED ISSUE #312: Unknown flags should be zeroed when deep copying
The C data interface doesn't define deep copying of arrays, so a deep copy is logically an import followed by an export. New but backward compatible flags could be added which are unknown to nanoarrow on import. Since nanoarrow does not know how to uphold the guarantees implied by an unknown flag when exporting, those unknown flags should probably be zeroed in the copy instead of being passed through unmodified.

As a trivial example, the flag might indicate that the buffers are allocated in a particular memory space (like cuda or hip's `unified` memory). A deep copy's buffers would be allocated elsewhere, and passing through the unified memory flag could break a consumer.

paleolimbot:
Closes #312.

`ArrowSchemaViewInit()` is called every time a schema is about to be interpreted (or an array corresponding to that schema is about to be interpreted), and so checking the flags is a must to ensure some future unsupported flag that affects the interpretation of an array will cause an error.

The `ArrowSchemaView` should arguably also keep track of `flags` (or perhaps each flag individually), but I will leave that to another PR.

### DIFF
--- a/src/nanoarrow/schema.c
+++ b/src/nanoarrow/schema.c
@@ -1186,8 +1186,7 @@ ArrowErrorCode ArrowSchemaViewInit(struct ArrowSchemaView* schema_view,
         ArrowSchemaViewValidate(schema_view, schema_view->type, error));
   }
 
-  int64_t unknown_flags = schema->flags & ~ARROW_FLAG_DICTIONARY_ORDERED &
-                          ~ARROW_FLAG_NULLABLE & ~ARROW_FLAG_MAP_KEYS_SORTED;
+  int64_t unknown_flags = schema->flags & ~NANOARROW_FLAG_ALL_SUPPORTED;
   if (unknown_flags != 0) {
     ArrowErrorSet(error, "Unknown ArrowSchema flag");
     return EINVAL;
--- a/src/nanoarrow/schema.c
+++ b/src/nanoarrow/schema.c
@@ -1178,16 +1178,32 @@ ArrowErrorCode ArrowSchemaViewInit(struct ArrowSchemaView* schema_view,
     schema_view->type = NANOARROW_TYPE_DICTIONARY;
   }
 
-  result = ArrowSchemaViewValidate(schema_view, schema_view->storage_type, error);
-  if (result != NANOARROW_OK) {
-    return result;
-  }
+  NANOARROW_RETURN_NOT_OK(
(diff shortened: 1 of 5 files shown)


src/nanoarrow/schema.c
```diff
@@ -1178,16 +1178,32 @@ ArrowErrorCode ArrowSchemaViewInit(struct ArrowSchemaView* schema_view,
...
   if (schema_view->storage_type != schema_view->type) {
-    result = ArrowSchemaViewValidate(schema_view, schema_view->type, error);
-    if (result != NANOARROW_OK) {
-      return result;
-    }
+    NANOARROW_RETURN_NOT_OK(
+        ArrowSchemaViewValidate(schema_view, schema_view->type, error));
+  }
+
+  int64_t unknown_flags = schema->flags & ~ARROW_FLAG_DICTIONARY_ORDERED &
```
danepitkin, line 1189: Would this be better as a macro or #define that is defined next to these other #defines? From a maintenance perspective, I'm mainly considering what happens if someone adds a new flag and forgets to update this part of the code.

src/nanoarrow/schema_test.cc
```diff
@@ -574,6 +574,24 @@ TEST(SchemaViewTest, SchemaViewInitErrors) {
...
+               "ARROW_FLAG_DICTIONARY_ORDERED is only relevant for dictionaries");
+
+  schema.flags = 0;
+  schema.flags |= ARROW_FLAG_MAP_KEYS_SORTED;
+  EXPECT_EQ(ArrowSchemaViewInit(&schema_view, &schema, &error), EINVAL);
+  EXPECT_STREQ(ArrowErrorMessage(&error),
+               "ARROW_FLAG_MAP_KEYS_SORTED is only relevant for a map type");
+
+  schema.flags = 0;
+  schema.flags |= (ARROW_FLAG_MAP_KEYS_SORTED << 1);
```
danepitkin, line 591: We could use the new macro/define here to guarantee we always get an undefined value.

danepitkin:
Overall LGTM! Just a few small questions not related to correctness.

src/nanoarrow/schema.c
paleolimbot: Great idea!

danepitkin:
✅

PR #369, opened 2024-01-24
chore: Replace CHANGELOG generator

### LINKED ISSUE #310: Fix changelog generation
Depending on exactly what version of commitizen is installed, we get a number of errors with the built-in verification scripts. The `cz` call happens in two places as well (in the prepare and sign steps), plus there is the R changelog which could probably be synthesized from the normal changelog.

paleolimbot:
Closes #310.

Basically `cz ch` was constantly doing the wrong thing...for the last two releases I had to do quite a lot of manual editing after it was run. The Python script that replaces it:

- Can run more than once (replaces the target version's updates if commits were added)
- Groups components (e.g., so I can copy/paste the R NEWS more readily)
- Knows about when nanoarrow considers the last version "released"

### DIFF
--- a/dev/release/.gitignore
+++ b/dev/release/.gitignore
@@ -16,3 +16,4 @@
 # under the License.
 
 .env
+__pycache__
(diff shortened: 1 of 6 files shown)


lidavidm:
Hmm, ADBC might want to borrow this too.

assignUser (approved):
Nice, looks like a neat addition, thanks for adding the tests. At the moment the tests are not run in CI right? (or are they part of pre-commit?)

PR #378, opened 2024-01-31
feat(python): Add array creation/building from buffers

paleolimbot:
The gist of this PR is that I'd like the ability to create arrays for testing without pyarrow so that nanoarrow's tests can run in more places. Other than building/running in odd corner-case environments, nanoarrow in R has been great at prototyping and/or creating test data (e.g., an array with a non-zero offset, an array with a rarely-used type). This is useful for both nanoarrow to test itself and perhaps others who might want to use nanoarrow in a similar way in Python.

This is a bit big...I did need to put all of it in one place to figure out what the end point was; however, I'm happy to split into smaller self-contained bits now that I know where I'm headed.

After this PR, we can create an array out-of-the-box from anything that supports the buffer protocol. Importantly, this includes numpy arrays so that you can do things like generate arrays with `n` random numbers.

```python
import nanoarrow as na
import numpy as np
```

```python
na.c_array_view(b"12345")
```

    <nanoarrow.c_lib.CArrayView>
    - storage_type: 'uint8'
    - length: 5
    - offset: 0
    - null_count: 0
    - buffers[2]:
      - validity <bool[0 b] >
      - data <uint8[5 b] 49 50 51 52 53>
    - dictionary: NULL
    - children[0]:

```python
na.c_array_view(np.array([1, 2, 3], np.int32))
```

```
<nanoarrow.c_lib.CArrayView>
- storage_type: 'int32'
- length: 3
- offset: 0
- null_count: 0
- buffers[2]:
  - validity <bool[0 b] >
  - data <int32[12 b] 1 2 3>
- dictionary: NULL
- children[0]:
```

While not built in to the main `c_array()` constructor, we can also now assemble an array from buffers. This has been very useful in R and ensures that we can construct just about any array if we need to.

```python
array = na.c_array_from_buffers(
    na.struct([na.int32()]),
    length=3,
    buffers=[None],
    children=[
        na.c_array_from_buffers(
            na.int32(),
            length=3,
            buffers=[None, na.c_buffer([1, 2, 3], na.int32())]
        )
    ],
)

na.c_array_view(array)
```

    <nanoarrow.c_lib.CArrayView>
    - storage_type: 'struct'
    - length: 3
    - offset: 0
    - null_count: 0
    - buffers[1]:
      - validity <bool[0 b] >
    - dictionary: NULL
    - children[1]:
      - <nanoarrow.c_lib.CArrayView>
        - storage_type: 'int32'
        - length: 3
        - offset: 0
        - null_count: 0
        - buffers[2]:
          - validity <bool[0 b] >
          - data <int32[12 b] 1 2 3>
        - dictionary: NULL
        - children[0]:

I also added the ability to construct a buffer from an iterable and wired that into the `c_array()` constructor although this is probably not all that fast. It does, however, make it much easier to write tests (because many of them currently start with `na_c_array(pa.array([1, 2, 3]))`.

```python
na.c_array_view([1, 2, 3], na.int32())
```

    <nanoarrow.c_lib.CArrayView>
    - storage_type: 'int32'
    - length: 3
    - offset: 0
    - null_count: 0
    - buffers[2]:
      - validity <bool[0 b] >
      - data <int32[12 b] 1 2 3>
    - dictionary: NULL
    - children[0]:

This allows creating an array from anything supported by the `struct` module which means we can create some of the less frequently used types.

```python
na.c_array_view([1, 2, 3], na.float16())
```

    CBuffer(half_float[6 b] 1.0 2.0 3.0)

```python
na.c_array_view([(1, 2), (3, 4), (5, 6)], na.interval_day_time())
```

    CBuffer(interval_day_time[24 b] (1, 2) (3, 4) (5, 6))

Because it's mentaly exhausting to bitpack buffers in my head and because Arrow uses them all the time, I also think it's mission-critical to be able to create bitmaps:

```python
na.c_buffer([True, False, True, True], na.bool())
```

    CBuffer(bool[1 b] 10110000)

This involved fixing some issues with the existing buffer view:

- The buffer view only ever saved a pointer to the device. This is a bit of a problem because even though the CPU device is static and lives forever, CUDA "device" objects will probably keep a CUDA context alive. Thus, we need a strong reference to the `CDevice` Python object (which ensures the underlying nanoarrow `Device*` remains valid).
- The buffer view only handled `BufferView` input where technically all it needs is a pointer and a length. This opens it up to represent other types of buffers than just something from nanoarrow (e.g., imported from dlpack or buffer protocol).

Implementing the buffer protocol as a consumer was done by wrapping the `ArrowBuffer` with a "deallocator" that holds the `Py_buffer` and ensures it is released. I still need to do some testing to ensure that it's actually released and that we're not leaking memory. This is how I do it in R and in geoarrow-c (Python) as well. Using the `ArrowBuffer` is helpful because the C-level array builder uses them to manage the memory and ensures they're all released when the array is released.

Implementing the build-from-iterable involved a few more things...notably, completing the "python struct format string" <-> "arrow data type" conversion. This allows the use of `struct.pack()` which takes care of things like half-float conversion and tuples of day, month, nano conversion.

I'm aware this could use a bit better documentation of the added classes/methods...I am assuming these will be internal for the time being but they definitely need a bit more than is currently there.


python/src/nanoarrow/_lib.pyx
```diff
@@ -1024,23 +1076,24 @@ cdef class CBufferView:
...
-    cdef ArrowDevice* _device
+    cdef CDevice _device
     cdef Py_ssize_t _element_size_bits
     cdef Py_ssize_t _shape
     cdef Py_ssize_t _strides
     cdef char _format[128]
 
-    def __cinit__(self, object base, uintptr_t addr,
+    def __cinit__(self, object base, uintptr_t addr, int64_t size_bytes,
                   ArrowBufferType buffer_type, ArrowType buffer_data_type,
```
jorisvandenbossche, line 1089: Not related to this PR, just noticing while reviewing: should the buffer type be part of the buffer? That doesn't necessarily make sense for a buffer itself, only in context of a buffer as part of an array. But if you create a standalone buffer, then it doesn't make sense?

jorisvandenbossche: It also seems we currently don't use this attribute, except for the public `type` property

jorisvandenbossche: (I know the same is true for the data type, but at least that one you need to support exporting the buffer to an array / the buffer protocol)

jorisvandenbossche (commented):
This first set of changes look good!

python/README.md
```diff
@@ -285,3 +264,9 @@ pip install -e .[test]
 # Run tests
 pytest -vvx
 ```
+
+If you need to debug or edit any .c files used to build the native extension, you can generate the `compile_commands.json` usable by most IDEs/clang tools using [Bear](https://github.com/rizsotto/Bear).
+
+```shell
+bear -- python setup.py build_ext --inplace && mv compile_commands.json build
+```
```
jorisvandenbossche, line 272: BTW, this is a bit cryptic to me what this does / can help you with, but I am also not familiar with the tool or the compile_commands

paleolimbot: I'll leave it out for now since it's probably just me who has to remember that command. If `build/compile_commands.json` is present, most editors "just work" with the .c/.cc./.h files without additional configuration.

python/src/nanoarrow/c_lib.py
```diff
@@ -125,10 +138,205 @@ def c_array(obj=None, requested_schema=None) -> CArray:
...
+    # Try buffer protocol (e.g., numpy arrays)
+    try:
+        return c_array_from_pybuffer(obj)
+    except Exception as e:
         raise TypeError(
             f"Can't convert object of type {type(obj).__name__} to nanoarrow.c_array"
+        ) from e
+
+
+def c_array_from_pybuffer(obj) -> CArray:
```
danepitkin, line 151: (Optional) It might be nice to put these into a class e.g.
```
class CArray:
    @staticmethod
    def from_pybuffers():
         ...

    @staticmethod
    def from_buffers():
         ...

File: python/src/nanoarrow/_lib.pyx [hunk 4]
```diff
@@ -176,6 +178,188 @@ cdef object alloc_c_array_shallow_copy(object base, const ArrowArray* c_array) n
...
+    except Exception as e:
+        PyBuffer_Release(&buffer)
+        raise e
+
+    # Transfers ownership of buffer to c_buffer, whose finalizer will be called by
+    # the capsule when the capsule is deleted or garbage collected
+    c_buffer[0].data = <uint8_t*>buffer.buf
+    c_buffer[0].size_bytes = <int64_t>buffer.len
+    c_buffer[0].capacity_bytes = 0
+    c_buffer[0].allocator = c_pybuffer_deallocator(&buffer)
```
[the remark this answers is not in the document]
paleolimbot: `PyObject_GetBuffer()` is sufficient (i.e., if you forget to `PyBuffer_Release()`, you will leak). I don't think there is a way to increment the reference (it's not a `PyObject`) and the buffer protocol documentation seemed to indicate that one should not increment or decrement the reference count of `buffer.obj`.

python/src/nanoarrow/c_lib.py
```diff
@@ -125,10 +138,205 @@ def c_array(obj=None, requested_schema=None) -> CArray:
...
+    builder.set_buffer(1, buffer)
+
+    # No nulls or offset from a PyBuffer
+    builder.set_null_count(0)
+    builder.set_offset(0)
+
+    return builder.finish()
+
+
+def c_array_empty(schema) -> CArray:
```
danepitkin, line 210: As a user, I typically would assume calling `c_array(schema)` could give me an empty array, too. Would it be more intuitive to consolidate APIs?

python/src/nanoarrow/_lib.pyx
```diff
@@ -176,6 +178,188 @@ cdef object alloc_c_array_shallow_copy(object base, const ArrowArray* c_array) n
...
+    if sys_byteorder == "little" and (">" in format or "!" in format):
+        raise ValueError(f"Can't convert format '{format}' to Arrow type")
+    elif sys_byteorder == "big" and  "<" in format:
+        raise ValueError(f"Can't convert format '{format}' to Arrow type")
+
+    # Strip system endian specifiers
+    format = format.strip("=@")
+
+    if format == "c":
+        return 0, NANOARROW_TYPE_STRING
```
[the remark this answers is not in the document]
paleolimbot: I stole this from how numpy deals with `np.array("abcdefg")`. This particular mapping allows `np.array("abc")` to roundtrip through a `c_buffer()` but I don't really have any strong feelings about it.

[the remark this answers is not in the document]
paleolimbot: Whoops, it's a little different...more like if you call `c_buffer(np.array(c_buffer(<that's a string>)))` you can get the roundtrip. `np.array(list("abcdefg"))` seems to do something slightly different.

```
import numpy as np
from nanoarrow.c_lib import c_buffer
np.array(c_buffer(b"abcdefg").set_format("c").data)
#> array([b'a', b'b', b'c', b'd', b'e', b'f', b'g'], dtype='|S1')
```

python/src/nanoarrow/c_lib.py
```diff
@@ -257,7 +465,74 @@ def c_array_view(obj, requested_schema=None) -> CArrayView:
     return CArrayView.from_cpu_array(c_array(obj, requested_schema))
 
 
-def allocate_c_schema():
+def c_buffer(obj) -> CBuffer:
+    """Owning, read-only ArrowBuffer wrapper
+
+    Wraps obj in nanoarrow's owning buffer structure, the ArrowBuffer,
```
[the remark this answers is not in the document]
paleolimbot: Yes! I'll add a note.

python/src/nanoarrow/c_lib.py
```diff
@@ -125,10 +138,205 @@ def c_array(obj=None, requested_schema=None) -> CArray:
...
+    - null_count: 0
+    - buffers[2]:
+      - validity <bool[0 b] >
+      - data <uint8[4 b] 49 50 51 52>
+    - dictionary: NULL
+    - children[0]:
+    """
+
+    buffer = CBuffer().set_pybuffer(obj)
+    view = buffer.data
```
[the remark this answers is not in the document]
paleolimbot: Agreed! `data` was the name of the `ArrowBuffer` C struct field but we've departed pretty significantly from it here, so definitely worth renaming (or maybe better: just implement the buffer protocol on the buffer as you suggested below!)

python/tests/test_c_lib.py
```diff
@@ -0,0 +1,489 @@
...
+    assert repr(invalid) == "CBuffer(<invalid>)"
+
+
+def test_c_buffer_constructor():
+    invalid = CBuffer()
+    assert c_buffer(invalid) is invalid
+
+    buffer = c_buffer(b"1234")
+    assert isinstance(buffer, CBuffer)
+    assert bytes(buffer.data) == b"1234"
```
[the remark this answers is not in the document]
paleolimbot: Totally! It could even be writable too...

python/src/nanoarrow/c_lib.py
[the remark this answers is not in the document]
danepitkin: Could this maybe be handled in `CBuffer.__init__()` instead of using a separate API?

danepitkin (commented):
Great work!

python/src/nanoarrow/c_lib.py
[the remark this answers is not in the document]
paleolimbot: Ok! It was a bit of a goose chase, but implementing the buffer protocol on the `CBuffer` (by caching a `CBufferView` on it) removed the need for `.data` alltogether and made the code much better! In theory also safer, since it's possible to check for "open" buffers before possibly modifying its contents.

python/src/nanoarrow/c_lib.py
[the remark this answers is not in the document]
paleolimbot: It would definitely be more Pythonic! Feel free to push back here, but the constructors for all the objects are like this for two reasons.

First, the docstring is inaccessible to IDEs if they live in Cython (or maybe: it would require learning a bunch of things I don't know how to do to make them accessible). Right now if you open VSCode with a script containing `import nanoarrow as na`, `na.c_buffer(` will display a pop-up with the docstring, whereas `na.c_lib.CArray(` will not.

The second reason is that iterating in Cython is slow: every time you make a change, you have to wait about ten seconds to rebuild and fix any syntax errors. With an editable install, changes in Python are instantaneous.

Both of these things mean that I've actively tried to avoid Cython, which results in a slightly non-Pythonic interface for classes that have to live there (because they manage memory with C calls or because they implement the buffer protocol, neither of which are easy or possible in Python).

python/tests/test_c_lib.py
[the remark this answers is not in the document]
paleolimbot: Done! I started to go down the "writable" rabbit hole but it requires some more "open buffer" counting. It might enable `pack_into()`, which might be faster than `pack()` + `write()` (but custom buffer/array builders are almost certainly faster than both).

python/src/nanoarrow/c_lib.py
paleolimbot: That would definitely be more Pythonic! I think when there is an `Array` class that lives in Python, `Array.from_buffers()` is absolutely the way to go.

This put me down a bit of a rabbit hole trying to find something that was better/raised fewer eyebrows, and I settled on solidifying `c_array()` as the one stop shop for "get me Arrow data". Like `pyarrow.array()` and `numpy.array()`, they accept existing arrays, things with the right protocol methods, and iterables. I had sort of done this already but solidifying and documenting that made it clear that `c_array_from_pybuffer()` should be internal (it's now `_c_array_from_pybuffer()`). I don't *think* there's un-workaround-able overlap...one could do `c_array(memoryview(obj))` to force buffer protocol, or `c_array(list(obj))` to force build-by-iterable.

I also did that for `c_buffer()`, which similarly made it clear (to me, maybe) that `c_buffer_from_iterable()` should be internal, and that `c_buffer()` should be exported.

`c_array_from_buffers()` doesn't quite fit the `c_array()` idea (I have something array like, get it to me in arrow format), so I kept that on its own. There's room for it to move somewhere else if it turns out to not be the right fit!

python/src/nanoarrow/c_lib.py
paleolimbot: I special-cased empty iterables so that you can do `c_array([], schema)`! I'm a tiny bit worried that somebody would have something like:

```python
def some_method(array, schema):
    array = c_array(array)
    schema = c_schema(schema)
```

...and if somebody mixes up the argument order, they might get `array` as an empty array and `schema` as the Array's schema.

python/src/nanoarrow/_lib.pyx
```diff
@@ -1177,11 +1429,395 @@ cdef class CBufferView:
...
+        return self._data_type
+
+    @property
+    def element_size_bits(self):
+        return self._element_size_bits
+
+    @property
+    def item_size(self):
+        self._refresh_view_if_needed()
+        return self._view.item_size
```
jorisvandenbossche, line 1572: The difference between both is bits vs bytes? (if so, we might want to use a more consistent naming scheme? item vs element)

python/src/nanoarrow/c_lib.py
```diff
@@ -75,35 +93,53 @@ def c_schema(obj=None) -> CSchema:
...
 
     Note that the :class:`CArray` objects returned by ``.child()`` hold strong
-    references to the original ``ArrowSchema`` to avoid copies while inspecting an
+    references to the original ``ArrowArray`` to avoid copies while inspecting an
     imported structure.
 
+    Parameters
+    ----------
+    obj : array-like
+        An object supporting the Arrow PyCapsule interface, the Python
```
jorisvandenbossche, line 118:
```suggestion
        An object supporting the Arrow PyCapsule interface, the Python buffer
```

python/src/nanoarrow/c_lib.py
```diff
@@ -120,15 +157,134 @@ def c_array(obj=None, requested_schema=None) -> CArray:
...
-    # for pyarrow < 14.0
-    if hasattr(obj, "_export_to_c"):
+    # Try buffer protocol (e.g., numpy arrays or a c_buffer())
+    if _obj_is_buffer(obj):
+        return _c_array_from_pybuffer(obj)
+
+    # Try import of bare capsule
+    if _obj_is_capsule(obj, "arrow_array"):
+        if requested_schema is None:
+            requested_schema_capsule = CSchema.allocate()._capsule
```
jorisvandenbossche, line 167: How does the import work if you don't have an actual schema? (I suppose the above allocates an "emtpy" ArrowSchema?)

python/src/nanoarrow/c_lib.py
```diff
@@ -120,15 +157,134 @@ def c_array(obj=None, requested_schema=None) -> CArray:
...
+
+    Given a schema, build an ArrowArray buffer-wise. This allows almost any array
+    to be assembled; however, requires some knowledge of the Arrow Columnar
+    specification. This function will do its best to validate the sizes and
+    content of buffers according to ``validation_level``, which can be set
+    to ``"full""`` for maximum safety.
+
+    Parameters
+    ----------
+
```
jorisvandenbossche, line 207:
```suggestion
```

python/src/nanoarrow/c_lib.py
```diff
@@ -120,15 +157,134 @@ def c_array(obj=None, requested_schema=None) -> CArray:
...
+    - buffers[2]:
+      - validity <bool[0 b] >
+      - data <uint8[5 b] 49 50 51 52 53>
+    - dictionary: NULL
+    - children[0]:
+    """
+    schema = c_schema(schema)
+    builder = CArrayBuilder.allocate()
+
+    # This is slightly wasteful: it will allocate arrays recursively and we are about
```
jorisvandenbossche, line 254: This is about the _child_ arrays? (because the parent array itself isn't released, for that we just set the buffers?)

python/src/nanoarrow/c_lib.py
```diff
@@ -257,7 +413,60 @@ def c_array_view(obj, requested_schema=None) -> CArrayView:
     return CArrayView.from_cpu_array(c_array(obj, requested_schema))
 
 
-def allocate_c_schema():
+def c_buffer(obj, requested_schema=None) -> CBuffer:
+    """Owning, read-only ArrowBuffer wrapper
+
+    If obj implement the Python buffer protocol, ``c_buffer()`` Wraps
```
jorisvandenbossche, line 419:
```suggestion
    If obj implement the Python buffer protocol, ``c_buffer()`` wraps
```

python/src/nanoarrow/_lib.pyx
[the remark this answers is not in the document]
jorisvandenbossche: Ah, I see, you need to call `PyBuffer_Release`, and I assume python will ensure it increases the ref on the object / keeps that alive, until that release is called (so essentially the same as with the C Data Interface release callback)

python/src/nanoarrow/_lib.pyx
```diff
@@ -1177,11 +1429,395 @@ cdef class CBufferView:
...
+            self._base = alloc_c_buffer(&self._ptr)
+        ArrowBufferReset(self._ptr)
+
+        self._data_type = NANOARROW_TYPE_BINARY
+        self._element_size_bits = 0
+        self._device = CDEVICE_CPU
+        self._reset_view()
+        return self
+
+    def set_pybuffer(self, obj):
```
jorisvandenbossche, line 1512: Instead of having this "mutable" interface, essentially this function just creates a CBuffer from a python buffer, right? So it could also be a class method, and instead of using ``CBuffer().set_pybuffer(obj)`` one would do `CBuffer.from_pybuffer(obj)` ?

jorisvandenbossche: What I think I find confusing about those `set_` methods, is that it makes CBuffer look like a builder class, while there _is_ a separate BufferBuilder class?

python/tests/test_c_array.py
```diff
@@ -0,0 +1,298 @@
...
+    assert c_array.offset == 0
+    assert na.c_schema_view(c_array.schema).type == "uint8"
+
+    c_array_view = na.c_array_view(c_array)
+    assert list(c_array_view.buffer(1)) == list(data)
+
+
+def test_c_array_from_pybuffer_string():
+    data = b"abcdefg"
+    buffer = na.c_buffer(data).set_format("c")
```
jorisvandenbossche, line 120: This could also look like `na.c_buffer(data, format="c")` ? 
(related to my comment above about those builder-like set methods on CBuffer itself)

python/src/nanoarrow/c_lib.py
```diff
@@ -120,15 +157,134 @@ def c_array(obj=None, requested_schema=None) -> CArray:
...
+
+    # This is slightly wasteful: it will allocate arrays recursively and we are about
+    # to immediately release them and replace them with another value. We could also
+    # create an ArrowArrayView from the buffers, which would make it more
+    # straightforward to check the buffer types and avoid the extra structure
+    # allocation.
+    builder.init_from_schema(schema)
+
+    # Set buffers. This moves ownership of the buffers as well (i.e., the objects
+    # in the input buffers are replaced with an empty ArrowBuffer)
```
jorisvandenbossche, line 262: This was a bit confusing (I ran into this while trying out the method):

```
In [74]: buf = na.c_buffer(b"0123")

In [75]: na.c_array_from_buffers(na.int32(), 1, [None, buf])
Out[75]: 
<nanoarrow.c_lib.CArray int32>
- length: 1
- offset: 0
- null_count: 0
- buffers: (0, 139688822562528)
- dictionary: NULL
- children[0]:

In [76]: na.c_array_from_buffers(na.int32(), 1, [None, buf])
---------------------------------------------------------------------------
NanoarrowException                        Traceback (most recent call last)
Cell In [76], line 1
----> 1 na.c_array_from_buffers(na.int32(), 1, [None, buf])

File ~/scipy/repos/arrow-nanoarrow/python/src/nanoarrow/c_lib.py:287, in c_array_from_buffers(schema, length, buffers, null_count, offset, children, validation_level)
    284 builder.resolve_null_count()
    286 # Validate + finish
--> 287 return builder.finish(validation_level=validation_level)

File src/nanoarrow/_lib.pyx:1818, in nanoarrow._lib.CArrayBuilder.finish()

File src/nanoarrow/_lib.pyx:422, in nanoarrow._lib.Error.raise_message_not_ok()

File src/nanoarrow/_lib.pyx:417, in nanoarrow._lib.Error.raise_message()

NanoarrowException: ArrowArrayFinishBuildingDefault() failed (22): Expected int32 array buffer 1 to have size >= 4 bytes but found buffer with 0 bytes

In [77]: buf
Out[77]: CBuffer(uint8[0 b] )
```

When passing python object through buffer protocol, the original object keeps owning the buffer content. We can't do something similar for CBuffer? (and that's because the CArray doesn't actually keep track of CBuffer objects, but only the C ArrowArray that has a pointer to the buffer?)

And where does this actually happen? I don't see anything in `CArrayBuilder.set_buffer` that would invalidate the buffer?

python/src/nanoarrow/_lib.pyx
```diff
@@ -1177,11 +1429,395 @@ cdef class CBufferView:
...
+    def set_format(self, str format):
+        self._assert_buffer_count_zero()
+        element_size_bytes, data_type = c_arrow_type_from_format(format)
+        self._data_type = data_type
+        self._element_size_bits = element_size_bytes * 8
+        format_bytes = format.encode("UTF-8")
+        snprintf(self._format, sizeof(self._format), "%s", <const char*>format_bytes)
+        return self
+
+    def set_data_type(self, ArrowType type_id, int element_size_bits=0):
```
jorisvandenbossche, line 1531: Related to the comment just above, does this method need to live on CBuffer? As in practice, it only seems to be used for CBufferBuilder?

python/src/nanoarrow/c_lib.py
```diff
@@ -257,7 +413,60 @@ def c_array_view(obj, requested_schema=None) -> CArrayView:
     return CArrayView.from_cpu_array(c_array(obj, requested_schema))
 
 
-def allocate_c_schema():
+def c_buffer(obj, requested_schema=None) -> CBuffer:
```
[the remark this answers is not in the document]
jorisvandenbossche: Ah, I see we used `requested_schema` in `c_array(..)` before as well. Personally, for places where the _user_ would provide a schema, I would go for `schema` instead of `requested_schema` (like `c_array_from_buffers` uses)

python/src/nanoarrow/_lib.pyx
jorisvandenbossche: And the same for `set_empty`, which in practice is also used for CBufferBuilder (except for a test where it's used to create an emtpy CBuffer, but those tests are explicitly testing the `set_empty` method)

python/src/nanoarrow/_lib.pyx
```diff
@@ -1177,11 +1429,395 @@ cdef class CBufferView:
...
+                buffer_item_i = 0
+
+        if buffer_item_i != 0:
+            code = ArrowBufferAppendInt8(self._ptr, buffer_item)
+            Error.raise_error_not_ok("ArrowBufferAppendInt8()", code)
+
+        return n_values
+
+    def finish(self):
+        return self
```
jorisvandenbossche, line 1697: This means that it returns a BufferBuilder, which has some additional methods to further append to the Buffer?

jorisvandenbossche (commented):
Another batch of comments ;) 

Most are minor details, but you will notice that the bulk of the comments has to do with CBuffer vs CBufferBuilder, and CBuffer being a "mutable" class with those set_.. methods. My initial reaction was that it might be cleaner to separate those two more clearly (eg CBuffer doesn't need those set_ methods, and the builder can maybe not subclass CBuffer / CBufferBuilder.finish can return a CBuffer instead of itself). 

(not having those set_.. methods on CBuffer might also avoid the need for the `_assert_buffer_count_zero` / `_refresh_view_if_needed` checks?)

Although while writing this now, I also realize that in Arrow C ++ we have the concept of a Buffer vs Mutable/ResizableBuffer as well.

python/src/nanoarrow/c_lib.py
jorisvandenbossche: > And where does this actually happen? I don't see anything in `CArrayBuilder.set_buffer` that would invalidate the buffer?

Ah, I suppose it is the `ArrowBufferMove` in the C code that does this?

python/src/nanoarrow/_lib.pyx
```diff
@@ -790,8 +1043,6 @@ cdef class CArray:
             respectively.
         """
         self._assert_valid()
-        if requested_schema is not None:
-            raise NotImplementedError("requested_schema")
```
jorisvandenbossche, line 794: Is there a reason this was removed? (because it's still not supported I think, the keyword is just ignored otherwise)

paleolimbot: I'll add it back...I had thought requesting a schema was best-effort and not cast-or-error but I just re-read the final language.

If you're a consumer passing a `requested_schema` through the protocol, it would be a very bad idea not to check that the schema you received was the one you requested (because if the producer got it wrong, you will crash). For producers like this one that don't support casting, it forces that check to happen twice (if you want to support the case where `schema` is already the correct type). If you can do casting but only support some of the conversions (e.g., you can return stringviews but not listviews), the current language of the spec forces you to error.

(I know, a discussion for another place).

python/src/nanoarrow/_lib.pyx
paleolimbot: Good call! I fixed this...it now returns a completely independent `CBuffer`.

python/src/nanoarrow/_lib.pyx
paleolimbot: I removed the set methods (or made them internal so that I can use them in tests)!

python/src/nanoarrow/c_lib.py
paleolimbot: I made the `ArrowBufferMove()` explicit (and made the "move + invalidate previous") behaviour opt-in, since it's definitely confusing if you did not expect it to happen).

python/tests/test_c_array.py
paleolimbot: This is pretty much just for testing, so I made it internal. I don't think it should be easy to blindly reinterpret cast a buffer but there are a few code paths that are hard to test if I can't do it in a test.

python/src/nanoarrow/_lib.pyx
paleolimbot: I removed them! The `CBuffer` should now be immutable (except in a few limited internal cases).

python/src/nanoarrow/c_lib.py
```diff
@@ -305,3 +514,101 @@ def allocate_c_array_stream():
...
+        return False
+
+    if not obj_type.__name__.endswith("Array") and obj_type.__name__ != "RecordBatch":
+        return False
+
+    return hasattr(obj, "_export_to_c")
+
+
+def _obj_is_iterable(obj):
+    return hasattr(obj, "__iter__")
```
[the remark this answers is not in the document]
paleolimbot: I think there are some valid iterators that don't set `__len__` that we probably want to support (although we might be able to do a better job reserving if we do check for a len attribute).

```python
hasattr(zip(b"123", "abc"), "__len__")
#> False
hasattr((x for x in b"123"), "__len__")
#> False
```

python/src/nanoarrow/c_lib.py
```diff
@@ -257,7 +413,60 @@ def c_array_view(obj, requested_schema=None) -> CArrayView:
...
+    (i.e., ``PyObject_GetBuffer()`` and ``PyBuffer_Release()``).
+
+    If obj is iterable, a buffer will be allocated and populated with
+    the contents of obj according to ``requested_schema``. The
+    ``requested_schema`` parameter is required to create a buffer from
+    a Python iterable. The ``struct`` module is currently used to encode
+    values from obj into binary form.
+
+    Unlike with :func:`c_array`, ``requested_schema`` is explicitly
+    honoured (or an error will be raised).
```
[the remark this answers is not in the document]
paleolimbot: I changed this to error if `schema` is passed with a buffer for now, which would force being explicit for now (e.g., c_buffer(iter(some_array), na.int32())`.

python/src/nanoarrow/c_lib.py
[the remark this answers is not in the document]
paleolimbot: I updated all user-facing `requested_schema` references to be `schema`!

python/src/nanoarrow/_lib.pyx
paleolimbot: I updated the `CBufferView`, the `CBuffer`, and the `CBufferBuilder` to all use `item`(s) for the definition of item that is used by the buffer protocol/memoryview, and `element` to support the definition we use, which is basically the same except supports bitmaps. I also added in a way to extract bitmaps since we'll need them shortly.

python/src/nanoarrow/c_lib.py
paleolimbot: Yes, it will leave you with an invalid schema. Nothing preventing the print method from working, though! (Or from using `._addr` to safely pass it somewhere else).

```python
import nanoarrow as na
import pyarrow as pa

na.c_array(pa.array([1, 2, 3]).__arrow_c_array__()[1])
# <nanoarrow.c_lib.CArray [invalid: schema is released]>
# - length: 3
# - offset: 0
# - null_count: 0
# - buffers: (0, 6021410128064)
# - dictionary: NULL
# - children[0]:
```

python/src/nanoarrow/c_lib.py
paleolimbot: I clarified this comment (yes, the child arrays). For every child there's an "allocate released `ArrowArray`" + "initialize the one with nanoarrow's release callback that lets you work with buffers"...technically we don't need the second half here but the validation is pretty nice to avoid memory leaks and if you have a gazillion columns you probably don't want to use this function anyway.

python/src/nanoarrow/_lib.pyx
jorisvandenbossche: I see. Can you add some docstrings to clarify that difference between "item_.." and "element_.."?

python/src/nanoarrow/_lib.pyx
```diff
@@ -1177,11 +1462,435 @@ cdef class CBufferView:
...
+
+        cdef int64_t count = ArrowBitCountSet(
+            validity_buffer.data,
+            self._ptr.offset,
+            self._ptr.length
+        )
+        self._ptr.null_count = self._ptr.length - count
+        return self
+
+    def set_buffer(self, int64_t i, CBuffer buffer, move=False):
```
jorisvandenbossche, line 1835: Can you add a bit of comment/docstring what this move=False vs True exactly does?

jorisvandenbossche: I see below that the docstring of the public function `c_array_from_buffers` has an explanation: "Use ``True`` to move ownership of any input buffers or children to the output array.". 

I would repeat that here for when looking at the code here. And it's also not super clear to me what the consequence is exactly of doing that? Or, why would you want to set it to True?

python/src/nanoarrow/c_lib.py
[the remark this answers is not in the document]
jorisvandenbossche: Hmm, yes, the second example of a generator is supported by `pa.array`, but eg `np.array` also doesn't work with that.

Let's leave this as a user's responsibility, for now, then

python/src/nanoarrow/_lib.pyx
paleolimbot: I added a docstring and also set it to `True` in the appropriate places in `c_lib`. It's to avoid the situation where you have an `ArrowBuffer` wrapping a `Py_buffer` wrapping an `ArrowBuffer` wrapping a `Py_buffer` (which is what would happened before the last commit if `move=False` and `buffers=(None, <some numpy array>)`).

jorisvandenbossche: Thanks, that expanded docstring explanation helps a lot!

PR #382, opened 2024-02-03
[r] Return package version as a package_version object

eddelbuettel:
The R package currently returns the (run-time or compile-time) package version as a string.  There is a simple 'quality of life' improvement to be had by wrapping this into `as.package_version()` as better comparison operators are available.  Default formatters exists so this should not have side-effects.

Also added a line to the documentation and re-ran `roxygenize()`.  

Please feel free to discard or close if this is any way against which package guidelines you may have here in the larger context of your organization.


eddelbuettel:
I see now that the `-SNAPSHOT` throws a spanner so I am closing this.

PR #384, opened 2024-02-12
chore: Add basic benchmark setup

paleolimbot:


assignUser (commented):
Nice, looks clean. Do you not want to use `find_package` to avoid system libraries for some reason (e.g. specific versions or something)?

PR #387, opened 2024-02-16
fix: Ensure that the deallocator called by ArrowBufferDeallocator() is called exactly once

paleolimbot:
This PR makes it more predictable to work with custom allocators, including the `ArrowBufferDeallocator` that powers the R and (soon) Python "build from buffers" behaviour. It also improves testing for using custom allocators and adds a templated C++ wrapper.

### DIFF
--- a/src/nanoarrow/nanoarrow.hpp
+++ b/src/nanoarrow/nanoarrow.hpp
@@ -307,7 +307,7 @@ static inline void BufferInitWrapped(struct ArrowBuffer* buffer, T obj, const vo
 /// Specifically, this uses obj.data() to set the buffer address and
 /// obj.size() * sizeof(T::value_type) to set the buffer size. This works
 /// for STL containers like std::vector, std::array, and std::string.
-/// This funciton moves obj and ensures it is deleted when ArrowBufferReset
+/// This function moves obj and ensures it is deleted when ArrowBufferReset
 /// is called.
 template <typename T>
 void BufferInitSequence(struct ArrowBuffer* buffer, T obj) {
--- a/src/nanoarrow/nanoarrow.hpp
+++ b/src/nanoarrow/nanoarrow.hpp
@@ -290,12 +290,14 @@ using UniqueArrayView = internal::Unique<struct ArrowArrayView>;
 /// \brief Initialize a buffer wrapping an arbitrary C++ object
 ///
 /// Initializes a buffer with a release callback that deletes the moved obj
-/// when ArrowBufferReset is called. T must be movable.
+/// when ArrowBufferReset is called. This version is useful for wrapping
+/// an object whose .data() member is missing or unrelated to the buffer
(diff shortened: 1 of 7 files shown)


src/nanoarrow/nanoarrow.hpp
```diff
@@ -273,6 +280,49 @@ using UniqueArrayView = internal::Unique<struct ArrowArrayView>;
...
+template <typename T>
+static inline void BufferInitWrapped(struct ArrowBuffer* buffer, T obj, const void* ptr,
+                                     int64_t size_bytes) {
+  T* obj_moved = new T(std::move(obj));
+  buffer->data = const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(ptr));
+  buffer->size_bytes = size_bytes;
+  buffer->capacity_bytes = 0;
+  buffer->allocator =
+      ArrowBufferDeallocator(&internal::DeallocateWrappedBuffer<T>, obj_moved);
+}
```
zeroshade, line 303: what kind of object is intended to be passed in here that would be valid to use as the data pointer?

src/nanoarrow/nanoarrow.hpp
```diff
@@ -273,6 +280,49 @@ using UniqueArrayView = internal::Unique<struct ArrowArrayView>;
...
+void BufferInitSequence(struct ArrowBuffer* buffer, T obj) {
+  // Move before calling .data() (matters sometimes).
+  T* obj_moved = new T(std::move(obj));
+  buffer->data =
+      const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(obj_moved->data()));
+  buffer->size_bytes = obj_moved->size() * sizeof(typename T::value_type);
+  buffer->capacity_bytes = 0;
+  buffer->allocator =
+      ArrowBufferDeallocator(&internal::DeallocateWrappedBuffer<T>, obj_moved);
+}
```
zeroshade, line 322: this is a really nice helper function :)

src/nanoarrow/nanoarrow.hpp
```diff
@@ -241,6 +241,13 @@ class Unique {
   T data_;
 };
 
+template <typename T>
+static inline void DeallocateWrappedBuffer(struct ArrowBufferAllocator* allocator,
+                                           uint8_t* ptr, int64_t size) {
+  auto obj = reinterpret_cast<T*>(allocator->private_data);
+  delete obj;
+}
```
zeroshade, line 249: definitely gonna save on creating lots of the same function :)

src/nanoarrow/nanoarrow.hpp
zeroshade: the problem with that is that it wouldn't be valid for that to be the `buffer->data` in the resulting `ArrowArray` object. Shouldn't that be placed in the `private_data`, not the `data`?

PR #391, opened 2024-02-20
feat(python): Add CArrayView -> Python conversion

paleolimbot:
This PR adds a framework for Python object creation from arrays and array streams with implementations for most arrow types. Notably, it includes implementations for nested types (struct, list, dictionary) to make sure that the framework won't have to be completely rewritten to accommodate them. A few types (decimal, datetime) aren't supported but should be reasonably easy to implement by wrapping existing iterator factories included in this PR.

None of these are exposed with `import nanoarrow as na` yet...I'm anticipating that the user-facing `nanoarrow.Array` and/or `nanoarrow.ArrayStream` to use the implementation here in methods.

A few changes were required at a lower level to make this work:

- It is now possible to use nanoarrow's `ArrowBasicArrayStream` implementation to create a stream from a previously-resolved list of arrays. This makes it easier to test streams since before we had no way to create them.
- The constructor for `c_array_stream()` now falls back on `c_array()` by wrapping it in a length-one stream. This makes it easier to write generic code that takes stream-like input (like the iterator).
- The `ArrowLayout` needed to be exposed to implement the fixed-size list implementation.
- I added tests for all the lower level changes, which I did in dedicated files. Some of these tests overlap with existing tests in test_nanoarrow...at some point we should go through test_nanoarrow and separate the tests (or create an integration test section since many of those early tests assumed pyarrow was available).

The implementation seems to be efficient given the constraint that assembling the iterators is currently done using Python code.

```python
import numpy as np
import pyarrow as pa
from nanoarrow import iterator

n = int(1e6)
n_cols = 10
arrays = [np.random.random(n) for _ in range(n_cols)]
batch = pa.record_batch(
    arrays,
    names=[f"col{i}" for i in range(n_cols)]
)

%timeit list(iterator.itertuples(batch))
#> 256 ms ± 4.61 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Just zipping the arrays
%timeit list(zip(*arrays))
#> 335 ms ± 1.15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# A few ways to do this from pyarrow
%timeit list(zip(*(col.to_pylist() for col in batch.columns)))
#> 1.99 s ± 52.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit list(zip(*(col.to_numpy() for col in batch.columns)))
#> 315 ms ± 1.33 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# Works if all columns are the same type (but rows are arrays, not tuples)
%timeit list(np.array(batch))
#> 131 ms ± 484 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

# Test some nested things
n = int(1e4)
n_cols = 10
big_list = [["a", "b", "c", "d", "e"]] * n

arrays = [big_list for _ in range(n_cols)]
batch = pa.record_batch(
    arrays,
    names=[f"col{i}" for i in range(n_cols)]
)

%timeit list(iterator.itertuples(batch))
#> 89.2 ms ± 756 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit list(zip(*(col.to_pylist() for col in batch.columns)))
#> 288 ms ± 748 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -1143,6 +1143,10 @@ cdef class CArrayView:
         self._device = CDEVICE_CPU
 
     @property
+    def storage_type_id(self):
+        return self._ptr.storage_type
+
+    @property
     def storage_type(self):
         cdef const char* type_str = ArrowTypeString(self._ptr.storage_type)
         if type_str != NULL:
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -767,6 +767,10 @@ cdef class CSchemaView:
         self._map_keys_sorted = schema._ptr.flags & ARROW_FLAG_MAP_KEYS_SORTED
 
     @property
+    def layout(self):
+        return CLayout(self, <uintptr_t>&self._schema_view.layout)
+
+    @property
     def type_id(self):
         return self._schema_view.type
 
@@ -859,6 +863,31 @@ cdef class CSchemaView:
 
     def __repr__(self):
         return _repr_utils.schema_view_repr(self)
+
+
+cdef class CLayout:
+    cdef ArrowLayout* _layout
+    cdef object _base
+
+    def __cinit__(self, base, uintptr_t ptr):
+        self._base = base
+        self._layout = <ArrowLayout*>ptr
+
+    @property
+    def buffer_type(self):
+        return tuple(self._layout.buffer_type[i] for i in range(3))
+
+    @property
+    def buffer_data_type_id(self):
+        return tuple(self._layout.buffer_data_type[i] for i in range(3))
+
+    @property
+    def element_size_bits(self):
+        return tuple(self._layout.element_size_bits[i] for i in range(3))
+
+    @property
+    def child_size_elements(self):
+        return self._layout.child_size_elements
 
 
 cdef class CSchemaBuilder:
@@ -1151,6 +1180,10 @@ cdef class CArrayView:
         cdef const char* type_str = ArrowTypeString(self._ptr.storage_type)
         if type_str != NULL:
             return type_str.decode('UTF-8')
+
+    @property
+    def layout(self):
+        return CLayout(self, <uintptr_t>&self._ptr.layout)
 
     @property
     def length(self):
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -1377,6 +1377,14 @@ cdef class CBufferView:
             return value
 
     def __iter__(self):
+        return self._iter_dispatch(0, len(self))
+
+    def _iter_dispatch(self, int64_t offset, int64_t length):
+        if offset < 0 or length < 0 or (offset + length) > len(self):
+            raise IndexError(
+                f"offset {offset} and length {length} do not describe a valid slice "
+                f"of buffer with length {len(self)}"
+            )
         # memoryview's implementation is very fast but not always possible (half float, fixed-size binary, interval)
         if self._data_type in (
             NANOARROW_TYPE_HALF_FLOAT,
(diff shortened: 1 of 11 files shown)


python/src/nanoarrow/_lib.pyx
```diff
@@ -861,6 +867,27 @@ cdef class CSchemaView:
...
+    cdef ArrowLayout* _layout
+    cdef object _base
+
+    def __cinit__(self, base, uintptr_t ptr):
+        self._base = base
+        self._layout = <ArrowLayout*>ptr
+
+    @property
+    def buffer_data_type_id(self):
+        return tuple(self._layout.buffer_data_type[i] for i in range(3))
```
danepitkin, line 880: Can we define the magic number 3 as a constant or variable somewhere? (or better yet, is it possible to take the `len()` of some object?)

python/tests/test_c_buffer_view.py
```diff
@@ -0,0 +1,103 @@
...
+
+    assert view.element_size_bits == 1
+    assert view.size_bytes == 1
+    assert view.data_type_id == na.Type.BOOL.value
+    assert view.data_type == "bool"
+    assert view.format == "B"
+
+    # Check item interface
+    assert len(view) == 1
+    assert view[0] == 1 + 8
```
danepitkin, line 36: nit: i'd prefer `0b1001` for readability. It took me a sec to realize what `1 + 8` meant haha

python/tests/test_c_buffer_view.py
```diff
@@ -0,0 +1,103 @@
...
+    # Check element interface
+    assert view.n_elements == 8
+    assert list(view.elements()) == [True, False, False, True] + [False] * 4
+    assert [view.element(i) for i in range(8)] == list(view.elements())
+
+    # Check element slices
+    assert list(view.elements(0, 4)) == [True, False, False, True]
+    assert list(view.elements(1, 3)) == [False, False, True]
+
+    with pytest.raises(IndexError, match="do not describe a valid slice"):
```
danepitkin, line 54: nit: I like to define static variables like strings once e.g. `msg = "<msg>"; with ... match=msg`

danepitkin (approved):
Nice work! I only added optional/minor comments. Overall LGTM

python/tests/test_c_buffer_view.py
paleolimbot: I didn't know you could do that!

python/src/nanoarrow/_lib.pyx
paleolimbot: This reminded me that I added NANOARROW_MAX_FIXED_BUFFERS a while ago (and also that the number of buffers had already been calculated elsewhere but made more sense here!)

python/src/nanoarrow/_lib.pyx
```diff
@@ -766,6 +766,10 @@ cdef class CSchemaView:
         self._nullable = schema._ptr.flags & ARROW_FLAG_NULLABLE
         self._map_keys_sorted = schema._ptr.flags & ARROW_FLAG_MAP_KEYS_SORTED
 
+    @property
+    def layout(self):
+        return CLayout(self, <uintptr_t>&self._schema_view.layout)
```
WillAyd, line 771: May or may not be an issue, but I think the canonical way to pass pointers via an integral value in CPython is to use the functions `PyLong_AsVoidPtr` and `PyLong_FromVoidPtr`

WillAyd (commented):
This is a cool looking PR - nice work @paleolimbot

python/src/nanoarrow/_lib.pyx
paleolimbot: I didn't know that...thanks! I am going to circle back closer to the next release and refactor some of the Cython structure and the patterns (like this one) that started early and proliferated more than I'd like.

PR #393, opened 2024-02-28
chore: Add basic benchmark suite to C library

paleolimbot:
This PR adds an initial set of benchmarks covering some realistic usage patterns. The general approach is to use doxygen comments to document the benchmarks, which will run against the released version and the previous version. I'm not sure exactly what the output format will be but I'd like the benchmarks to be written in such a way that there's a path to programatically generating a report (maybe using conbench, maybe just a Quarto document).

Work in progress!

### DIFF
--- a/src/nanoarrow/array_benchmark.cc
+++ b/src/nanoarrow/array_benchmark.cc
@@ -15,9 +15,6 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#include <algorithm>
-#include <iostream>
-
 #include <benchmark/benchmark.h>
 
 #include "nanoarrow.hpp"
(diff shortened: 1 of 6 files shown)


pitrou:
Can you show what the results look like?

src/nanoarrow/array_benchmark.cc
```diff
@@ -0,0 +1,135 @@
...
+/// \brief Use ArrowArrayViewGetIntUnsafe() to consume an int64 array (checking for nulls)
+static void BM_ArrayViewGetIntUnsafeInt64CheckNull(benchmark::State& state) {
+  BaseArrayViewGetIntUnsafe<int64_t, NANOARROW_TYPE_INT64>(state, 0.2);
+}
+
+BENCHMARK(BM_ArrayViewGetIntUnsafeInt8);
+BENCHMARK(BM_ArrayViewGetIntUnsafeInt16);
+BENCHMARK(BM_ArrayViewGetIntUnsafeInt32);
+BENCHMARK(BM_ArrayViewGetIntUnsafeInt64);
+BENCHMARK(BM_ArrayViewGetIntUnsafeInt64CheckNull);
```
felipecrv, line 135: hmm in Arrow C++ we don't start the benchmarks with `BM_`. Is that necessary?

pitrou:
> This is from the CI run (so timings are maybe meaningless), but the output looks like:

Thank you! This looks fine to me.

src/nanoarrow/array_benchmark.cc
paleolimbot: Good catch! I think it was rather contagious copy/paste from the first example in the benchmark library README 😬

PR #396, opened 2024-03-01
feat(python): Add user-facing `Array` class

paleolimbot:
This PR implements the `nanoarrow.Array` which basically a `pyarrow.ChunkedArray`. This can represent a `Table`, `RecordBatch`, `ChunkedArray`, and `Array`. It doesn't quite play nicely with pyarrow's ChunkedArray (but will after the next release, since `__arrow_c_stream__` was just added there).

The user-facing class is backed by a Cython class, the `CMaterializedArrayStream`, which manages some of the c-level details like resolving a chunk + offset when there is more than one chunk in the array. An early version of this PR implemented the `CMaterializedArrayStream` using C pointers (e.g., `ArrowArray* arrays`), but I decided that was to complex and went back to `List[CArray]`. I think this is also better for managing ownership (e.g., more unneeded `CArray` instances can be released by the garbage collector).

The `Array` class as implemented here is device-aware, although until we have non-CPU support it's difficult to test this. The methods I added here are basically stubs just to demonstrate the intention.

This PR also implements the `Scalar`, whose main purpose for testing and other non-performance sensitive things (like lazier reprs for very large items or interactive inspection). They may also be useful for working with arrays that contain elements with very long strings or large arrays (e.g., geometry).

I also added some basic accessors like `buffer()`, `child()`, and some ways one might want to iterate over an `Array` to make the utility of this class more clear. 

Basic usage:

```python
import nanoarrow as na

na.Array(range(100), na.int64())
```

```
nanoarrow.Array<int64>[100]
0
1
2
3
4
5
6
7
8
9
...and 90 more items
```

More involved example reading from an IPC stream:

```python
import nanoarrow as na
from nanoarrow.ipc import Stream

url = "https://github.com/apache/arrow-testing/raw/master/data/arrow-ipc-stream/integration/1.0.0-littleendian/generated_primitive.stream"

with Stream.from_url(url) as inp:
    array = na.Array(inp)

array.child(25)
```

```
nanoarrow.Array<string>[37]
'co矢2p矢m'
'wÂ€acrd'
'kjd1dlô'
'pib矢d5w'
'6nnpwôg'
'ndj£h£4'
'ôôf4aµg'
'kwÂh£fr'
'°g5dk€e'
'r€cbmdn'
...and 27 more items
```

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -2221,27 +2221,23 @@ cdef class CMaterializedArrayStream:
     def schema(self):
         return self._schema
 
-    @property
-    def array_ends(self):
-        return self._array_ends
-
     def __getitem__(self, k):
         cdef int64_t kint
         cdef int array_i
         cdef const int64_t* sorted_offsets = <int64_t*>self._array_ends._ptr.data
 
-        if not isinstance(k, slice):
-            kint = k
-            if kint < 0:
-                kint += self._total_length
-            if kint < 0 or kint >= self._total_length:
-                raise IndexError(f"Index {kint} is out of range")
+        if isinstance(k, slice):
+            raise NotImplementedError("index with slice")
 
-            array_i = ArrowResolveChunk64(kint, sorted_offsets, 0, len(self._arrays))
-            kint -= sorted_offsets[array_i]
-            return self._arrays[array_i], kint
+        kint = k
+        if kint < 0:
+            kint += self._total_length
+        if kint < 0 or kint >= self._total_length:
+            raise IndexError(f"Index {kint} is out of range")
 
-        raise NotImplementedError("index with slice")
+        array_i = ArrowResolveChunk64(kint, sorted_offsets, 0, len(self._arrays))
+        kint -= sorted_offsets[array_i]
+        return self._arrays[array_i], kint
 
     def __len__(self):
         return self._array_ends[len(self._arrays)]


python/src/nanoarrow/_ipc_lib.pyx
paleolimbot, line 1: These changes are because `Array(Stream.from_url(...))` crashed due to these methods not managing the GIL properly. I can move these to a separate PR.

jorisvandenbossche:
> An early version of this PR implemented the `CMaterializedArrayStream` using C pointers (e.g., `ArrowArray* arrays`), but I decided that was to complex and went back to `List[CArray]`. I think this is also better for managing ownership (e.g., more unneeded `CArray` instances can be released by the garbage collector).

If it would at some point turn out to be a perf bottleneck to use an untyped list, we can also consider compiling the cython code to C++ instead of C and then can use things like vector

python/src/nanoarrow/array.py
```diff
@@ -0,0 +1,243 @@
...
+            scalar = Scalar()
+            scalar._c_scalar = c_scalar
+            scalar._schema = self.schema
+            scalar._device = self._device
+            yield scalar
+
+    def __repr__(self) -> str:
+        width_hint = 80
+        n_items = 10
+        lines = [f"Array<{self.schema.type.name}>[{len(self)}]"]
```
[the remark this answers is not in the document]
jorisvandenbossche:
```suggestion
        lines = [f"nanoarrow.Array<{self.schema.type.name}>[{len(self)}]"]
```

We also might want to be consistent in including the library's name?

python/src/nanoarrow/_lib.pyx
```diff
@@ -2198,6 +2250,129 @@ cdef class CArrayStream:
...
+    def schema(self):
+        return self._schema
+
+    @property
+    def array_ends(self):
+        return self._array_ends
+
+    cdef int _resolve_chunk(self, const int64_t* sorted_offsets, int64_t index, int64_t start_offset_i,
+                           int64_t end_offset_i) noexcept nogil:
+        if start_offset_i >= (end_offset_i - 1):
```
jorisvandenbossche, line 2281: Can you add a bit of documentation here about the keywords?

python/src/nanoarrow/_lib.pyx
```diff
@@ -2198,6 +2250,129 @@ cdef class CArrayStream:
         return _repr_utils.array_stream_repr(self)
 
 
+cdef class CMaterializedArrayStream:
+    cdef CSchema _schema
+    cdef CBuffer _array_ends
+    cdef list _arrays
+    cdef int64_t _capacity_arrays
```
jorisvandenbossche, line 2257: This is not used (anymore)?

python/src/nanoarrow/_lib.pyx
```diff
@@ -50,6 +51,7 @@ from cpython.ref cimport Py_INCREF, Py_DECREF
 from nanoarrow_c cimport *
 from nanoarrow_device_c cimport *
 
+from functools import cached_property
```
jorisvandenbossche, line 54: Not used here?

python/src/nanoarrow/_lib.pyx
```diff
@@ -1066,10 +1068,15 @@ cdef class CArray:
...
+        cdef int64_t kint
+
         if not isinstance(k, slice):
-            raise TypeError(
-                f"Can't slice CArray with object of type {type(k).__name__}"
-            )
+            kint = k
+            if kint < 0:
+                kint += self._ptr.length
+            return CScalar(self, kint)
```
jorisvandenbossche, line 1079: Is this needed for CArray? We could also leave this for just the higher-level Array object to give scalars?

jorisvandenbossche (commented):
Nice work!
Some general thoughts:

- I think I certainly like having a "container" Array class that hides the complexity of Array vs ChunkedArray like we have in pyarrow. For certain use cases, that is definitely a lot easier to work with.
- But one downside of this is that there still isn't an easy to use but still low-level "ArrowArray" class that combines some features from `CArray` and `CArrayView`. For other use cases, it would be nice to dig into the object you have, check the buffers, understand what data you got, etc. Right now the Array class still gives you somewhat access to the CArray (see bullet point below as well), but when getting that you still need to do `na.c_array_view(c_arr)` to then get access to eg the Buffers.
- I am the least sure (or enthusiastic, that's probably the better word ;)) about adding a Scalar class. Probably there is no way around needing it, if you want to be able to represent single values while adhering to the C interface type system. But it would also be nice to keep it as limited as possible (for example, do we think someone should ever iterate over all values of the Array as Scalar objects?)

- Do we still want to expose the underlying CArrays a bit more publicly? Right now I can get them with `list(arr._data.arrays)`, but we should maybe make that public (maybe a `data` attribute that returns that list? Although "data" is of course a very generic term ;))

danepitkin:
> I am the least sure (or enthusiastic, that's probably the better word ;)) about adding a Scalar class. Probably there is no way around needing it, if you want to be able to represent single values while adhering to the C interface type system. But it would also be nice to keep it as limited as possible (for example, do we think someone should ever iterate over all values of the Array as Scalar objects?)

This idea came out of https://github.com/apache/arrow-nanoarrow/pull/397 where I suggested having types implement their own `repr`s vs having a top-down function that introspects how to represent types. Does this feel like a good use case for the Scalar class?

python/src/nanoarrow/_ipc_lib.pyx
danepitkin: I think leaving it in is fine!

python/src/nanoarrow/_ipc_lib.pyx
```diff
@@ -50,22 +50,32 @@ cdef extern from "nanoarrow_ipc.h" nogil:
...
+        self._close_stream = close_stream
+        self._addr = NULL
+        self._size_bytes = 0
+
+    @property
+    def obj(self):
+        return self._obj
+
+    @property
+    def close_stream(self):
```
danepitkin, line 69: Why does this object hold `close_stream` state? Do we not want `py_input_stream_release` to always close the stream?

python/src/nanoarrow/_lib.pyx
```diff
@@ -1192,6 +1199,51 @@ cdef class CArray:
         return _repr_utils.array_repr(self)
 
 
+cdef class CScalar:
+    """Low-level scalar implementation
+
+    This is an implementation of a "Scalar" that is a thin
+    wrapper around a (Python) CArray with the ability to
+    materialize a length-one array.
```
danepitkin, line 1207: Clever! :)

python/src/nanoarrow/_lib.pyx
danepitkin: optional: i challenge you to implement this iteratively instead of recursively to save on stack space usage ;)

python/src/nanoarrow/array.py
```diff
@@ -0,0 +1,243 @@
...
+        if self._schema is None:
+            self._schema = Schema(self._c_scalar.schema)
+        return self._schema
+
+    def as_py(self):
+        """Get the Python object representation of this scalar"""
+        return next(iterator(self._c_scalar))
+
+    def __repr__(self) -> str:
+        width_hint = 80
```
danepitkin, line 74: Should we make the width_hint configureable via an additional api such as `to_string()`? Then `repr` can call `self.to_string(length=80)`

python/src/nanoarrow/array.py
```diff
@@ -0,0 +1,243 @@
...
+        """Get the device on which the buffers for this array are allocated."""
+        return self._device
+
+    @cached_property
+    def schema(self) -> Schema:
+        """Get the schema (data type) of this Array"""
+        return Schema(self._data.schema)
+
+    @property
+    def n_chunks(self) -> int:
```
danepitkin, line 141: nit: I'd prefer `num_chunks` for naming

python/src/nanoarrow/array.py
```diff
@@ -0,0 +1,243 @@
...
+        """Get the number of chunks in the underlying representation of this Array."""
+        return self._data.n_arrays
+
+    @property
+    def chunks(self) -> Iterable:
```
danepitkin, line 146: nit: I think `iter_chunks()` could be a better name and is better defined as a function. I typically think of properties as a static value.

danepitkin (commented):
Awesome work!

python/src/nanoarrow/array.py
```diff
@@ -0,0 +1,243 @@
...
+    --------
+
+    >>> import nanoarrow as na
+    >>> na.array([1, 2, 3], na.int32())
+    Array<INT32>[3]
+    1
+    2
+    3
+    """
+    return Array(obj, schema=schema)
```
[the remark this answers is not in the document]
paleolimbot: I modelled this after the fact that we have `na.schema()`. I would be game to drop `na.schema()` and `na.array()` in favour of the regular class constructor!

Good call on the error!

python/src/nanoarrow/_lib.pyx
paleolimbot: Good call! Just leaving myself a mental note that the Arrow C++ implementation is here: https://github.com/apache/arrow/blob/1eb46f763a73d313466fdc895eae1f35fac37945/cpp/src/arrow/chunk_resolver.h#L133-L162

python/src/nanoarrow/_ipc_lib.pyx
paleolimbot: It currently only closes a file object if the higher level Python wrapping this was the code that opened it (e.g., it won't close a connection that a user opened in case the user had other plans or the IPC data was only part of the file payload).

paleolimbot: Probably this should be `close_file_obj` to disambiguate the three types of streams we have going on here (ArrowArray, ArrowIpcInput, Python file object).

python/src/nanoarrow/array.py
paleolimbot: Good call! That's the `pa.ChunkedArray` name too 👍

python/src/nanoarrow/array.py
paleolimbot: For better or worse, nanoarrow is pretty deep in the "properties return iterators" pattern (e.g., `schema.children` and many others). pyarrow's ChunkedArray has a `chunks` property that returns a `list()` rather than an iterator, which was the vague inspiration. For now maybe I'll keep this returning an iterator but I'm definitely down to circle back and fix all the existing usage since I agree that it's a potentially confusing use of the `@property`.

python/src/nanoarrow/array.py
```diff
@@ -0,0 +1,243 @@
...
+from typing import Iterable
+
+from nanoarrow._lib import CDEVICE_CPU, CArray, CDevice, CMaterializedArrayStream
+from nanoarrow.c_lib import c_array, c_array_stream
+from nanoarrow.iterator import iterator
+from nanoarrow.schema import Schema
+
+
+class Scalar:
+    """Generic wrapper around an :class:`Array` element
```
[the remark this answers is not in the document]
paleolimbot: Actually, I didn't think of it when I wrote this, but the `CScalar` probably should *not* implement `__arrow_c_array__` because it is not necessarily representing a CPU array. To me the main utility of the Scalar class is on the interactive end of things (where the non-Cython wrapper gives a better doc/IDE experience).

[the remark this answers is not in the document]
paleolimbot: Hmm...then the `CArray` would face the same problem and it would be very strange if the `CArray` did not implement `__arrow_c_array__`. I'll add it to the `Scalar` for now...we probably need the `device_id` and `device_type` in the `CArray` 🙁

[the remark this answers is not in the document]
paleolimbot: Good call...I removed the CScalar!

python/src/nanoarrow/array.py
paleolimbot: Just kidding...good call! I organized all the fun ways you can iterate over an `Array` into a bunch of `iter_xxx()` methods, including `iter_chunks()`.

python/src/nanoarrow/_lib.pyx
paleolimbot: Ok, this is now `ArrowResolveChunk64()` in the C library. It will also be needed for random access into REE arrays and I'd like to use it in the R package to do something similar to this PR in the R package.

python/src/nanoarrow/array.py
paleolimbot: I went back to `n_chunks` here because everywhere else in the package uses `n_xxxxx`. I can't revert some of those because they're the same as the `ArrowArray`/`ArrowSchema` C structs and it would be strange if they were different 🤷

danepitkin:
From a user perspective, I would guess that most all python users want to interact with objects like `Array` and `Scalar`, while seeing other objects as implementation details to be avoided, such as `CArray` and `CMaterializedArray`.

python/src/nanoarrow/_lib.pyx
```diff
@@ -2198,6 +2200,130 @@ cdef class CArrayStream:
...
+    def __iter__(self):
+        for c_array in self._arrays:
+            for item_i in range(c_array.length):
+                yield c_array, item_i
+
+    def array(self, int64_t i):
+        return self._arrays[i]
+
+    @property
+    def n_arrays(self):
```
danepitkin, line 2258:
```suggestion
    def num_arrays(self):
```

just if we want to be consistent with `num_chunks`

python/src/nanoarrow/array.py
```diff
@@ -0,0 +1,498 @@
...
+        """
+        return self._device
+
+    @cached_property
+    def schema(self) -> Schema:
+        """Get the schema (data type) of this Array"""
+        return Schema(self._data.schema)
+
+    @property
+    def n_buffers(self) -> int:
```
danepitkin, line 210:
```suggestion
    def num_buffers(self) -> int:
```

You should consider my `n` vs `num` suggestion optional, but I do think its best to be consistent. I personally try to avoid as much ambiguity when naming things, hence my minor preference for `num`.

danepitkin (commented):
Wow, this PR is beautiful. I gave it a quick pass and overall love the direction its taken. It generally LGTM, but I'll go back and review the smaller details while giving Joris a chance for a re-review as well.

python/src/nanoarrow/array.py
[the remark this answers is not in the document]
jorisvandenbossche: > Actually, I didn't think of it when I wrote this, but the `CScalar` probably should _not_ implement `__arrow_c_array__` because it is not necessarily representing a CPU array.

Yeah, it's true that things are not fully correct regarding devices right now, but I think that is something we can tackle later. i.e. when there is better device support, we would add `__arrow_c_device_array__` in addition, which always works, and let `__arrow_c_array__` raise an error or something like that.

The same happens in pyarrow actually. A pyarrow.Array can right now be backed by non-CPU data, but we still have a `_arrow_c_array__` defined on it.

python/src/nanoarrow/array.py
jorisvandenbossche: > For better or worse, nanoarrow is pretty deep in the "properties return iterators" pattern (e.g., `schema.children` and many others). pyarrow's ChunkedArray has a `chunks` property that returns a `list()` rather than an iterator, which was the vague inspiration.

Is there a performance reason that you prefer to work with iterators? 
In most of those cases, the data already exists (most what might happen is that it gets wrapped in a new class), and so all those things should be pretty cheap?

If that's the case, I think it would actually be nice to keep the properties but let them just return a list.

python/src/nanoarrow/_lib.pyx
danepitkin: Hmm, I think num_chunks was removed in a follow up PR already!

paleolimbot: I'm pretty sure they're all `n_xxx` now but will double check! I agree that `num_xxx` is better, but that's a separate change to consider since we already have `CArray.n_children`, `CArray.n_buffers`, `CArrayView.n_buffers`, `CArrayView.n_children`, `Schema.n_fields`, `CSchema.n_fields`, `CLayout.n_buffers`, etc.

danepitkin: No worries, if there's already a precedent then you can ignore me. n_* isn't bad either

danepitkin (approved):
LGTM! Just a few small optional comments

python/src/nanoarrow/_lib.pyx
```diff
@@ -2198,6 +2200,130 @@ cdef class CArrayStream:
...
+    @property
+    def array_ends(self):
+        return self._array_ends
+
+    def __getitem__(self, k):
+        cdef int64_t kint
+        cdef int array_i
+        cdef const int64_t* sorted_offsets = <int64_t*>self._array_ends._ptr.data
+
+        if not isinstance(k, slice):
```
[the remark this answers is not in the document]
paleolimbot: Done!

python/src/nanoarrow/_lib.pyx
```diff
@@ -2198,6 +2200,130 @@ cdef class CArrayStream:
...
+
+    cdef _finalize(self):
+        self._array_ends._set_data_type(NANOARROW_TYPE_INT64)
+
+    @property
+    def schema(self):
+        return self._schema
+
+    @property
+    def array_ends(self):
```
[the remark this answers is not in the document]
paleolimbot: Done!

PR #397, opened 2024-03-02
feat(python): Add repr iterator

paleolimbot:
The existing iterators are optimized for completely consuming input; however, this is very bad if (e.g.) you have one element that's a few hundred MB in size. Work in progress!

### DIFF
--- a/python/src/nanoarrow/iterator.py
+++ b/python/src/nanoarrow/iterator.py
@@ -211,6 +211,9 @@ class RowIterator(ArrayViewIterator):
                 yield list(islice(child_iter, fixed_size))
 
     def _string_iter(self, offset, length):
+        return self._binary_iter(offset, length, lambda x: str(x, "UTF-8"))
+
+    def _binary_iter(self, offset, length, fun=bytes):
         view = self._array_view
         offset += view.offset
         offsets = memoryview(view.buffer(1))[offset : (offset + length + 1)]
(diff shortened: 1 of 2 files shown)


python/src/nanoarrow/iterator.py
```diff
@@ -276,6 +301,217 @@ def _iter1(self, offset, length):
...
+    def full(self):
+        return self._size > self._max_size
+
+    def write_null(self):
+        return self.write(repr(None))
+
+    def write(self, content):
+        """Write string content. Returns True if there are not yet max_size
+        characters in this items's repr and False otherwise."""
+        self._out.write(content)
```
danepitkin, line 500: What if instead we only write the first bytes of content equal to `self._max_size - self._size`? This could get rid of unnecessary writing of data that would otherwise be truncated in `finish()`

python/src/nanoarrow/iterator.py
```diff
@@ -276,6 +301,217 @@ def _iter1(self, offset, length):
...
+        self._size += len(content)
+        return not self.full()
```
danepitkin, line 502: Typically, I expect a `write()` API to return nothing or return True only if the call succeeded. I'd recommend leaving the caller to check `self.full()`.

python/src/nanoarrow/iterator.py
```diff
@@ -83,8 +84,42 @@ def itertuples(obj, schema=None) -> Iterable[Tuple]:
     return RowTupleIterator.get_iterator(obj, schema=schema)
 
 
+def iterrepr(obj, schema=None, max_width: int = 80) -> Iterable[str]:
```
danepitkin, line 87: Hmm, how would you feel about approaching this a different way? Instead of 
```
    >>> for item in iterator.iterrepr(array, max_width=9):
    ...     print(item)
```
we could potentially do something like
```
    >>> for item in array:
    ...     print(item.to_string(max_width=9))
```
This would be more of a bottoms-up approach rather than top-down, where each data type has its own `to_string()` method that gets called and we don't need to keep track of the nested depth in an auxiliary class like `ReprIterator`

python/src/nanoarrow/iterator.py
```diff
@@ -276,6 +301,217 @@ def _iter1(self, offset, length):
...
+
+    def __init__(self, schema, *, _array_view=None, max_width=120, out=None):
+        # Pass the same instance of out to all children. If unspecified,
+        # this is the top-level iterator that will request the value from
+        # out for each item.
+        if out is None:
+            self._out = ReprBuilder(max_width)
+            self._top_level = True
+        else:
+            self._out = out
```
danepitkin, line 320: Why would a caller want to specify their own `out`?

python/src/nanoarrow/iterator.py
```diff
@@ -276,6 +301,217 @@ def _iter1(self, offset, length):
...
+            self._top_level = False
+
+        super().__init__(schema, _array_view=_array_view)
+
+    def _make_child(self, schema, array_view):
+        # Ensure children all refer to the same instance of the ReprBuilder
+        return ReprIterator(schema, _array_view=array_view, out=self._out)
+
+    def _iter1(self, offset, length):
```
danepitkin, line 329: Can we find a function name that is more descriptive? I find this particular function to be quite complex.

danepitkin (commented):
I did leave some larger comments this time around. I'll be curious to know your thoughts. Great job getting this up and running!

danepitkin:
Sounds great! I really like your proposed approach

PR #399, opened 2024-03-13
fix: Relax comparison strictness such that integration tests pass

paleolimbot:
These changes are the changes required such that https://github.com/apache/arrow/pull/39302 results in passing integration tests for nanoarrow. The changes are mostly related to comparison:

- We needed an option to allow metadata to be compared on a key/value basis without considering order (for Java, which seems to reorder metadata on read)
- We needed the ability to treat NULL metadata and zero-size metadata as equivalent (for Go, which always exports zero-length metadata)
- We needed an option to ignore flags for top-level batches (for C#, which exports nullable structs)
- We needed to ensure that the last few bits of the validity buffer were zeroed (for C#, although this is now fixed in C# on Arrow main)
- We needed to ensure that no buffers were NULL (For C#, which leaks the top-level array if it encounters one, at least in the integration tests. This should really be fixed in C#).

### DIFF
--- a/src/nanoarrow/nanoarrow_testing.hpp
+++ b/src/nanoarrow/nanoarrow_testing.hpp
@@ -2551,22 +2551,16 @@ class TestingJSONComparison {
   /// Some Arrow implementations store metadata using structures (e.g., hash map) that
   /// reorder metadata items. Use false to consider metadata whose keys/values have
   /// been reordered as equivalent.
-  void set_compare_metadata_order(bool value) { compare_metadata_order_ = value; }
+  void set_compare_metadata_order(bool value) {
+    compare_metadata_order_ = value;
 
-  /// \brief Consider omitted metadata equivalent to empty metadata
-  ///
-  /// Some Arrow implementations always write an empty metadata string (i.e.,
-  /// \0\0\0\0) instead of omitting it (i.e., nullptr). T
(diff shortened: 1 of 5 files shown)


src/nanoarrow/nanoarrow_testing.hpp
```diff
@@ -2749,6 +2790,79 @@ class TestingJSONComparison {
...
+
+  ArrowErrorCode MetadataToMap(const char* metadata,
+                               std::unordered_map<std::string, std::string>* out) {
+    ArrowMetadataReader reader;
+    NANOARROW_RETURN_NOT_OK(ArrowMetadataReaderInit(&reader, metadata));
+
+    ArrowStringView key, value;
+    while (reader.remaining_keys > 0) {
+      NANOARROW_RETURN_NOT_OK(ArrowMetadataReaderRead(&reader, &key, &value));
+      out->insert({std::string(key.data, key.size_bytes),
```
benibus, line 2862: Just to be sure, can we safely rule out the possibility of duplicate keys by this point?

paleolimbot: Ok! I now check for duplicate keys, although technically the comparison will just fail if this happens. The real solution is probably to fix the Java implementation to stop reordering metadata and add test cases for this in the archery data generator.

PR #404, opened 2024-03-18
feat: Add `ArrowArray` and `ArrowArrayStream` C++ iterators 

### LINKED ISSUE #403: Add c++ range-for loop helpers
Benchmarks and tests are written in C++ and currently there's a lot of boilerplate around looping through streams and containers. It'd improve readability to add some helpers for easier iteration:

```c++
struct ArrowArrayStream* stream = get_stream();
for (const ArrowArray* array : *stream) {
  for (Maybe<int32_t> value : ArrayAs<int32_t>(array)) {
    // ...
  }
}
```

For anything we don't need to build with compilers that don't support C++17, we could also add ergonomic `zip` and `enumerate` utilities for parallel iteration of two sequences:

```c++
for (auto [i, f, idx] : Zip(ArrayAs<int32_t>(ints), ArrayAs<float>(floats)), Enumerate) {
  // ...
}
```

bkietz:
All of this *should* be c++11

```diff
-  auto arrow_array = ImportArray(&array, large_utf8());
-  ARROW_EXPECT_OK(arrow_array);
-
-  auto builder = LargeStringBuilder();
-  ARROW_EXPECT_OK(builder.Append("1234"));
-  ARROW_EXPECT_OK(builder.AppendNulls(2));
-  ARROW_EXPECT_OK(builder.Append("56789"));
-  ARROW_EXPECT_OK(builder.AppendEmptyValue());
-  auto expected_array = builder.Finish();
-
-  EXPECT_TRUE(arrow_array.ValueUnsafe()->Equals(expected_array.ValueUnsafe()));
+  EXPECT_THAT(nanoarrow::ViewAsBytes<64>(&array),
+              ElementsAre("1234"_v, kNull, kNull, "56789"_v, ""_v));
```

fixes #403


src/nanoarrow/nanoarrow.hpp
```diff
@@ -548,6 +558,366 @@ class VectorArrayStream {
...
+  const_iterator begin() const { return range_.begin(); }
+  const_iterator end() const { return range_.end(); }
+  value_type operator[](int64_t i) const { return range_.get(i); }
+};
+
+struct ErrorWithCode : ArrowError {
+  ErrorWithCode() : ArrowError{}, code{NANOARROW_OK} {}
+  ArrowErrorCode code;
+  constexpr operator bool() const { return code != NANOARROW_OK; }
+};
```
paleolimbot, line 806: The error handling is definitely the most annoying part of having the range-based for be perfect (in the absence of exceptions). It's not perfect, but

```cpp
ViewStream viewer(stream, &error);
for (ArrowArray* item : viewer) {
  // stuff
}
NANOARROW_RETURN_NOT_OK(viewer.code());
```

...is the same number of lines as 

```cpp
ErrorWithCode error;
for (ArrowArray* item : ViewStream(stream, &error)) {
  // stuff
}
NANOARROW_RETURN_NOT_OK(error.code);
```

...and invents one less concept. It would be a mega bonus if we could crash (in debug mode) if somebody forgets to check the return code (not checking the return code is a frequent source of crashes at unrelated lines of code that has been a bit of a struggle to communicate to users).

src/nanoarrow/nanoarrow.hpp
```diff
@@ -548,6 +558,366 @@ class VectorArrayStream {
...
+  ArrowError* error_;
+  ArrowErrorCode* code_;
+
+ public:
+  using iterator = typename InputRange<Next>::iterator;
+  iterator begin() { return range_.begin(); }
+  iterator end() { return range_.end(); }
+};
+
+#if NANOARROW_HAS_ZIP
```
paleolimbot, line 848: I love the Zip! I wonder if `dev/benchmarks/c/utils.h` would be a better home for this at the moment since that's where it can be used. The range-based iterator and the `ViewAs` have direct connections to nanoarrow C structs and so this header feels like a good place for them...these are generic helpers and I'm not sure that *this* header is the best home (for now).

paleolimbot (commented):
I know you're still working on this...just a few early comments!

src/nanoarrow/nanoarrow.hpp
```diff
@@ -548,6 +558,366 @@ class VectorArrayStream {
...
+
+  using value_type = typename RandomAccessRange<Get>::value_type;
+  using const_iterator = typename RandomAccessRange<Get>::const_iterator;
+  const_iterator begin() const { return range_.begin(); }
+  const_iterator end() const { return range_.end(); }
+  value_type operator[](int64_t i) const { return range_.get(i); }
+};
+
+template <>
+class ViewAsBytes<0> {
```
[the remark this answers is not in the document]
bkietz: It's a hack to reuse ViewAsBytes to indicate fixed size binary. I'll just rename it

src/nanoarrow/nanoarrow.hpp
```diff
@@ -548,6 +558,366 @@ class VectorArrayStream {
...
+    value_type operator*() const { return *stashed; }
+  };
+
+  iterator begin() { return {this}; }
+  iterator end() { return {}; }
+};
+
+// C++17 could do all of this with a lambda
+template <typename T>
+class ViewAs {
```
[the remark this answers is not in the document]
bkietz: If it views a single buffer it can't detect null bits

src/nanoarrow/nanoarrow.hpp
bkietz: I was trying to let ViewStream use an existing ArrowError if one is already in scope, which would make an internal one redundant. I can assert the return code is checked in either case. I'll rewrite to what you suggest

src/nanoarrow/nanoarrow.hpp
bkietz: After I move this, should I remove the preprocessor conditions in favor of raising an error if !NANOARROW_HAS_ZIP? https://github.com/apache/arrow-nanoarrow/pull/404/files#diff-52c5c898fc0e86beb4c6c267d8cc09469f26695a2e401b4ee6c49613226432f8R106

paleolimbot: Definitely! (Not great to have two versions floating around)

src/nanoarrow/nanoarrow.hpp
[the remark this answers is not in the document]
paleolimbot: So maybe `ViewArrayAs` since it's intended for viewing arrays? If there's an STL term that would help a user understand what this is for that may also be useful to put in the name (e.g., is this an STL "sequence"? Maybe `ArrayAsSequenceOf<int_32>()`?)

src/nanoarrow/nanoarrow.hpp
lidavidm: Presumably `view` comes from ranges: https://en.cppreference.com/w/cpp/ranges/view

src/nanoarrow/nanoarrow.hpp
```diff
@@ -548,6 +550,312 @@ class VectorArrayStream {
 
 /// @}
 
+struct Nothing {};
```
paleolimbot, line 553: Do you have bandwidth to add tests for these? (I can try also)

paleolimbot (commented):
```
In file included from /home/runner/work/arrow-nanoarrow/arrow-nanoarrow/dev/benchmarks/c/array_benchmark.cc:22:
  /home/runner/work/arrow-nanoarrow/arrow-nanoarrow/dev/benchmarks/c/util.h:25:2: error: #error "Zip cannot be supported without C++17 features"
     25 | #error "Zip cannot be supported without C++17 features"
        |  ^~~~~
```

...might be fixed by `set(CMAKE_CXX_STANDARD 17)` in the `dev/benchmarks/CMakeLists.txt` (unless you were checking that it failed on purpose, in which case ignore me 🙂 ).

src/nanoarrow/nanoarrow.hpp
bkietz: I've added a test for Maybe, do you want tests for the range utils as well?

src/nanoarrow/utils_test.cc
```diff
@@ -538,3 +538,36 @@ TEST(DecimalTest, DecimalRoundtripBitshiftTest) {
 
   ArrowBufferReset(&buffer);
 }
+
+TEST(MaybeTest, ConstructionAndConversion) {
+  // NOTE these will not print nicely in GTest unless we write PrintTo,
```
paleolimbot, line 543: Does defining `<<` work?

https://github.com/apache/arrow-nanoarrow/blob/dc50114756b7e9067b42181a6a86f928effc6e68/extensions/nanoarrow_ipc/src/nanoarrow/nanoarrow_ipc_files_test.cc#L309-L312

bkietz: It would, but I will still need to add a source file to house it

bkietz: ... or if it's not a problem, let nanoarrow.hpp `#include <iostream>`

src/nanoarrow/nanoarrow.hpp
paleolimbot: Sorry I missed this! I think it makes sense for anything user facing to have a test. There is `docker compose run --rm coverage` to check.

src/nanoarrow/utils_test.cc
paleolimbot: Hmm...right now nanoarrow.hpp is mostly about digging a pit of success for not leaking memory and is pretty minimal. That said, the failure mode for tests that use it is currently not all that informative:

```
[ctest] /Users/dewey/Desktop/rscratch/arrow-nanoarrow/src/nanoarrow/array_test.cc:577: Failure
[ctest] Value of: nanoarrow::ViewArrayAs<uint32_t>(&array)
[ctest] Expected: has 2 elements where
[ctest] element #0 is equal to 8-byte object <02-00 00-00 01-00 00-00>,
[ctest] element #1 is equal to 8-byte object <03-00 00-00 01-00 00-00>
[ctest]   Actual: { 8-byte object <01-00 00-00 01-00 00-00>, 8-byte object <03-00 00-00 01-00 00-00> }, whose element #0 doesn't match
```

I would perhaps lean towards `#include <iostream>` for the purposes of this PR, and before the release we can figure out how best to structure the C++ component. Maybe something like `src/nanoarrow/nanoarrow_cpp(hpp?)` could contain a few files with smaller scopes (e.g., the current `nanoarrow.hpp`), and `src/nanoarrow/nanoarrow.hpp` would include all of them? That would let anybody who cared keep a minimal scope but let most people benefit from the epic that this PR contains?

lidavidm: I think it would work to define PrintTo in a different header + compilation unit that you could include only in tests.

src/nanoarrow/nanoarrow.hpp
bkietz: This doesn't seem to handle templates too well. I've added tests for RandomAccessRange and InputRange as well. If we bring in ArrayFromJSON or something, we could also have meaningful tests for View*

paleolimbot: Thanks! I'm comfortable not adding tests, although if there are no tests it should probably live in `nanoarrow_testing.hpp` until we have the collective bandwidth to add them?

src/nanoarrow/array_test.cc
```diff
@@ -30,8 +32,11 @@
 #include <arrow/util/decimal.h>
 
 #include "nanoarrow/nanoarrow.h"
+#include "nanoarrow/nanoarrow.hpp"
```
paleolimbot, line 35:
```suggestion
#include "nanoarrow/nanoarrow.hpp"
```

src/nanoarrow/nanoarrow.hpp
```diff
@@ -222,6 +221,9 @@ class Unique {
   T* operator->() noexcept { return &data_; }
   const T* operator->() const noexcept { return &data_; }
 
+  /// \brief Check for validity
+  explicit operator bool() const { return data_.release != nullptr; }
```
paleolimbot, line 225: This seems like it might be confusing for something like `nanoarrow::UniqueBuffer` that doesn't have a `.release` callback?

src/nanoarrow/nanoarrow.hpp
```diff
@@ -550,6 +552,314 @@ class VectorArrayStream {
...
+      : ViewArrayStream{stream, &internal_code_, &internal_error_} {}
+
+  // disable copy/move of this view, since its error references may point into itself
+  ViewArrayStream(ViewArrayStream&&) = delete;
+  ViewArrayStream& operator=(ViewArrayStream&&) = delete;
+  ViewArrayStream(const ViewArrayStream&) = delete;
+  ViewArrayStream& operator=(const ViewArrayStream&) = delete;
+
+  // ensure the error code of this stream was accessed at least once
+  ~ViewArrayStream() { NANOARROW_DCHECK(code_was_accessed_); }
```
paleolimbot, line 814: Thank you for adding this bit!

src/nanoarrow/nanoarrow.hpp
```diff
@@ -550,6 +552,314 @@ class VectorArrayStream {
...
+  };
+
+  internal::InputRange<Next> range_;
+  ArrowError* error_;
+  ArrowErrorCode* code_;
+  ArrowError internal_error_ = {};
+  ArrowErrorCode internal_code_;
+  bool code_was_accessed_;
+  // ArrowErrorCode internal_code_ = NANOARROW_OK;
+  // bool code_was_accessed_ = false;
```
paleolimbot, line 839: Do these need to be removed?

paleolimbot (commented):
Thank you for working on this!

Just small comments from me...the big one is that I'd prefer to put the `ViewArrayAs` helper in `nanoarrow_testing.hpp` if we don't have the bandwidth to dedicate test coverage for it right now.

src/nanoarrow/nanoarrow.hpp
paleolimbot: I know it is a bit circular, but testing the "create by buffer, consume by View" direction for at least one instantation of each template in the user-facing header is what I had in mind. Creating by buffer is not too bad in C++ using the pattern from the benchmarks:

https://github.com/apache/arrow-nanoarrow/blob/db6630b7770d33edc5227f60929a269a4aff8a04/dev/benchmarks/c/array_benchmark.cc#L36-L43

...plus the newly added helpers to create buffers from C++ sequences (which we can't use in the bechmarks or they won't build against the previous version):

https://github.com/apache/arrow-nanoarrow/blob/db6630b7770d33edc5227f60929a269a4aff8a04/src/nanoarrow/nanoarrow_hpp_test.cc#L206-L234

src/nanoarrow/nanoarrow.hpp
bkietz: Fair enough; I'll remove this. Unless there are validity conditions for all the other objects wrapped by Unique which I could specialize this for? For example, in the case of `nanoarrow::UniqueBuffer` I could check whether `buffer->data` is null

bkietz: Well, that can be a follow up then.

paleolimbot (commented):
(Pushes to the branch at Ben's invitation 🙂 )

Thank you for bearing with me on this PR and thank you for the tests! I made a few changes to satisfy the constraints enforced by CI here:

- Because we test against the bundled version (as defined by `cmake . -NANOARROW_BUNDLE=ON && cmake --install .`), the extra file had to be bundled in the same way as all the others.
- The benchmarks are specifically designed to build against previous versions (so that they can be run by downstream users to determine if they should upgrade their version pin). This means that `ViewArrayAs` would have to be guarded with an `#ifdef`, but then different versions would be benchmarking different code. It seems as though the use of `Zip` is mostly useful alongside `ViewArrayAs`, so I removed that particular change from this PR (but saved those changes and pushed them to #443). I'm sorry for not catching that at an earlier stage in the review 😬 

Before this merges I think the new helpers need at least a cursory documentation section since this is a user-facing change (I'm happy to circle back and add this and will definitely do so before the next release).

paleolimbot (approved):
Thank you!

PR #412, opened 2024-03-26
chore(dev/benchmarks): Add Python/asv benchmarking setup

paleolimbot:
This PR adds a basic asv benchmarking setup for the Python bindings with a small suite of high-level benchmarks tracking a few of the opportunities I happen to know involve some Python looping.


.gitignore
```diff
@@ -26,3 +26,5 @@ CMakeUserPresets.json
 .Rproj.user
 .cache
 *.swp
+nanoarrow-python-asv
```
jorisvandenbossche, line 29: Does asv create this directory?

asv.conf.json
jorisvandenbossche, line 1: FWIW, if you would prefer to not clutter the root of the repo more, this file can also live in `dev/benchmarks`

jorisvandenbossche (commented):
Haven't looked in detail at what is being benchmarked, but the setup looks good to me!

.gitignore
paleolimbot: It did before I figured out how asv works! (Removed!)

asv.conf.json
paleolimbot: Nice! (Done!)

PR #417, opened 2024-04-05
feat(python): Add Arrow->Python datetime support

paleolimbot:
This PR adds support for converting Arrow date, time, timestamp, and duration arrays to Python objects. 

```python
import pyarrow as pa
import datetime
import zoneinfo
import nanoarrow as na

dt = datetime.datetime.now()
list(na.Array(pa.array([dt])).iter_py())
#> [datetime.datetime(2024, 4, 8, 16, 25, 41, 216438)]

dt_tz = datetime.datetime.now(zoneinfo.ZoneInfo("America/Halifax"))
list(na.Array(pa.array([dt_tz])).iter_py())
#> [datetime.datetime(2024, 4, 8, 16, 29, 7, 226832, tzinfo=zoneinfo.ZoneInfo(key='America/Halifax'))]

tdelta = datetime.timedelta(123, 456, 678)
list(na.Array(pa.array([tdelta])).iter_py())
#> [datetime.timedelta(days=123, seconds=456, microseconds=678)]

just_time = datetime.time(15, 27, 43, 12)
list(na.Array(pa.array([just_time])).iter_py())
#> [datetime.time(15, 27, 43, 12)]
```

It is probably faster to use the DateTime C API, but the timings seem reasonable:

```python
import pyarrow as pa
import datetime
import zoneinfo
import nanoarrow as na

n = int(1e6)

dt = datetime.datetime.now()
dt_array = pa.array([dt + datetime.timedelta(i) for i in range(n)])
%timeit dt_array.to_pylist()
%timeit list(na.Array(dt_array).iter_py())
#> 805 ms ± 21.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
#> 804 ms ± 1.79 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

tdelta_array = pa.array([datetime.timedelta(123 + i, 456, 678) for i in range(n)])
%timeit tdelta_array.to_pylist()
#> 574 ms ± 3.78 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit list(na.Array(tdelta_array).iter_py())
#> 399 ms ± 612 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

just_time_array = pa.array([datetime.time(15, 27, 43, i) for i in range(n)])
%timeit just_time_array.to_pylist()
#> 831 ms ± 6.59 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit list(na.Array(just_time_array).iter_py())
#> 399 ms ± 856 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/python/src/nanoarrow/iterator.py
+++ b/python/src/nanoarrow/iterator.py
@@ -245,7 +245,7 @@ class PyIterator(ArrayViewIterator):
                 yield bytes(data[start:end])
 
     def _timestamp_iter(self, offset, length):
-        from datetime import UTC, datetime
+        from datetime import datetime
 
         fromtimestamp = datetime.fromtimestamp
 
@@ -267,7 +267,7 @@ class PyIterator(ArrayViewIterator):
             tz_fromtimestamp = tz
         else:
             tz = None
-            tz_fromtimestamp = UTC
+            tz_fromtimestamp = _get_tzinfo("UTC")
 
         for parent in self._primitive_iter(offset, length):
             if parent is None:
--- a/python/src/nanoarrow/iterator.py
+++ b/python/src/nanoarrow/iterator.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import warnings
 from functools import cached_property
 from itertools import islice
 from typing import Iterable, Tuple
@@ -83,6 +84,14 @@ def iter_tuples(obj, schema=None) -> Iterable[Tuple]:
     return RowTupleIterator.get_iterator(obj, schema=schema)
 
 
+class InvalidArrayWarning(UserWarning):
+    pass
+
+
+class LossyConversionWarning(UserWarning):
+    pass
+
+
 class ArrayViewIterator:
     """Base class for iterators that use an internal ArrowArrayView
     as the basis for conversion to Python objects. Intended for internal use.
@@ -115,6 +124,13 @@ class ArrayViewIterator:
     def _child_names(self):
         return [child.name for child in self._schema.children]
 
+    @cached_property
+    def _object_label(self):
+        if self._schema.name:
+            return f"{self._schema.name} <{self._schema_view.type}>"
+        else:
+            return f"<unnamed {self._schema_view.type}>"
+
     def _contains_nulls(self):
         return (
             self._schema_view.nullable
@@ -125,6 +141,9 @@ class ArrayViewIterator:
     def _set_array(self, array):
         self._array_view._set_array(array)
         return self
+
+    def _warn(self, message, category):
+        warnings.warn(f"{self._object_label}: {message}", category)
 
 
 class PyIterator(ArrayViewIterator):
@@ -271,6 +290,9 @@ class PyIterator(ArrayViewIterator):
                 yield None
             else:
                 days, hours, mins, secs, us = item
+                if days != 0:
+                    self._warn("days != 0", InvalidArrayWarning)
+
                 yield time(hours, mins, secs, us)
 
     def _timestamp_iter(self, offset, length):
@@ -287,7 +309,7 @@ class PyIterator(ArrayViewIterator):
         elif unit == "us":
             scale = 1_000_000
         elif unit == "ns":
-            storage = _scale_and_round_maybe_none(storage, 0.001)
+            storage = self._iter_us_from_ns(storage)
             scale = 1_000_000
 
         tz = self._schema_view.timezone
@@ -298,12 +320,12 @@ class PyIterator(ArrayViewIterator):
             tz = None
             tz_fromtimestamp = _get_tzinfo("UTC")
 
-        for parent in storage:
-            if parent is None:
+        for item in storage:
+            if item is None:
                 yield None
             else:
-                s = parent // scale
-                us = parent % scale * (1_000_000 // scale)
+                s = item // scale
+                us = item % scale * (1_000_000 // scale)
                 yield fromtimestamp(s, tz_fromtimestamp).replace(
                     microsecond=us, tzinfo=tz
                 )
@@ -321,7 +343,7 @@ class PyIterator(ArrayViewIterator):
         elif unit == "us":
             to_us = 1
         elif unit == "ns":
-            storage = _scale_and_round_maybe_none(storage, 0.001)
+            storage = self._iter_us_from_ns(storage)
             to_us = 1
 
         for item in storage:
@@ -341,7 +363,7 @@ class PyIterator(ArrayViewIterator):
         elif unit == "us":
             to_us = 1
         elif unit == "ns":
-            storage = _scale_and_round_maybe_none(storage, 0.001)
+            storage = self._iter_us_from_ns(storage)
             to_us = 1
 
         us_per_sec = 1_000_000
@@ -363,6 +385,15 @@ class PyIterator(ArrayViewIterator):
                 secs = us // us_per_sec
                 us = us % us_per_sec
                 yield days, hours, mins, secs, us
+
+    def _iter_us_from_ns(self, parent):
+        for item in parent:
+            if item is None:
+                yield None
+            else:
+                if item % 1000 != 0:
+                    self._warn("nanoseconds discarded", LossyConversionWarning)
+                yield item // 1000
 
     def _primitive_iter(self, offset, length):
         view = self._array_view
@@ -432,14 +463,6 @@ def _get_tzinfo(tz_string, strategy=None):
     raise RuntimeError(
         "zoneinfo (Python 3.9+), pytz, or dateutil is required to resolve timezone"
     )
-
-
-def _scale_and_round_maybe_none(parent, scale):
-    for item in parent:
-        if item is None:
-            yield None
-        else:
-            yield round(item * scale)
 
 
 _ITEMS_ITER_LOOKUP = {
--- a/python/src/nanoarrow/iterator.py
+++ b/python/src/nanoarrow/iterator.py
@@ -244,6 +244,36 @@ class PyIterator(ArrayViewIterator):
             for start, end in zip(starts, ends):
                 yield bytes(data[start:end])
 
+    def _timestamp_iter(self, offset, length):
+        from datetime import datetime
+        import dateutil
+
+        unit = self._schema_view.time_unit
+        if unit == "s":
+            scale = 1
+        elif unit == "ms":
+            scale = 1000
+        elif unit == "us":
+            scale = 1_000_000
+        elif unit == "ns":
+            raise NotImplementedError(
+                "Timestamp with unit 'ns' is not currently supported"
+            )
+
+        tz = self._schema_view.timezone
+        if tz:
+            tz = dateutil.tz.gettz(tz)
+        else:
+            tz = None
+
+        for parent in self._primitive_iter(offset, length):
+            if parent is None:
+                yield None
+            else:
+                s = parent // scale
+                us = parent % scale * (1_000_000 // scale)
+                yield datetime.fromtimestamp(s, tz).replace(microsecond=us)
+
     def _primitive_iter(self, offset, length):
         view = self._array_view
         offset += view.offset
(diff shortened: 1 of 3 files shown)


eddelbuettel:
(Micro-nit:  Missing second r in Arrow in Subject)

python/src/nanoarrow/iterator.py
```diff
@@ -244,6 +244,126 @@ def _binary_iter(self, offset, length):
...
+
+        storage = self._primitive_iter(offset, length)
+        epoch = date(1970, 1, 1)
+
+        if self._schema_view.type_id == CArrowType.DATE32:
+            for item in storage:
+                if item is None:
+                    yield item
+                else:
+                    yield epoch + timedelta(item)
```
jorisvandenbossche, line 258: `datetime.date` also has a `fromtimestamp` method that accepts seconds since epoch, so you could also use that, which seems to be slightly faster:

```
In [53]: item = 10957

In [54]: epoch = datetime.date(1970, 1, 1)

In [55]: epoch + datetime.timedelta(item)
Out[55]: datetime.date(2000, 1, 1)

In [56]: datetime.date.fromtimestamp(60*60*24*item)
Out[56]: datetime.date(2000, 1, 1)

In [57]: %timeit epoch + datetime.timedelta(item)
164 ns ± 0.569 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [58]: %timeit datetime.date.fromtimestamp(60*60*24*item)
118 ns ± 0.0869 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

jorisvandenbossche: And the same for DATE64 if converting the milliseconds to seconds

python/src/nanoarrow/iterator.py
```diff
@@ -244,6 +244,126 @@ def _binary_iter(self, offset, length):
...
+
+        tz = self._schema_view.timezone
+        if tz:
+            tz = _get_tzinfo(tz)
+            tz_fromtimestamp = tz
+        else:
+            tz = None
+            tz_fromtimestamp = _get_tzinfo("UTC")
+
+        for parent in storage:
```
jorisvandenbossche, line 301:
```suggestion
        for item in storage:
```

nitpick to be consistent in naming? (above you are using item)

python/src/nanoarrow/iterator.py
```diff
@@ -244,6 +244,126 @@ def _binary_iter(self, offset, length):
...
+            if parent is None:
+                yield None
+            else:
+                s = parent // scale
+                us = parent % scale * (1_000_000 // scale)
+                yield fromtimestamp(s, tz_fromtimestamp).replace(
+                    microsecond=us, tzinfo=tz
+                )
```
jorisvandenbossche, line 309: I think `fromtimestamp` should work with fractional seconds? 

```
In [63]: import time

In [64]: time.time()
Out[64]: 1712740039.6300623

In [65]: datetime.datetime.fromtimestamp(time.time())
Out[65]: datetime.datetime(2024, 4, 10, 11, 7, 34, 904224)
```

So in that case could we do a normal division and just pass that to fromtimestamp?

python/src/nanoarrow/iterator.py
```diff
@@ -244,6 +244,126 @@ def _binary_iter(self, offset, length):
...
+
+        unit = self._schema_view.time_unit
+        if unit == "s":
+            scale = 1
+        elif unit == "ms":
+            scale = 1000
+        elif unit == "us":
+            scale = 1_000_000
+        elif unit == "ns":
+            storage = _scale_and_round_maybe_none(storage, 0.001)
```
jorisvandenbossche, line 290: Should we silently discard the nanoseconds? An alternative could be to raise an error if the nanoseconds are not zero, or warn.

python/src/nanoarrow/iterator.py
```diff
@@ -244,6 +244,126 @@ def _binary_iter(self, offset, length):
...
+                )
+
+    def _duration_iter(self, offset, length):
+        from datetime import timedelta
+
+        storage = self._primitive_iter(offset, length)
+
+        unit = self._schema_view.time_unit
+        if unit == "s":
+            to_us = 1_000_000
```
jorisvandenbossche, line 318: For unit of seconds, it's probably a bit more efficient to pass the seconds directly to `timedelta(seconds=..)` than first converting to microseconds, because then the `timedelta` constructor will then convert the microseconds back to seconds internally.

Don't know if that is worth it though, because of course the current logic makes the loop a bit simpler

python/src/nanoarrow/iterator.py
```diff
@@ -244,6 +244,126 @@ def _binary_iter(self, offset, length):
...
+
+    def _time_iter(self, offset, length):
+        from datetime import time
+
+        for item in self._iter_datetime_components(offset, length):
+            if item is None:
+                yield None
+            else:
+                days, hours, mins, secs, us = item
+                yield time(hours, mins, secs, us)
```
jorisvandenbossche, line 274: `days` is just ignored here. Should we assert that it is 0? (the Arrow spec strictly speaking says that the value should never exceed the number of seconds/milliseconds/.. of 1 day, so for valid data we can be sure to not have a `day` here)

python/src/nanoarrow/iterator.py
```diff
@@ -244,6 +244,126 @@ def _binary_iter(self, offset, length):
...
+            storage = _scale_and_round_maybe_none(storage, 0.001)
+            to_us = 1
+
+        for item in storage:
+            if item is None:
+                yield None
+            else:
+                yield timedelta(microseconds=item * to_us)
+
+    def _iter_datetime_components(self, offset, length):
```
jorisvandenbossche, line 333: Is `_iter_time_components` a better name? Because I think this is only used for the time type, right?

python/src/nanoarrow/iterator.py
```diff
@@ -278,6 +398,50 @@ def _iter1(self, offset, length):
         return self._struct_tuple_iter(offset, length)
 
 
+def _get_tzinfo(tz_string, strategy=None):
+    # We can handle UTC without any imports
+    if tz_string.upper() == "UTC":
+        try:
+            # Added in Python 3.11
+            from datetime import UTC
```
jorisvandenbossche, line 406: This is an alias of `datetime.timezone.utc` (lower case), which might be available in older versions as well. So could simplify to use that

python/src/nanoarrow/iterator.py
```diff
@@ -278,6 +398,50 @@ def _iter1(self, offset, length):
...
+            return timezone(timedelta(0), "UTC")
+
+    # Try zoneinfo.ZoneInfo() (Python 3.9+)
+    if strategy is None or "zoneinfo" in strategy:
+        try:
+            from zoneinfo import ZoneInfo
+
+            return ZoneInfo(tz_string)
+        except ImportError:
+            pass
```
jorisvandenbossche, line 421: If the timezone doesn't exist (Arrow C++ itself never does any validation of this, so this is typically "user input"), then this will actually raise a `ZoneInfoNotFoundError`, not `ImportError`. And you could also catch a KeyError to keep it generic.

```
In [76]: zoneinfo.ZoneInfo("bal")
...
ZoneInfoNotFoundError: 'No time zone found with key bal'

In [77]: zoneinfo.ZoneInfoNotFoundError.mro()
Out[77]: 
[zoneinfo._common.ZoneInfoNotFoundError,
 KeyError,
 LookupError,
 Exception,
 BaseException,
 object]
```

I suppose the ImportError was for older Python versions, but so I would do that (`import zoneinfo`) in a separate try/except

python/src/nanoarrow/iterator.py
```diff
@@ -278,6 +398,50 @@ def _iter1(self, offset, length):
...
+    if strategy is None or "dateutil" in strategy:
+        try:
+            from dateutil.tz import gettz
+
+            return gettz(tz_string)
+        except ImportError:
+            pass
+
+    raise RuntimeError(
+        "zoneinfo (Python 3.9+), pytz, or dateutil is required to resolve timezone"
```
jorisvandenbossche, line 433:
```suggestion
        "zoneinfo (Python 3.9+) or dateutil is required to resolve timezone"
```

python/src/nanoarrow/iterator.py
jorisvandenbossche: Ah, but of course you are just catching the ImportError to then be able to try dateutil, and so the `ZoneInfoNotFoundError` will just be bubbled up to the user, which is fine. So ignore my comment ;)

jorisvandenbossche (commented):
Nice! Added a bunch of comments

For the zoneinfo vs dateutil, we _could_ also bump the minimum Python version from 3.8 to 3.9, and then we can always assume `zoneinfo` is available (Python 3.8 is almost end-of-life, see https://devguide.python.org/versions/).

For timezones, there is one aspect not covered by this PR. The Arrow spec also allows fixed offsets of the form "+XX:XX" or "-XX:XX". Those can be handled with creating a timedelta and constructing a `datetime.timezone()` from that, I think, see https://docs.python.org/3/library/datetime.html#datetime.timezone) 
Fixed offsets are not very useful, so it's fine to not (yet) handle that, but maybe add a TODO comment about it, and we should maybe also raise a better error message in `_get_tzinfo` about it)

python/src/nanoarrow/iterator.py
paleolimbot: Nice! I do see in the Python documentation ( https://docs.python.org/3/library/datetime.html#datetime.date.fromtimestamp )

> It’s common for this to be restricted to years from 1970 through 2038

...which is ominous (if up-to-date, which it might not be anymore).

jorisvandenbossche: Hmm, good point. But this is true for `datetime.datetime.fromtimestamp` as well, though ..

python/src/nanoarrow/iterator.py
paleolimbot: I bit the bullet and added a warning system that warns in the case that `days != 0` (although it should probably be a bit smarter and emit fewer warnings at some point).

python/src/nanoarrow/iterator.py
paleolimbot: I think there is a lot that could be optimized here...this pass is mostly for completeness/correctness. Probably this is a job for C or C++ + and Python C API where we can do some of these things efficiently.

python/src/nanoarrow/iterator.py
paleolimbot: The iterator will now emit a `LossyConversionWarning` when this happens (although maybe eventually we want to make this quieter).

python/src/nanoarrow/iterator.py
paleolimbot: I think there is a precision issue that kicks in with floating point representations of timestamps. I changed this to use an `epoch` + reusing the duration iterator, although this is quite a bit slower than using `fromtimestamp()`. Eventually this is a job for C or C++.

python/src/nanoarrow/iterator.py
paleolimbot: I changed this to avoid `fromtimestamp()` for now in both cases (even though it's quite a bit slower) and added some dates that will make a future test fail if this is actually a problem. Optimizing these conversions is probably a good project for some future (less busy) time.

python/src/nanoarrow/iterator.py
```diff
@@ -278,6 +416,44 @@ def _iter1(self, offset, length):
         return self._struct_tuple_iter(offset, length)
 
 
+def _get_tzinfo(tz_string, strategy=None):
+    import re
+    from datetime import timedelta, timezone
+
+    # We can handle UTC without any imports
+    if re.search(r"^utc$", tz_string, re.IGNORECASE):
```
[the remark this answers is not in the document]
paleolimbot: I changed it back! It was a paper-thin theory that maybe it was less confusing to use `re` since it was used a few lines later, too.

PR #421, opened 2024-04-09
chore: Clean up Compiler warnings

WillAyd:
When compiling with the meson configuration which sets -Wall and -Wextra I noticed a lot of warnings for unused parameters and sign comparisons

I think this is also worth enforcing in CI - any objection to updating the current CMake jobs? Or maybe we can add one for meson to run on every PR?

### DIFF
--- a/dev/benchmarks/c/array_benchmark.cc
+++ b/dev/benchmarks/c/array_benchmark.cc
@@ -231,7 +231,7 @@ static void BenchmarkArrayViewGetString(benchmark::State& state) {
 
   int64_t n_alphabets = n_values / alphabet.size() + 1;
   std::vector<char> data(alphabet.size() * n_alphabets);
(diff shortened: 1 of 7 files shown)


WillAyd:
Ah nice work on that PR

> We could also add `-Werror` to the Meson build (I assume there's a way to do this?) and see how often that is a problem.

To get this added to meson all we would have to do is add `werror=true` to the project `default_options` configuration in the top level meson.build file

WillAyd:
I noticed that some of the warnings were coming upstream from Arrow. See https://github.com/apache/arrow/pull/41107#issuecomment-2046276564 and https://github.com/apache/arrow/pull/41111

(there are still more warnings than these but I think this is a good chunk)

WillAyd:
Yea its a tough balance. I would probably leave it out of the local configuration but if we run meson in CI can also just pass `-Dwerror=true` during setup / configure to get the same effect

PR #422, opened 2024-04-10
Increase CMake minimum version, move requirement

eddelbuettel:
This is a very minor 'quality of life' improvement to `linesplitter` as `cmake` nags about the minimum requirement being after (rather than before) the main project declaration, and increasing the version (which I understand you may not want to increase by as much) silences the TIMESTAMP nag.  Feel free to adjust or ignore, I just liked the quieter build.

### DIFF
--- a/examples/linesplitter/CMakeLists.txt
+++ b/examples/linesplitter/CMakeLists.txt
@@ -15,10 +15,16 @@
 # specific language governing permissions and limitations
 # under the License.
 
-cmake_minimum_required(VERSION 3.24)
+cmake_minimum_required(VERSION 3.14)
 project(linesplitter)
 
 set(CMAKE_CXX_STANDARD 11)
+
+# Avoids a warning about timestamps on downloaded files (prefer new policy
+# if available))
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.23")
+  cmake_policy(SET CMP0135 NEW)
+endif()
 
 include(FetchContent)
 
--- a/examples/linesplitter/CMakeLists.txt
+++ b/examples/linesplitter/CMakeLists.txt
@@ -15,8 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
+cmake_minimum_required(VERSION 3.24)
 project(linesplitter)
-cmake_minimum_required(VERSION 3.14)
 
 set(CMAKE_CXX_STANDARD 11)
 


examples/linesplitter/CMakeLists.txt
```diff
@@ -15,8 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
+cmake_minimum_required(VERSION 3.24)
```
paleolimbot, line 18: The earliest version we run in CI is 3.17 (`cmake3` on centos7), and I think 3.16 is the version that is on Ubuntu 20.04. It is probably the most informative to put 3.16 here (in case anybody reads this as "the minimum CMake version you can use with nanoarrow).

paleolimbot (commented):
Thanks!

examples/linesplitter/CMakeLists.txt
eddelbuettel: Sure feel free to flip it back or lower it 'per policy'.  For me as a CMake neophyte it was the cheapest way to silence this:

```-- Building using CMake version: 3.27.4
CMake Warning (dev) at /usr/share/cmake-3.27/Modules/FetchContent.cmake:1316 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of         
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.                                 
Call Stack (most recent call first):
  CMakeLists.txt:38 (fetchcontent_declare)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found Python: /usr/bin/python3 (found version "3.11.6") found components: Interpreter 
```

which trust you know how to silence otherwise.  By all means commit into the branch (if the remainder is of use to you) to address this otherwise.

paleolimbot: Great! Happy to clean this up and merge in a bit.

We use this to silence that message:

https://github.com/apache/arrow-nanoarrow/blob/7f6e4df3f853d179ca3b17e0ab8ba06bc4e9b803/CMakeLists.txt#L53-L57

...but I think you can do `if(POLICY ...)` too.

eddelbuettel: That was easy enough so I just attempted a cleanup.  Back to 3.14 and now a policy setter.  Should have looked on directory up...

paleolimbot (approved):

PR #425, opened 2024-04-11
feat(python): Support Decimal types in convert to Python

paleolimbot:
I experimented with a few different methods here...I think this one has a good balance of speed and not messing with the global precision.

```python
import pyarrow as pa
import decimal
from nanoarrow.iterator import iter_py

items = [decimal.Decimal("12.3450"), None, decimal.Decimal("1234567.3456")]
array = pa.array(items, pa.decimal128(11, 4))
list(iter_py(array))
#> [Decimal('12.3450'), None, Decimal('1234567.3456')]
```

This seems to be vaguely on par with pyarrow convert:

```python
import pyarrow as pa
import decimal
import numpy as np
from nanoarrow.iterator import iter_py

floats = np.random.random(int(1e6))
items = [decimal.Decimal(item) for item in floats]
array = pa.array(items)

%timeit array.to_pylist()
#> 799 ms ± 6.24 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit list(iter_py(array))
#> 431 ms ± 8.65 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/python/src/nanoarrow/iterator.py
+++ b/python/src/nanoarrow/iterator.py
@@ -264,7 +264,7 @@ class PyIterator(ArrayViewIterator):
                 yield bytes(data[start:end])
 
     def _decimal_iter(self, offset, length):
-        from decimal import Decimal, Context
+        from decimal import Context, Decimal
         from sys import byteorder
 
         storage = self._primitive_iter(offset, length)
--- a/python/src/nanoarrow/iterator.py
+++ b/python/src/nanoarrow/iterator.py
@@ -263,6 +263,25 @@ class PyIterator(ArrayViewIterator):
             for start, end in zip(starts, ends):
                 yield bytes(data[start:end])
 
+    def _decimal_iter(self, offset, length):
+        from decimal import Decimal, Context
+        from sys import byteorder
+
+        storage = self._primitive_iter(offset, length)
+        precision = self._schema_view.decimal_precision
+        scaleb = Decimal(-self._schema_view.decimal_scale)
+        context = Context(prec=precision)
+
+        # It would probably be fastest to go straight from binary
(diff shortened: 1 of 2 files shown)


python/src/nanoarrow/iterator.py
```diff
@@ -263,6 +263,25 @@ def _binary_iter(self, offset, length):
             for start, end in zip(starts, ends):
                 yield bytes(data[start:end])
 
+    def _decimal_iter(self, offset, length):
+        from decimal import Context, Decimal
+        from sys import byteorder
+
+        storage = self._primitive_iter(offset, length)
+        precision = self._schema_view.decimal_precision
+        scaleb = Decimal(-self._schema_view.decimal_scale)
```
jorisvandenbossche, line 272: Does this need to be converted to a Decimal? 
The docs seem to suggest this can be just an integer (https://docs.python.org/3/library/decimal.html#decimal.Decimal.scaleb)

Also, I might add some comments about what is going on for people not familiar with this method (I had to look it up to understand what you wrote, but certainly makes sense now)

jorisvandenbossche (approved):
Looks good!

python/src/nanoarrow/iterator.py
paleolimbot: Apparently it does! https://github.com/python/cpython/blob/main/Lib/_pydecimal.py#L3684

I'll add some comments here.

jorisvandenbossche:
Not clear if this started with this PR, but on main the merged commit from this PR is failing in the wheel building: for Python 3.12, running the tests is segfaulting (https://github.com/apache/arrow-nanoarrow/actions/runs/8689330033/job/23826851854)

PR #431, opened 2024-04-16
feat(python): Implement extension type and Schema metadata support

paleolimbot:
The initial motivation of this PR was to ensure that extension types are handled in nanoarrow's Schemas; however, this exposed that metadata was not handled by the `Schema` in either create or consume mode.

After this PR, extension types can be created and inspected. This is really just creating a schema with some metadata (and looking for specific metadata when consuming).

```python
import nanoarrow as na

schema = na.extension_type(na.int32(), "arrow.example", b'{"some_param": 1234}')
if schema.type == na.Type.EXTENSION:
    print(schema.extension.name, schema.extension.metadata)
#> arrow.example b'{"some_param": 1234}'
```

In doing some testing, there were a number of places where extension schemas/arrays were implicitly treated as their storage types. I've tried to error or warn for these cases as much as possible:

```python
import nanoarrow as na

ext = na.extension_type(na.int32(), "arrow.example")
na.Array([1, 2, 3], ext)
#> TypeError: ...
#> Can't create buffer from extension type arrow.example
```

It's a little hard to create an extension array at the moment (and there should probably be a similar option to strip the extension type from an `Array` to just get the storage), but I think that is maybe a job for another PR that is more about arrays and less about schemas:

```python
import nanoarrow as na

ext = na.extension_type(na.int32(), "arrow.example")
storage = na.c_array([1, 2, 3], ext.extension.storage)
_, storage_capsule = storage.__arrow_c_array__()
extension = na.Array(storage_capsule, ext)
list(extension.iter_py())
#> UnregisteredExtensionWarning: <unnamed int32>: Converting unregistered extension 'arrow.example' as storage type
#> [1, 2, 3]
```

A side effect of all of this is that there is better support for modifying schemas:

```python
import nanoarrow as na

na.Schema(na.int32(), name="some_col")
#> Schema(INT32, name='some_col')
schema = na.Schema(na.int32(), metadata={"some_key": "some_value"})
schema.metadata[b"some_key"]
#> b'some_value

na.c_schema(na.int32()).modify(
    name="some_col",
    metadata={"some_key": "some_value"},
    nullable=False
)
#> <nanoarrow.c_lib.CSchema int32>
#> - format: 'i'
#> - name: 'some_col'
#> - flags: 0
#> - metadata:
#>   - b'some_key': b'some_value'
#> - dictionary: NULL
#> - children[0]:
```


jorisvandenbossche:
Do we want to go the route of a registry and having users define their own?

For nanoarrow, I would personally stick to what it in essence is: metadata (and we have the advantage that here the "data type" (ArrowSchema) of an array actually has this field metadata, and so we don't have the problem like in pyarrow that the Array object looses that information)

jorisvandenbossche:
(we could of course still provide some more ergonomic access to the extension name/metadata, e.g. by detecting if those keys are present, and in that case showing them more prominent in certain reprs, or provide an easier getter property for them, etc)

jorisvandenbossche:
> The part where the concept of a registry is difficult to avoid (global or tightly scoped) is when doing conversion to/from Python objects with nested type support. 

The user has always cheap access to the storage type, or the raw data, and can do a custom conversion from that. For the example of Polygons, I wouldn't want some iterative approach to convert to python objects, I would just want to have access to the underlying coordinates and offsets to do a custom bulk conversion.

I personally would punt on this problem of custom conversion for now, and see what kind of needs arise (we don't even support this in pyarrow at the moment)

If some kind of registry is needed specifically for the conversion, the sqlite interface could be an inspiration? (not necessarily the exact names, because I find those a bit confusing. But the fact that you register just something for conversion from/to python, and that also doesn't necessarily need to be limited to extension types. Users might want to influence how for example a map or interval type is converted to python)

python/src/nanoarrow/_lib.pyx
```diff
@@ -967,10 +989,50 @@ cdef class CSchemaBuilder:
         if self._ptr.release == NULL:
             ArrowSchemaInit(self._ptr)
 
+    @staticmethod
+    def copy_existing(CSchema existing_schema):
+        return CSchemaBuilder(existing_schema.__deepcopy__())
```
danepitkin, line 994: nit: I'd prefer a shorter name. WDYT? 
```suggestion
    def copy(CSchema schema):
        return CSchemaBuilder(schema.__deepcopy__())
```

python/src/nanoarrow/_lib.pyx
```diff
@@ -967,10 +989,50 @@ cdef class CSchemaBuilder:
...
+
     @staticmethod
     def allocate():
         return CSchemaBuilder(CSchema.allocate())
 
+    def clear_metadata(self):
```
danepitkin, line 1000: Optional: What if we made this more object oriented? e.g. `CSchemaBuilder.clear_metadata()` vs `CSchemaBuilder.metadata.clear()`. Same with `append_metadata`. Does it make sense to implement this way? If not (or its a lot of work), ignore this comment.

python/src/nanoarrow/iterator.py
```diff
@@ -160,6 +164,17 @@ def get_iterator(cls, obj, schema=None):
                 yield from iterator._iter1(0, array.length)
 
     def _iter1(self, offset, length):
+        # Check for and extension type first since this isn't reflected by
```
danepitkin, line 167:
```suggestion
        # Check for an extension type first since this isn't reflected by
```

python/src/nanoarrow/schema.py
```diff
@@ -108,10 +115,73 @@ def create(obj):
...
+        extension_metadata = self._schema._c_schema_view.extension_metadata
+        return extension_metadata if extension_metadata else None
+
+    @property
+    def storage(self):
+        """Storage type for this extension type"""
+        metadata = dict(self._schema.metadata.items())
+        del metadata[b"ARROW:extension:name"]
+        if b"ARROW:extension:metadata" in metadata:
+            del metadata[b"ARROW:extension:metadata"]
```
danepitkin, line 141: why do we delete the objects here?

python/src/nanoarrow/schema.py
```diff
@@ -108,10 +115,73 @@ def create(obj):
...
+        return Schema(self._schema, metadata=metadata)
+
+
 class Schema:
-    """The Schema is nanoarrow's high-level data type representation whose scope maps to
-    that of the ArrowSchema in the Arrow C Data interface. See :func:`schema` for class
-    details.
+    """Create a nanoarrow Schema
+
+    The Schema is nanoarrow's high-level data type representation, encompasing
```
danepitkin, line 149:
```suggestion
    The Schema is nanoarrow's high-level data type representation, encompassing
```

danepitkin (commented):
Left some minor comments! Nice work

python/src/nanoarrow/schema.py
paleolimbot: I added a comment! (Removing the metadata keys is required, or else the schema will still be treated as an extension).

python/src/nanoarrow/_lib.pyx
paleolimbot: At this level of abstraction it would be hard (since it is basically wrapping a single C API call) but I agree that it would be a better interface!

Issue #435, opened 2024-04-19
python: inspect or view function to print details of a (C)Array?


jorisvandenbossche:
When you have an Array object and want to inspect the details (buffers, children, etc), you currently essentially have 2x 2 functions that each give slightly different information, the `c_array`, `c_array_view`, `c_schema`, `c_schema_view`. 
In the tutorial material we are writing, we are mostly using `c_array_view` as that gives the most comprehensive and understandable repr.

Example:

```python
In [1]: import nanoarrow as na

In [2]: pa_arr = pa.array([[0, 1], [None, 3], None, [4]])

In [3]: arr = na.Array(pa_arr)

In [4]: arr
Out[4]: 
nanoarrow.Array<list<item: int64>>[4]
[0, 1]
[None, 3]
None
[4]

In [5]: na.c_array(arr)
Out[5]: 
<nanoarrow.c_lib.CArray list<item: int64>>
- length: 4
- offset: 0
- null_count: 1
- buffers: (140041326198848, 140041326198784)
- dictionary: NULL
- children[1]:
  'item': <nanoarrow.c_lib.CArray int64>
    - length: 5
    - offset: 0
    - null_count: 1
    - buffers: (140041326198912, 140041326198976)
    - dictionary: NULL
    - children[0]:

In [6]: na.c_array_view(arr)
Out[6]: 
<nanoarrow.c_lib.CArrayView>
- storage_type: 'list'
- length: 4
- offset: 0
- null_count: 1
- buffers[2]:
  - validity <bool[1 b] 11010000>
  - data_offset <int32[20 b] 0 2 4 4 5>
- dictionary: NULL
- children[1]:
  - <nanoarrow.c_lib.CArrayView>
    - storage_type: 'int64'
    - length: 5
    - offset: 0
    - null_count: 1
    - buffers[2]:
      - validity <bool[1 b] 11011000>
      - data <int64[40 b] 0 1 0 3 4>
    - dictionary: NULL
    - children[0]:

In [7]: na.c_schema(arr.schema)
Out[7]: 
<nanoarrow.c_lib.CSchema list>
- format: '+l'
- name: ''
- flags: 2
- metadata: NULL
- dictionary: NULL
- children[1]:
  'item': <nanoarrow.c_lib.CSchema int64>
    - format: 'l'
    - name: 'item'
    - flags: 2
    - metadata: NULL
    - dictionary: NULL
    - children[0]:

In [8]: na.c_schema_view(arr.schema)
[...]

PR #436, opened 2024-04-20
feat(python): function to inspect a single-chunk Array

jorisvandenbossche:
Addresses https://github.com/apache/arrow-nanoarrow/issues/435

Reusing the existing repr utilities, the current code gives this:

```
In [1]: import nanoarrow as na

In [2]: pa_arr = pa.array([[0, 1], [None, 3], None, [4]])

In [3]: arr = na.Array(pa_arr)

In [4]: arr.inspect()
<ArrowArray list<item: int64>>
- length: 4
- offset: 0
- null_count: 1
- buffers[2]:
  - validity <bool[1 b] 11010000>
  - data_offset <int32[20 b] 0 2 4 4 5>
- dictionary: NULL
- children[1]:
  'item': <ArrowArray int64>
    - length: 5
    - offset: 0
    - null_count: 1
    - buffers[2]:
      - validity <bool[1 b] 11011000>
      - data <int64[40 b] 0 1 0 3 4>
    - dictionary: NULL
    - children[0]:
```

### DIFF
--- a/python/src/nanoarrow/_repr_utils.py
+++ b/python/src/nanoarrow/_repr_utils.py
@@ -250,7 +250,7 @@ def device_repr(device):
     return "\n".join([title_line, device_type, device_id])
 
 
-def array_dump(array, indent=0, max_char_width=80):
+def array_inspect(array, indent=0, max_char_width=80):
     array_view = array.view()
 
     if max_char_width < 20:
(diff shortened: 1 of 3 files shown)


python/src/nanoarrow/array.py
```diff
@@ -490,3 +490,7 @@ def to_string(self, width_hint=80, items_hint=10) -> str:
 
     def __repr__(self) -> str:
         return self.to_string()
+
+    def inspect(self): # or dump?
```
paleolimbot, line 494: I quite like `inspect()`!

This could probably fairly easily be extended to multiple chunks (by printing out in a loop?)

paleolimbot (approved):
Thanks!

python/src/nanoarrow/array.py
jorisvandenbossche: > This could probably fairly easily be extended to multiple chunks (by printing out in a loop?)

That's indeed true, that shouldn't be that hard. We don't need it right now, so would prefer to do it later

python/src/nanoarrow/_repr_utils.py
```diff
@@ -248,3 +248,46 @@ def device_repr(device):
     device_type = f"- device_type: {device.device_type.name} <{device.device_type_id}>"
     device_id = f"- device_id: {device.device_id}"
     return "\n".join([title_line, device_type, device_id])
+
+
+def array_dump(array, indent=0, max_char_width=80):
```
paleolimbot, line 253:
```suggestion
def array_inspect(array, indent=0, max_char_width=80):
```

Issue #434, opened 2024-04-19
python: consider adding back a na.array(..) constructor


jorisvandenbossche:
While writing the example for our tutorial, the usage of `na.Array(..)` stood a bit out. There is of course nothing "wrong" with it, but it does feel inconsistent with how many other libraries provide their general purpose constructor (eg `pyarrow.array()`, `numpy.array()`, `torch.tensor()`, etc), and also somewhat internally with most or our other constructors (eg `na.c_array()` and not `na.CArray()`, and several other such cases)

So the suggestion to consider adding again a `nanoarrow.array(..) -> Array` constructor function (I think at some point in the Array PR it did exist?)

paleolimbot:
I forget exactly which review suggested I drop it (perhaps from @danepitkin ?), but I don't have strong feelings either way. Right now it gives a very bad error message (below), so I'm inclined to add it back (but perhaps keeping the documentation with the `Array` constructor?).

```python
import nanoarrow as na
na.array([1, 2, 3], na.int32())
#> TypeError: 'module' object is not callable
```

PR #441, opened 2024-04-22
feat(python): add back nanoarrow.array(..) constructor

jorisvandenbossche:
Closes https://github.com/apache/arrow-nanoarrow/issues/434

### DIFF
--- a/python/tests/test_array.py
+++ b/python/tests/test_array.py
@@ -39,7 +39,7 @@ def test_array_construct():
 
 
 def test_array_constructor():
-    array = na.Array([1, 2, 3], na.int32())
+    array = na.array([1, 2, 3], na.int32())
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
Another option would be to just link to `:class:Array` from `na.array()` for more details (but I don't mind either way and we can workshop that later when I go through the docs before release).

python/tests/test_array.py
```diff
@@ -38,6 +38,11 @@ def test_array_construct():
         iter(array)
 
 
+def test_array_constructor():
+    array = na.Array([1, 2, 3], na.int32())
```
paleolimbot, line 42:
```suggestion
    array = na.array([1, 2, 3], na.int32())
```

PR #442, opened 2024-04-22
ci(python): Set unique development version before building wheels

### LINKED ISSUE #437: python: Nightly wheel upload is failing
After #427 we now upload wheels! This job seems to now be failing:

https://github.com/apache/arrow-nanoarrow/actions/runs/8768760925/job/24063892961#step:5:78

The issue seems to be that the version is identical:

```
Uploading nanoarrow-0.5.0.dev0-pp39-pypy39_pp73-manylinux_2_5_i686.manylinux1_i686.manylinux_2_17_i686.manylinux2014_i686.whl - this version already exists
Uploading nanoarrow-0.5.0.dev0-pp39-pypy39_pp73-win_amd64.whl - this version already exists
Uploading nanoarrow-0.5.0.dev0.tar.gz - this version already exists
There was a problem uploading at least 1 package
```

cc @jorisvandenbossche @raulcd

paleolimbot:
Closes #437.

This PR uses the Arrow monorepo tag system, where the tag `apache-arrow-x.x.x.dev` signals where the changelog should start for the next version (and makes it easier to use various `git` tools to calculate the development version number for Python).

It also refactors what was previously `changelog.py` (and is now `release_tools.py`) since there's more than one thing it does now (and because I refuse to add any more untestable/untested bash to dev/release).


python/set_dev_version.py
```diff
@@ -0,0 +1,61 @@
...
+# under the License.
+
+"""Set development version from git
+
+This is used to set the Python package version before building
+nightly wheels, since at least some tools do not allow
+duplicate versions when uploading sdist/wheels. This requires
+that the repo has a full checkout with tags.
+
+The Python package version is hard-coded in a _static_version.py
```
jorisvandenbossche, line 25: Alternatively, we could also use more of the original https://github.com/jbweston/miniver code (the version setup we now have in Python is a slimmed down version of that, to remove the "get version from git" part)

paleolimbot: When I set this up I tried that, and while I forget the details, I remember that it was somewhat difficult in practice to satisfy all the constraints. I think one of the main pieces was ensuring that the wheels were built with the release version as part of the RC process (i.e., before the tag is actually created but after the versions have been bumped). I'm sure it's possible to do, but the static version string fits better with the other version bumpers (until now!).

Probably it would be better to move this to dev/release (with the rest of the version bumpers) so that this doesn't stick out quite as much.

.github/workflows/python-wheels.yaml
```diff
@@ -49,6 +51,11 @@ jobs:
       run: |
         pip install build twine
 
+    - name: Set nanoarrow Python dev version
+      # if: github.ref == 'refs/heads/main'
```
danepitkin, line 55: Is this commented code or is it a comment describing when this runs by default?

danepitkin (commented):
I'm not a python release expert so can't comment whether it looks good or not. I don't notice any issues, so seems fine to me. I'll defer to Joris otherwise.

.github/workflows/python-wheels.yaml
paleolimbot: Good catch! It was to make sure this worked in the PR; however, we don't want the dev version modified when we create the release branch or it will clobber the release version.

PR #444, opened 2024-04-26
refactor(python): Reorganize strategies for building arrays

paleolimbot:
This PR reorganizes the implementation for `na.c_array(<something>)`. The motivation here is that there are some Arrow types that aren't supported when creating Arrow arrays (e.g., decimal), and some container types that are not supported (e.g., data frame protocol). I don't think those should be supported right now but I would like there to be an obvious route to adding support. The `ArrayBuilder` class and the dispatch code that chooses the appropriate class/method is probably not perfect but does provide a more reasonable path to adding support for a new type (add a method to the `ArrayFromIterableBuilder` class) or new container (add a new `ArrayBuilder` subclass).

In order to avoid a circular import, this would all have had to be added to `c_lib.py`. Instead of doing a lazy import, I separated `c_lib.py` into `c_schema.py`, `c_buffer.py`, `c_array.py`, and `c_array_stream.py`. The tests for these functions had already been separated and I quite like the organization (reasonably obvious which file you might find the `c_array()` function in, for example).


python/src/nanoarrow/_lib.pyx
```diff
@@ -858,6 +858,24 @@ cdef class CSchemaView:
     def storage_type_id(self):
         return self._schema_view.storage_type
 
+    @property
+    def buffer_format(self):
+        if not self.extension_name and self._schema_view.type == self._schema_view.storage_type:
+            return self.storage_buffer_format
```
danepitkin, line 864: Can we add `return None` to be explicit? In Python, a function without `return` is commonly perceived similar to a void function in C/C++.

python/src/nanoarrow/_lib.pyx
```diff
@@ -858,6 +858,24 @@ cdef class CSchemaView:
...
+        cdef char out[128]
+        cdef int element_size_bits = 0
+        if self._schema_view.type == NANOARROW_TYPE_FIXED_SIZE_BINARY:
+            element_size_bits = self._schema_view.fixed_size * 8
+
+        try:
+            c_format_from_arrow_type(self._schema_view.type, element_size_bits, sizeof(out), out)
+            return out.decode()
+        except ValueError:
+            pass
```
danepitkin, line 877: same here (can we add `return None`)

danepitkin (commented):
I like the rearrangement! Is the code from c_lib.py purely a relocation? (If there are any changes made to the new files, it's hard to tell what they would be)

danepitkin (approved):
Thanks! Overall LGTM then

jorisvandenbossche:
> Instead of doing a lazy import, I separated `c_lib.py` into `c_schema.py`, `c_buffer.py`, `c_array.py`, and `c_array_stream.py`

FWIW it might have been easier to do this splitting in a separate PR first with just splitting/reorganising code, which would make the diff of the new stuff here easier to review (at this point it is probably harder to still do that, so let it be ;))

python/src/nanoarrow/c_array.py
```diff
@@ -0,0 +1,573 @@
...
+        )
+        self._append_using_buffer_builder(c_builder, wrapper)
+
+        _, null_count, validity = wrapper.finish()
+        if validity is not None:
+            c_builder.set_buffer(0, validity, move=True)
+
+        c_builder.set_null_count(null_count)
+
+    def _append_using_array(self, c_builder: CArrayBuilder, obj: Iterable) -> None:
```
jorisvandenbossche, line 499: What's the difference between using array vs using buffer builder? The latter is for the ones there is no matching buffer protocol type?

python/src/nanoarrow/c_array.py
```diff
@@ -0,0 +1,573 @@
...
+    def __init__(self, schema):
+        self._schema = c_schema(schema)
+        self._schema_view = c_schema_view(self._schema)
+
+    def build_c_array(self, obj):
+        builder = CArrayBuilder.allocate()
+        builder.init_from_schema(self._schema)
+        self.start_building(builder)
+        self.append(builder, obj)
+        return self.finish_building(builder)
```
jorisvandenbossche, line 326: Instead of having to pass the c_builder around to each method, could this be part of the object state?

python/tests/test_c_array.py
```diff
@@ -81,7 +69,7 @@ def _export_to_c(self, *args):
 
     # Make sure that this heuristic won't result in trying to import
     # something else that has an _export_to_c method
-    with pytest.raises(TypeError, match="Can't convert object of type DataType"):
+    with pytest.raises(TypeError, match="Can't resolve ArrayBuilder"):
```
jorisvandenbossche, line 72: Could we try to keep that original error message? (it seems clearer) Or is this only for the low-level c array, not for the user-facing array?

python/tests/test_c_schema_view.py
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from nanoarrow.c_lib import c_schema_view
+from nanoarrow.c_schema import c_schema_view
```
jorisvandenbossche, line 18: Nit: this is also available from the top-level namespace, so if changing this, maybe directly to that?

jorisvandenbossche:
The file re-org certainly makes sense! I have to say that I find the logic in the ArrayBuilder (sub)classes a bit harder to follow (e.g. compared to the current `_c_array_from_iterable` function)

python/src/nanoarrow/c_array.py
```diff
@@ -0,0 +1,573 @@
...
+    if hasattr(obj, "__arrow_c_array__"):
+        schema_capsule = None if schema is None else schema.__arrow_c_schema__()
+        return CArray._import_from_c_capsule(
+            *obj.__arrow_c_array__(requested_schema=schema_capsule)
+        )
+
+    # Try import of bare capsule
+    if _obj_is_capsule(obj, "arrow_array"):
+        if schema is None:
+            schema_capsule = CSchema.allocate()._capsule
```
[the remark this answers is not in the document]
paleolimbot: You can inspect the fields (`length`, `offset`, etc.) but that is about it. Its utility would mostly be in an interactive debug scenario where you want to actually look the contents of a capsule (there's really no other way to do that).

python/src/nanoarrow/c_array.py
paleolimbot: The `array` module only implements a subset of buffer protocol strings (basically just integers and doubles and floats). Float16, interval, and fixed-size binary all have valid buffer protocol format strings (but no array implementation), and boolean has no buffer protocol format string (but nanoarrow's Cython buffer builder can handle it anyway).

paleolimbot: It may be faster to wire up nanoarrow's (C) `ArrowArrayAppendInt`/`ArrowArrayAppendDouble` eventually...after dealing with `None` elements, the `_append_using_array()` approach is not particularly fast.

python/tests/test_c_array.py
paleolimbot: Good call! I made sure this wrapped at the top level:

```
import nanoarrow as na
na.Array(None)
#> TypeError: An error occurred whilst converting NoneType to nanoarrow.c_array_stream or nanoarrow.c_array: 
#>  Can't convert object of type NoneType to nanoarrow.c_array: 
#>  Can't resolve ArrayBuilder for object of type NoneType
```

python/src/nanoarrow/c_array.py
paleolimbot: I changed these all to `self._c_builder` as you suggested because it made way more sense with the current code structure. I think building recursive things will change this but that is a battle for another day 🙂

PR #446, opened 2024-04-29
ci(python): Add Python to release verification with TEST_WITH_MEMCHECK=1

paleolimbot:
This can be run locally with:

```shell
export NANOARROW_PLATFORM=ubuntu
# On M1 you might need:
# export NANOARROW_ARCH=arm64
docker compose run --rm -e TEST_WITH_MEMCHECK=1 -e TEST_DEFAULT=0 -e TEST_PYTHON=1 -e VERBOSE=1 verify
```

### DIFF
--- a/dev/release/verify-release-candidate.sh
+++ b/dev/release/verify-release-candidate.sh
@@ -354,6 +354,30 @@ test_python() {
   show_info "Testing wheel"
   python -m pytest -vv
 
+  if [ ${TEST_WITH_MEMCHECK} -gt 0 ]; then
+    show_info "Setup environment for Python/valgrind"
+
+    # To have more control over exactly which dependencies are available,
+    # set up a separate venv to run. Essentially, we check for leaks in
+    # all self-contained tests but not tests involving pyarrow or numpy.
+    VALGRIND_PYTHON_VENV="${NANOARROW_TMPDIR}/python/valgrind_venv"
+    python -m venv "${VALGRIND_PYTHON_VENV}"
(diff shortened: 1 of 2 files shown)


.github/workflows/verify.yaml
```diff
@@ -78,22 +78,17 @@ jobs:
...
         shell: bash
         run: |
-          curl https://dlcdn.apache.org/arrow/arrow-14.0.2/apache-arrow-14.0.2.tar.gz | \
-            tar -zxf -
-          mkdir arrow-build && cd arrow-build
-          cmake ../apache-arrow-14.0.2/cpp -DCMAKE_INSTALL_PREFIX=../arrow
-          cmake --build .
-          cmake --install . --prefix=../arrow ${{ matrix.config.extra_cmake_install }}
-          cd ..
+          src/ci/scripts/build-arrow-cpp-minimal.sh 15.0.2 arrow
```
raulcd, line 91: Is the update to 15.0.2 required for this PR or should this be done separately?

jorisvandenbossche:
Should we add Valgrind to normal CI instead of release verification? (is there something release specific about it?)

.github/workflows/verify.yaml
paleolimbot: It turned out there were a few changes required to make the job pass (the actual change that fixes it is adding `runner.arch` to the cache key). I don't mind doing this as a separate PR, although the Arrow version is not particularly important here (we support back to 9.0.0).

PR #447, opened 2024-04-30
chore(python): Remove C sources from wheels

paleolimbot:
Even though the binaries were quite small, nanoarrow's wheels were about 1 MB and the installed size was about 5 MB. This is because the flatcc headers and cython generated code was rather extensive and had ended up in the built wheel. After this change the wheels are ~400kb and installed size is 1.5 MB (for me).

The main change here is to move the directory where we vendor the C files from elsewhere in the repo outside `src/` such they are not automatically included in the wheel.

### DIFF
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-src/nanoarrow/vendor
+vendor/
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
> After this change the wheels are ~400kb and installed size is 1.5 MB (for me).

Nice improvement! ;)

PR #448, opened 2024-04-30
Add Meson build with Werror

WillAyd:
This should be a relatively easy way to enforce -Werror in CI; moved the Meson build from a scheduled job to be part of the normal CI runs.

Looks like there are still a few more upstream issues to be fixed before this is ready. Just posting for visibility


ci/scripts/build-arrow-cpp-minimal.sh
WillAyd, line 1: These changes are not permanent - just setting them up to see what we need for a clean build of Arrow.

src/nanoarrow/buffer_test.cc
```diff
@@ -141,9 +141,6 @@ TEST(BufferTest, BufferTestFill) {
   }
 
   ArrowBufferReset(&buffer);
-
```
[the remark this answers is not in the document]
paleolimbot: Would changing them to `ArrowBufferReserve()` work? (Also fine to remove if neither will compile...testing ENOMEM is hard and we mostly don't test it other places, too).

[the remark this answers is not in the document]
WillAyd: The BufferTestError checks for ENOMEM using `ArrowBufferResize` - do you think it would be better to add the `ArrowBufferReserve` test there for consistency or do you want to keep within this test?

meson.build
```diff
@@ -30,7 +30,7 @@ project(
 )
 
 if meson.get_compiler('cpp').get_id() == 'gcc' or meson.get_compiler('cpp').get_id() == 'clang'
-  add_project_arguments(['-fvisibility=hidden'], language: 'cpp')
+  add_project_arguments(['-fvisibility=hidden', '-Wno-unused-parameter'], language: 'cpp')
```
WillAyd, line 33: Since these aren't done upstream in Arrow kind of hard to keep things synced; probably just easiest to ignore until we potentially drop the Arrow testing dependency

ci/scripts/build-arrow-cpp-minimal.sh
```diff
@@ -45,10 +45,10 @@ ARROW_CPP_SCRATCH_DIR="arrow-cpp-build-${ARROW_CPP_VERSION}"
 mkdir "${ARROW_CPP_SCRATCH_DIR}"
 pushd "${ARROW_CPP_SCRATCH_DIR}"
 
-curl -L "https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/arrow-${ARROW_CPP_VERSION}/apache-arrow-${ARROW_CPP_VERSION}.tar.gz" | \
+curl -L "https://github.com/apache/arrow/archive/refs/heads/main.tar.gz" | \
   tar -zxf -
 mkdir build && cd build
-cmake ../apache-arrow-${ARROW_CPP_VERSION}/cpp \
+cmake ../arrow-main/cpp \
```
paleolimbot, line 51: Will the Arrow 17.0.0 release include the upstream changes required to make this work? We can't merge this particular change but we can update the version we use in CI once released.

WillAyd: Don't think so. I fixed a few but they keep popping up, especially after having added metal / ipc support. Can keep trying but I think unless Arrow comprehensively wanted to fix these warnings its going to be hard downstream to keep this clean

An alternate solution may be to treat Arrow as a system include in the build systems so that it doesn't show any warnings from that library. That might allow us to at least keep this clean within nanoarrow

paleolimbot: I'll pipe the Arrow C++ removal through in the next few weeks. We've let that dependency complicate quite a lot of things and I've mentioned its hypothetical removal quite a lot of times.

meson.build
```diff
@@ -161,7 +167,8 @@ if get_option('tests')
   # Similarly code coverage has a built in option users should use instead
   # https://mesonbuild.com/Unit-tests.html#coverage
 
-  arrow_dep = dependency('arrow')
+  # The system include suppresses compilation warnings from Arrow
+  arrow_dep = dependency('arrow', include_type: 'system')
```
WillAyd, line 171: Rather than chasing down the Arrow failures, I figured we can just ignore them

meson.build
```diff
@@ -35,6 +35,12 @@ project(
 #   add_project_arguments(['-fvisibility=hidden'], language: 'cpp')
 # endif
 
+cpp = meson.get_compiler('cpp')
+add_project_arguments(
+    cpp.get_supported_arguments(['-Wno-misleading-indentation']),
```
WillAyd, line 40: This warning was generated by the flatcc_generated header. There are a ton of these.

Perhaps there is a way to have it fixed on generation, but for now figure fine to ignore

WillAyd (commented):
This is going to require https://github.com/apache/arrow-nanoarrow/pull/595 going in first, but afterwards I think this might be ready

meson.build
paleolimbot: This had been fixed (https://github.com/apache/arrow-nanoarrow/pull/420) but we revendored flatcc and regenerated the header. I think ignoring for now is fine (although I'd like to redo the fix at some point since these errors show up when compiling Python and obscure other actual warnings!)

src/nanoarrow/common/array_test.cc
```diff
@@ -98,9 +98,11 @@ TEST(ArrayTest, ArrayTestAllocateChildren) {
   ArrowArrayRelease(&array);
 
   ASSERT_EQ(ArrowArrayInitFromType(&array, NANOARROW_TYPE_STRUCT), NANOARROW_OK);
+#if !defined(__SANITIZE_ADDRESS__)
```
WillAyd, line 101: I did also explicitly add in this code that takes out blocks which are known to be flagged by the sanitizers. This is a bit of "extra credit" but I think generally safer than ignoring a broad class of potential errors.

If we didn't want to use these macros I'm happy to revert. Otherwise, we could look at a suppression file instead of changes to the code:

https://clang.llvm.org/docs/AddressSanitizer.html#issue-suppression

meson.build
WillAyd: Do you know which version of flatcc was generated? Meson still uses flatcc as a subproject which is 0.6.1. However, the generated flatcc_generated.h file in the source tree uses some symbols that don't appear in that release. 

Should set the meson subproject up to use the same version as was vendored for CMake I think

src/nanoarrow/common/array_test.cc
WillAyd: Nevermind figured this out - all green

meson.build
paleolimbot: I think so! We use a git hash right now because @bkietz found a bug/suboptimal behaviour of every flatbuffer implementation *except* flatcc that forced a change in flatcc. I think using the vendored version is safer but you could probably also use the git hash.

WillAyd: Ended up changing the subproject to use the same git hash. Can relook at the design in a separate PR, but the Meson subproject approach does generally make things much simpler from a build configuration standpoint

src/nanoarrow/common/array_test.cc
paleolimbot: I think this is OK. We should really have a better system for testing out-of-memory anyway (but that is a not-today battle!)

paleolimbot (approved):

PR #450, opened 2024-05-01
feat(python): Implement bitmap unpacking

paleolimbot:
In prototyping a real-world use case, I remembered that unpacking bits is exceedingly difficult to get right if you need to support an arbitrary offset/length. The math for this is very fiddly and we spent a few rounds getting it right in the C function `ArrowBitsUnpackInt(8|32)`. This PR makes that available so that we can do things like (1) convert bool arrays to numpy and (2) convert null masks to something that somebody else can work with (e.g., a numpy mask).

This seems to be relatively performant (thanks to @WillAyd's work optimizing this!)

```python
import numpy as np
import nanoarrow as na
import pyarrow as pa

bool_np = np.random.random(int(1e6)) > 0.5
bool_na = na.Array(iter(bool_array), na.bool_())
bool_pa = pa.array(bool_np)

def to_numpy_na(x):
    x_view = na.c_array(x).view()
    out = np.empty(x_view.length, bool)
    x_view.buffer(1).unpack_bits_into(out, x_view.offset, x_view.length)
    return out

%timeit to_numpy_na(bool_na)
#> 162 µs ± 812 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit bool_pa.to_numpy(False)
#> 609 µs ± 833 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -1834,11 +1834,12 @@ cdef class CBufferView:
             PyBuffer_Release(&buffer)
             raise ValueError("Destination buffer has itemsize != 1")
 
-        if buffer.len < (dest_offset + length):
+        if dest_offset < 0 or buffer.len < (dest_offset + length):
             buffer_len = buffer.len
             PyBuffer_Release(&buffer)
             raise IndexError(
-                f"Can't unpack {length} elements into buffer of size {buffer_len}"
+                f"Can't unpack {length} elements into buffer of size {buffer_len} "
+                f"with dest_offset = {dest_offset}"
             )
 
         ArrowBitsUnpackInt8(
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -1815,7 +1815,7 @@ cdef class CBufferView:
         else:
             return self._iter_dispatch(offset, length)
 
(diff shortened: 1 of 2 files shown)


python/src/nanoarrow/_lib.pyx
```diff
@@ -1815,6 +1815,51 @@ cdef class CBufferView:
...
+                f"of buffer with {self.n_elements} elements"
+            )
+
+        cdef Py_buffer buffer
+        PyObject_GetBuffer(dest, &buffer, PyBUF_WRITABLE | PyBUF_ANY_CONTIGUOUS)
+        if buffer.itemsize != 1:
+            PyBuffer_Release(&buffer)
+            raise ValueError("Destination buffer has itemsize != 1")
+
+        if buffer.len < (dest_offset + length):
```
WillAyd, line 1837: Maybe worth also checking that `dest_offset` is non-negative? I think you'd get UB if a user decides to do something like `unpack_bits_into(..., len(dest) + 1, -1)` (not that they should...)

WillAyd (commented):
Really cool change - I can see there being a lot of uses for this

python/src/nanoarrow/_lib.pyx
paleolimbot: Good catch! Thanks for the review!

PR #451, opened 2024-05-02
fix(python): Make shallow CArray copies less shallow to accomodate moving children

paleolimbot:
This PR updates the logic that creates a "shallow copy" of an `ArrowArray`. Before, it simply made a shallow copy of the outer array, which works in most cases. However, the spec allows for "moving" child arrays, which means that the guarantee of a valid *outer* array does not guarantee anything about the validity of the `ArrowArray*` child pointers. It's difficult, but possible, to trigger this in Python (below); however, I think this sort of code is common for importers (because it allows the lifecycle of columns in an array to be independent).

```python
import nanoarrow as na
import pyarrow as pa

# Given some array
array = na.c_array_from_buffers(
    na.struct({"col1": na.int32()}),
    3,
    [None],
    children=[na.c_array([1, 2, 3], na.int32())]
)

user_array = na.Array(array)
user_array
#> nanoarrow.Array<int32>[3]
#> {'col1': 1}
#> {'col1': 2}
#> {'col1': 3}

# totally valid shallow copy of this array
schema_capsule, array_capsule = array.__arrow_c_array__()
array2 = na.c_array(array_capsule, schema_capsule)

# A consumer is technically allowed to move a child array
x = pa.Array._import_from_c(array2.child(0)._addr(), pa.int32())
del array
del x

# With the previous shallow copy implementation, this could segfault or fail
list(user_array.iter_py())
# Instead of a segfault I tend to get:
#> NanoarrowException: ArrowBasicArrayStreamValidate() failed (22): Expected int32 array buffer 1 to have size >= 12 bytes but found buffer with 0 bytes
```

After this PR the original array remains valid:

```python
list(user_array.iter_py())
#> [{'col1': 1}, {'col1': 2}, {'col1': 3}]
```

This does, however, have a non-trivial effect when making a shallow copy where a lot of children are involved:

```python
import nanoarrow as na
import pyarrow as pa

n_col = int(1e3)
n_items = int(1e5)

array_wide = na.c_array_from_buffers(
    na.struct({f"col{i}": na.int32() for i in range(n_col)}),
    n_items,
    [None],
    children=[na.c_array(range(n_items), na.int32())] * n_col
)

%timeit pa.array(array_wide)
#> Before this PR
#> 595 µs ± 5.72 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
#> After this Pr
#> 828 µs ± 2.83 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -206,7 +206,8 @@ cdef void c_array_shallow_copy(object base, const ArrowArray* src, ArrowArray* d
     for i in range(src.n_buffers):
         if src.buffers[i] != NULL:
             c_pyobject_buffer(base, src.buffers[i], 0, ArrowArrayBuffer(tmp, i))
-            tmp.buffers[i] = src.buffers[i]
+
+        tmp.buffers[i] = src.buffers[i]
 
     tmp.n_buffers = src.n_buffers
 
@@ -223,7 +224,7 @@ cdef void c_array_shallow_copy(object base, const ArrowArray* src, ArrowArray* d
         code = ArrowArrayAllocateDictionary(tmp)
         Error.raise_error_not_ok("ArrowArrayAllocateDictionary()", code)
 
-        c_array_shallow_copy(base, src.dictionary, dst.dictionary)
+        c_array_shallow_copy(base, src.dictionary, tmp.dictionary)
 
     # Move tmp into dst
     ArrowArrayMove(tmp, dst)
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -166,14 +166,19 @@ cdef object alloc_c_array_view(ArrowArrayView** c_array_view):
     return PyCapsule_New(c_array_view[0], 'nanoarrow_array_view', &pycapsule_array_view_deleter)
 
 
-cdef void arrow_array_release(ArrowArray* array) noexcept with gil:
-    Py_DECREF(<object>array.private_data)
-    array.private_data = NULL
-    array.release = NULL
+# Provide a way to validate that we release all references we create
+cdef int64_t pyobject_buffer_count = 0
+
+def get_pyobject_buffer_count():
+    global pyobject_buffer_count
+    return pyobject_buffer_count
 
 
 cdef void c_deallocate_pyobject_buffer(ArrowBufferAllocator* allocator, uint8_t* ptr, int64_t size) noexcept with gil:
     Py_DECREF(<object>allocator.private_data)
+
+    global pyobject_buffer_count
+    pyobject_buffer_count -= 1
 
 
 cdef void c_pyobject_buffer(object base, const void* buf, int64_t size_bytes, ArrowBuffer* out):
@@ -184,6 +189,9 @@ cdef void c_pyobject_buffer(object base, const void* buf, int64_t size_bytes, Ar
         <void*>base
     )
     Py_INCREF(base)
+
+    global pyobject_buffer_count
+    pyobject_buffer_count += 1
 
 
 cdef void c_array_shallow_copy(object base, const ArrowArray* src, ArrowArray* dst):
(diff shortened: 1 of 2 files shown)


python/src/nanoarrow/_lib.pyx
```diff
@@ -160,68 +160,93 @@ cdef void pycapsule_array_view_deleter(object array_capsule) noexcept:
...
-cdef void arrow_array_release(ArrowArray* array) noexcept with gil:
-    Py_DECREF(<object>array.private_data)
-    array.private_data = NULL
-    array.release = NULL
+# Provide a way to validate that we release all references we create
+cdef int64_t pyobject_buffer_count = 0
 
+def get_pyobject_buffer_count():
+    global pyobject_buffer_count
+    return pyobject_buffer_count
```
jorisvandenbossche, line 174: My first thought was whether we should only do this in some kind of debug build, but AFAIK this is not actually used in the code itself (se don't check the value), and only accessed in test code? (and so that should be safe and cheap)

python/src/nanoarrow/_lib.pyx
```diff
@@ -160,68 +160,93 @@ cdef void pycapsule_array_view_deleter(object array_capsule) noexcept:
...
 
-cdef void c_device_array_shallow_copy(object base, const ArrowDeviceArray* c_array,
-                                      ArrowDeviceArray* c_array_out) noexcept:
-    # shallow copy
-    memcpy(c_array_out, c_array, sizeof(ArrowDeviceArray))
-    c_array_out.array.release = NULL
-    c_array_out.array.private_data = NULL
+    # Copy data for this array, adding a reference for each buffer
+    # This allows us to use nanoarrow's ArrowArray implementation without
+    # writing our own release callbacks/private_data.
```
jorisvandenbossche, line 209: I don't understand this comment, because we _are_ in nanoarrow?

python/src/nanoarrow/_lib.pyx
```diff
@@ -160,68 +160,93 @@ cdef void pycapsule_array_view_deleter(object array_capsule) noexcept:
...
+    tmp.offset = src.offset
+    tmp.null_count = src.null_count
 
-    # track original base
-    c_array_out.array.private_data = <void*>base
-    Py_INCREF(base)
-    c_array_out.array.release = arrow_array_release
+    for i in range(src.n_buffers):
+        if src.buffers[i] != NULL:
+            c_pyobject_buffer(base, src.buffers[i], 0, ArrowArrayBuffer(tmp, i))
```
jorisvandenbossche, line 216: Why the 0 size?

python/src/nanoarrow/_lib.pyx
```diff
@@ -160,68 +160,93 @@ cdef void pycapsule_array_view_deleter(object array_capsule) noexcept:
...
-    depended on by some other Python object, we implement a shallow copy
-    whose constructor calls Py_INCREF() on a Python object responsible
-    for the ArrowArray's lifecycle and whose deleter calls Py_DECREF() on
-    the same object.
-    """
-    cdef ArrowArray* c_array_out
-    array_capsule = alloc_c_array(&c_array_out)
-    c_array_shallow_copy(base, c_array, c_array_out)
-    return array_capsule
+cdef void c_array_shallow_copy(object base, const ArrowArray* src, ArrowArray* dst):
```
jorisvandenbossche, line 197: Can you add a docstring that says explains something about "how shallow" this function?

jorisvandenbossche (commented):
Just some small comments asking for clarifications!

python/src/nanoarrow/_lib.pyx
paleolimbot: I checked the generated C code and it looks like it is "just" incrementing and decrementing `static int64_t __pyx_v_9nanoarrow_4_lib_pyobject_buffer_count;` (which I don't believe is prohibitively expensive). I have a similar reference counting scheme in the R package and it has been mission critical in debugging!

python/src/nanoarrow/_lib.pyx
paleolimbot: I added a comment...it's because we don't know how long the buffer is (although it wouldn't necessarily matter since the only function that will ever see it is our own deallocator).

python/src/nanoarrow/_lib.pyx
paleolimbot: I added a nice long one!

python/src/nanoarrow/_lib.pyx
paleolimbot: I added a comment here to specifically refer to the `ArrowArray` release callback provided by `ArrowArrayInitFromType()` (which walks over a private array of `ArrowBuffer` and calls `ArrowBufferReset()`).

python/src/nanoarrow/_lib.pyx
jorisvandenbossche: Thanks!

PR #454, opened 2024-05-04
feat(python): Add visitor pattern + builders for column sequences

paleolimbot:
Assembling columns from chunked things is rather difficult to do and is a valid thing that somebody might want to assemble from Arrow data. This PR adds a "visitor" pattern that can be extended to build "column"s, which are currently just `list()`s. Before trimming down this PR to a managable set of changes, I also implemented the visitor that concatenates data buffers for single data buffer types ( https://gist.github.com/paleolimbot/17263e38b5d97c770e44d33b11181eaf ), which will be needed for `to_columns()` to be used in any kind of serious way.

To support the "visitor" pattern, I moved some of the `PyIterator`-specific pieces into the `PyIterator` so that the visitor can re-use the relevant pieces of `ArrayViewBaseIterator`. This pattern also solves one of the problems I had when attempting a "repr" iterator, which is that I was trying to build something rather than iterate over it.

```python
import nanoarrow as na
import pandas as pd
from nanoarrow import visitor

url = "https://github.com/apache/arrow-experiments/raw/main/data/arrow-commits/arrow-commits.arrows"
array = na.ArrayStream.from_url(url).read_all()

# to_columns() doesn't (and won't) produce anything numpy or pandas-related
names, columns = visitor.to_columns(array)

# ..but lets data frames be built rather compactly
pd.DataFrame({k: v for k, v in zip(names, columns)})
```

### DIFF
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -1,0 +1,55 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from nanoarrow._lib import CArray, CArrayView, CSchema, CSchemaView
+from nanoarrow.c_array_stream import c_array_stream
+from nanoarrow.iterator import ArrayViewBaseIterator
+
+
+class ArrayStreamVisitor:
+    def __init__(self, iterator_cls=ArrayViewBaseIterator) -> None:
+        self._iterator_cls = iterator_cls
+
+    def visit(self, obj, schema=None):
+        with c_array_stream(obj, schema=schema) as stream:
+            iterator = self._iterator_cls(stream._get_cached_schema())
+            state = self.visit_schema(iterator._schema, iterator._schema_view)
+
+            iterator_set_array = iterator._set_array
+            visit_array = self.visit_array
+            array_view = iterator._array_view
+
+            for array in stream:
+                iterator_set_array(array)
+                visit_array(array, array_view, iterator, state)
+
+        return self.finish(state)
+
+    def visit_schema(self, schema: CSchema, schema_view: CSchemaView):
+        return None
+
+    def visit_array(
+        self,
+        array: CArray,
+        array_view: CArrayView,
+        iterator: ArrayViewBaseIterator,
+        state,
+    ):
+        pass
+
+    def finish(self, state):
+        return state
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -183,3 +183,47 @@ class UnpackedBitmapConcatenator(BufferConcatenator):
         writable_buffer.advance(length)
 
         return out_start + length, buffer_index, writable_buffer
+
+    def finish(self, state: Any) -> CBuffer:
+        return memoryview(super().finish(state)).cast("?")
+
+
+class ValidityBufferConcatenator(UnpackedBitmapConcatenator):
+
+    def __init__(self, schema, *, _array_view=None):
+        super().__init__(schema, buffer_index=0, _array_view=_array_view)
+
+    def begin(self, total_elements: Union[int, None] = None):
+        return 0, total_elements, None
+
+    def visit_chunk_view(self, array_view: CArrayView, state: Any) -> Any:
+        total_count, total_elements, super_state = state
+
+        length = array_view.length
+
+        if self._contains_nulls():
+            if super_state is None:
+                super_state = super().begin(total_elements)
+                super_state = self._fill_valid(super_state, total_count)
+
+            super_state = super().visit_chunk_view(array_view, super_state)
+        elif super_state is not None:
+            super_state = self._fill_valid(super_state, length)
+
+        total_count += length
+        return total_count, total_elements, super_state
+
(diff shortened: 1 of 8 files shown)


python/src/nanoarrow/visitor.py
```diff
@@ -0,0 +1,179 @@
...
+
+from typing import Any, List, Sequence, Tuple, Union
+
+from nanoarrow._lib import CArrayView
+from nanoarrow.c_array_stream import c_array_stream
+from nanoarrow.iterator import ArrayViewBaseIterator, PyIterator
+from nanoarrow.schema import Type
+
+
+def to_pylist(obj, schema=None) -> List:
```
danepitkin, line 26: I think my personal preference would be to have `to_pylist` and `to_columns` be class APIs instead of helper functions. e.g. 

```
    >>> import nanoarrow as na
    >>> array = na.c_array([1, 2, 3], na.int32())
    >>> array.to_pylist()
```

You could put these two apis in an abstract base class and add the base class to other classes. WDYT?

python/src/nanoarrow/visitor.py
```diff
@@ -0,0 +1,179 @@
...
+                visit_chunk_view(array_view)
+
+        return visitor.finish()
+
+    def begin(self, total_elements: Union[int, None] = None):
+        """Called after the schema has been resolved but before any
+        chunks have been visited. If the total number of elements
+        (i.e., the sum of all chunk lengths) is known, it is provided here.
+        """
+        pass
```
danepitkin, line 116: Maybe raise `NotImplementedError()` here? Alternatively, you could also look into https://docs.python.org/3/library/abc.html and abstract methods.

python/src/nanoarrow/visitor.py
```diff
@@ -0,0 +1,179 @@
...
+        """Called exactly one for each chunk seen."""
+        pass
+
+    def finish(self) -> Any:
+        """Called exactly once after all chunks have been visited."""
+        return None
+
+
+class ListBuilder(ArrayStreamVisitor):
+    def __init__(self, schema, *, iterator_cls=PyIterator, _array_view=None):
```
danepitkin, line 128: nit: `_array_view` -> `array_view` in ListBuilder/ColumnsBuilder

python/tests/test_visitor.py
```diff
@@ -0,0 +1,46 @@
...
+# specific language governing permissions and limitations
+# under the License.
+import pytest
+
+import nanoarrow as na
+from nanoarrow import visitor
+
+
+def test_to_pylist():
+    assert visitor.to_pylist([1, 2, 3], na.int32()) == [1, 2, 3]
```
danepitkin, line 24: This feels like an odd test because the input and output are technically both pylists. Would it be better to build an array and use that?

danepitkin (commented):
Nice work!

python/tests/test_visitor.py
paleolimbot: Good call!

python/src/nanoarrow/visitor.py
paleolimbot: Good call! I'd used `_array_view` because it's sort of internal; however, the whole API is currently internal. If/when it's made public there should perhaps be a better system for instantiating/efficiently reusing child array views that does not stick out like this does.

python/src/nanoarrow/visitor.py
paleolimbot: It is rare, but valid, that an implementation of a visitor does not need to do anything at one or more of `begin`, `visit`, and `finish()` (e.g., print chunks, time the consumption process). If/when this is public it's definitely worth revisiting how best to communicate this (e.g., maybe force implementations to explicitly do nothing for clarity).

danepitkin (approved):
LGTM! Nice updates!

python/src/nanoarrow/visitor.py
paleolimbot: Probably all the visitors and iterators would be nice in an abstract base class shared between the `Array` and `ArrayStream`! I added the user-facing bit here to ensure that `na.Array(...).to_pylist()` is what users type and if there comes a time we have to start adding documentation in multiple places (very possible with `to_columns()`) we can revisit the base class.

python/src/nanoarrow/visitor.py
```diff
@@ -0,0 +1,179 @@
...
+from nanoarrow.c_array_stream import c_array_stream
+from nanoarrow.iterator import ArrayViewBaseIterator, PyIterator
+from nanoarrow.schema import Type
+
+
+def to_pylist(obj, schema=None) -> List:
+    """Convert ``obj`` to a ``list()` of Python objects
+
+    Computes an identical value to ``list(iterator.iter_py())`` but is several
+    times faster.
```
jorisvandenbossche, line 30: What is actually the reason that it is multiple times faster? As I would expect it uses the same iteration code under the hood?

paleolimbot: I have no idea! "Several times" is probably too strong here. The iteration code is slightly different, though (`out_list.extend(array_iter) in a loop` vs `list(yield from array_iter)`.

```python
import nanoarrow as na
import numpy as np

big = na.Array(np.random.random(int(1e6)))
%timeit big.to_pylist()
#> 19.8 ms ± 586 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit list(big.iter_py())
#> 50.6 ms ± 325 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

jorisvandenbossche: Looking at the code, the main difference seems to be that the faster one reuses the `array_view` argument to the iterator constructor? (although that shouldn't make any difference for a case with one chunk like the above?)

paleolimbot: I believe that they both reuse the array_view...it seems to be `yield from` but I'm not an expert.

```python
import numpy as np

list_of_lists = []
for i in range(1000):
    list_of_lists.append(list(np.random.random(1000)))

def iter_all():
    for item in list_of_lists:
        yield from item
    
def extend_all():
    out = []
    for item in list_of_lists:
        out.extend(item)
    return out
        
assert list(iter_all()) == extend_all()

%timeit list(iter_all())
#> 34.5 ms ± 522 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit extend_all()
#> 4.35 ms ± 517 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

PR #455, opened 2024-05-06
feat(python): Add copy_into() to CBufferView

paleolimbot:
This is the non-bitmap equivalent of #450, useful for the same purpose (concatenating one big data buffer from chunks).


danepitkin (approved):
LGTM!

PR #456, opened 2024-05-06
feat(python): Add `Array.from_chunks()` constructor

paleolimbot:
This PR adds a public route to construct chunked arrays (and makes the other constructors safer to account for the fact that they are now user-facing). I use this quite a lot interactively to test that things work in the chunked case, and for nanoarrow to be useful in a "I can help you export things" kind of way, it needs to be able to do this (because string arrays with more than 2 GB of text or binary are not uncommon).

The main safety consideration here is ensuring that all chunks have a schema of the same type, so I had to add a function to check for that (and ensure it was being checked).

```python
import nanoarrow as na
import numpy as np

na.Array.from_chunks([[1, 2, 3], [4, 5, 6]], na.int32())
na.Array.from_chunks((np.random.random(int(1e3)) for _ in range(int(1e3))))
```

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -769,7 +769,7 @@ cdef class CSchema:
     def __repr__(self):
         return _repr_utils.schema_repr(self)
 
-    def type_equals(self, CSchema other):
+    def type_equals(self, CSchema other, check_nullability=False):
         self._assert_valid()
 
         if self._ptr == other._ptr:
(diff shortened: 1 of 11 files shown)


python/src/nanoarrow/_lib.pyx
```diff
@@ -2700,20 +2729,36 @@ cdef class CArrayStream:
...
         cdef ArrowArray tmp
         cdef CArray array
         for i in range(len(arrays)):
             array = arrays[i]
+
+            if validate and not validate_schema.type_equals(array.schema):
+                raise ValueError(
+                    f"Expected schema {validate_schema._to_string()} "
+                    f"but got {array.schema._to_string()}"
+                )
```
danepitkin, line 2760: It might be nice to put this in a helper function since it's duplicated below.

danepitkin (commented):
Just a few small comments! Nice work.

python/src/nanoarrow/_lib.pyx
```diff
@@ -2921,6 +2972,32 @@ cdef class CMaterializedArrayStream:
...
 
+    @staticmethod
+    def from_c_arrays(arrays, CSchema schema, bint validate=True):
+        cdef CMaterializedArrayStream out = CMaterializedArrayStream()
+
+        cdef CArray array
+        for item in arrays:
+            array = item
+
+            if array._ptr.length == 0:
```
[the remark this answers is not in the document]
paleolimbot: This was a bit of a rabbit hole but should have been added a long time ago!

python/src/nanoarrow/_lib.pyx
paleolimbot: Done! (`CSchema.assert_type_equal()`)

python/src/nanoarrow/array.py
```diff
@@ -161,6 +163,27 @@ def __init__(self, obj, schema=None, device=None) -> None:
         with c_array_stream(obj, schema=schema) as stream:
             self._data = CMaterializedArrayStream.from_c_array_stream(stream)
 
+    @staticmethod
+    def from_chunks(obj, schema=None, validate=True):
```
jorisvandenbossche, line 167: Can you add a docstring here?

jorisvandenbossche (commented):
Looks good!

python/src/nanoarrow/_lib.pyx
```diff
@@ -769,6 +769,35 @@ cdef class CSchema:
...
+
+        if self._ptr == other._ptr:
+            return True
+
+        if self.format != other.format:
+            return False
+
+        if self.flags != other.flags:
+            return False
+
```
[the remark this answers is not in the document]
paleolimbot: It's the appropriate check for the purposes of ensuring an iterable of arrays can be safely interpreted according to a given schema but it's a good point that nullability is not strictly part of the "type". I'll make that part optional (although this is internal for now).

danepitkin (approved):
LGTM!

PR #457, opened 2024-05-07
feat(python): Ensure that buffer produced by `CBufferView.unpack_bits()` has a boolean type

paleolimbot:
This is small change to ensure that `np.array(some_buffer.unpack_bits())` "just works" without nanoarrow having to know about numpy dtypes. Basically we just need to ensure that we can create/export a buffer with a `"?"` format string.

```python
import nanoarrow as na
import numpy as np

bool_array = na.Array([True, True, True, False, False, True], na.bool_())
np.array(bool_array.buffer(1).unpack_bits(0, len(bool_array)))
#> array([ True,  True,  True, False, False,  True])
```


danepitkin (approved):
LGTM

PR #458, opened 2024-05-07
feat(python): Unify printing of type information across classes

paleolimbot:
This PR unifies how types are printed, which was different among the `Array`, `Scalar`, `ArrayStream`, and `Schema`. These now all use the same function such that type information is communicated in a consistent way for all user-facing classes.

The general approach is to print all possible information on the first line. Most types have descriptions like `int32`, so this is a good fit and there's no need for multiple lines. This makes a list of schemas (e.g., `schema.fields`) have a nice repr, too. The Schema repr additionally prints out metadata if present (this is skipped by the other classes).

Some examples:

```python
import nanoarrow as na

# Schema with no parameters
na.int32()
#> <Schema> int32

# Type parameters are already part of the schema's to_string
na.fixed_size_binary(123)
#> <Schema> fixed_size_binary(123)

# non-nullable schema
na.int32(nullable=False)
#> <Schema> non-nullable int32

# named schema
na.Schema(na.int32(), name="some_col")
#> <Schema> 'some_col': int32

# ordered dictionary
na.dictionary(na.int32(), na.string(), True)
#> <Schema> ordered dictionary(int32)<string>

# Schema with metadata
na.Schema(na.int32(), metadata={"some key": "some value"})
#> <Schema> int32
#> - metadata: {b'some key': b'some value'}

# fields are already a part of the schema's to_string
struct = na.struct({"some col": na.int32(), "some other col": na.string()})
struct
#> <Schema> struct<some col: int32, some other col: string>

# ...and you can inspect them more closely with .fields
struct.fields
#> [<Schema> 'some col': int32, <Schema> 'some other col': string]
```

### DIFF
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -1721,8 +1721,7 @@ cdef class SchemaMetadata:
         return list(self)
 
     def values(self):
-        for _, value in self.items():
-            yield value
+        return [value for _, value in self.items()]
 
     def items(self):
         cdef ArrowStringView key
--- a/python/src/nanoarrow/_lib.pyx
+++ b/python/src/nanoarrow/_lib.pyx
@@ -1717,6 +1717,13 @@ cdef class SchemaMetadata:
         for key, _ in self.items():
             yield key
 
+    def keys(self):
+        return list(self)
+
+    def values(self):
+        for _, value in self.items():
+            yield value
+
     def items(self):
         cdef ArrowStringView key
         cdef ArrowStringView value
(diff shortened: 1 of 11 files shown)


python/src/nanoarrow/_lib.pyx
```diff
@@ -1717,6 +1717,13 @@ cdef class SchemaMetadata:
         for key, _ in self.items():
             yield key
 
+    def keys(self):
+        return list(self)
+
+    def values(self):
+        for _, value in self.items():
+            yield value
```
danepitkin, line 1725: Should we make `keys` and `values` return the same container type? Right now one is a list and the other is a generator. Maybe use lists for both?

danepitkin (approved):
Overall LGTM! Just added one small comment

python/src/nanoarrow/schema.py
```diff
@@ -481,7 +499,7 @@ def null(nullable: bool = True) -> Schema:
 
     >>> import nanoarrow as na
     >>> na.null()
-    Schema(NULL)
+    <Schema> na
```
jorisvandenbossche, line 502: This "na" is defined by the nanoarrow C sources? Because I think using "null" is 1) much clearer for the user, and 2) more in alignment with the actual spec

paleolimbot: Yes, it comes from `ArrowTypeString()`, which draws from `NANOARROW_TYPE_NA`, which came from `ARROW_TYPE_NA`. It should be updated since it now creeps into a number of user-facing places (in R too).

PR #459, opened 2024-05-08
fix!: Change `ArrowBufferResize()` and `ArrowBitmapResize()` to update `size_bytes`

paleolimbot:
Currently, `ArrowBufferResize()` and `ArrowBitmapResize()` update the *capacity* of the underlying buffer but do not update the *size*. This was documented and is consistent with Arrow C++'s `BufferBuilder` but inconsistent with `std::vector`, Arrow C++'s `PoolBuffer`, and user expectations given the name of the function.

After this PR, `ArrowBufferResize()` and `ArrowBitmapResize()` update the `size_bytes` of the underlying buffer.

This is a breaking change.

### DIFF
--- a/src/nanoarrow/buffer_inline.h
+++ b/src/nanoarrow/buffer_inline.h
@@ -479,32 +479,36 @@ static inline void ArrowBitmapMove(struct ArrowBitmap* src, struct ArrowBitmap* 
 static inline ArrowErrorCode ArrowBitmapReserve(struct ArrowBitmap* bitmap,
                                                 int64_t additional_size_bits) {
   int64_t min_capacity_bits = bitmap->size_bits + additional_size_bits;
-  if (min_capacity_bits <= (bitmap->buffer.capacity_bytes * 8)) {
+  int64_t min_capacity_bytes = _ArrowBytesForBits(min_capacity_bits);
+
+  if (min_capacity_bytes <= bitmap->buffer.capacity_bytes) {
     return NANOARROW_OK;
   }
 
-  NANOARROW_RETURN_NOT_OK(
-      ArrowBufferReserve(&bitmap->buffer, _ArrowBytesForBits(additional_size_bits)));
+  int64_t additional_capacity_bytes = min_capacity_bytes - bitmap->buffer.size_bytes;
+  NANOARROW_RETURN_NOT_OK(ArrowBufferReserve(&bitmap->buffer, additional_capacity_bytes));
 
-  bitmap->buffer.data[bitmap->buffer.capacity_bytes - 1] = 0;
+  // Zero out the last byte for deterministic output for the common case
+  // of reserving a known remaining size.
+  if (bitmap->buffer.capacity_bytes > 0) {
+    bitmap->buffer.data[bitmap->buffer.capacity_bytes - 1] = 0;
+  }
+
   return NANOARROW_OK;
 }
 
 static inline ArrowErrorCode ArrowBitmapResize(struct ArrowBitmap* bitmap,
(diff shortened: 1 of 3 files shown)


src/nanoarrow/buffer_inline.h
```diff
@@ -468,32 +479,36 @@ static inline void ArrowBitmapMove(struct ArrowBitmap* src, struct ArrowBitmap*
...
   }
 
-  NANOARROW_RETURN_NOT_OK(
-      ArrowBufferReserve(&bitmap->buffer, _ArrowBytesForBits(additional_size_bits)));
+  int64_t additional_capacity_bytes = min_capacity_bytes - bitmap->buffer.size_bytes;
+  NANOARROW_RETURN_NOT_OK(ArrowBufferReserve(&bitmap->buffer, additional_capacity_bytes));
+
+  // Zero out the last byte for deterministic output for the common case
+  // of reserving a known remaining size.
+  if (bitmap->buffer.capacity_bytes > 0) {
```
zeroshade, line 493: we should probably change this to `if (additional_capacity_bytes > 0)` so that we don't accidentally zero out a byte that had data in it if the additional bits didn't end up adding a new byte

zeroshade: should probably add a test for that exact case, where we have an existing bitmap and then we reserve an additional N bits but not enough to force any additional bytes, and ensure we don't end up zero-ing out valid data

paleolimbot: Ok! I think I've done this. The logic above these lines should have ensured an early return if no extra bytes were needed (so I added a dcheck here + a test to ensure that this is the case).

src/nanoarrow/buffer_inline.h
```diff
@@ -468,32 +479,38 @@ static inline void ArrowBitmapMove(struct ArrowBitmap* src, struct ArrowBitmap*
...
+  if (min_capacity_bytes <= current_capacity_bytes) {
     return NANOARROW_OK;
   }
 
-  NANOARROW_RETURN_NOT_OK(
-      ArrowBufferReserve(&bitmap->buffer, _ArrowBytesForBits(additional_size_bits)));
+  int64_t additional_capacity_bytes = min_capacity_bytes - current_size_bytes;
+  NANOARROW_RETURN_NOT_OK(ArrowBufferReserve(&bitmap->buffer, additional_capacity_bytes));
 
+  // Zero out the last byte for deterministic output in the common case
```
danepitkin, line 493: Why zero out the last byte only and not all unused bits/bytes?

paleolimbot: When I first wrote this I copied this behaviour from Arrow C++:

https://github.com/apache/arrow/blob/7aea8bf7a65d679bd71d973b358f997eb3b6c6af/cpp/src/arrow/buffer.cc#L183-L189

...although there is also precedent in Arrow C++ to zero out any additional allocated memory:

https://github.com/apache/arrow/blob/7aea8bf7a65d679bd71d973b358f997eb3b6c6af/cpp/src/arrow/buffer_builder.h#L418-L423

In general nanoarrow's "reserve" step will never initialize memory since it assumes that the caller is about to do that. For a bitmap the caller might want to initialize to `0xff` or `0x00` depending on what they're about to do...the final byte is just to make it more likely that somebody building a bitmap of a pre-known length will get deterministic output (without any performance issue).

danepitkin: That's fair! Thanks for the explanation.

danepitkin (approved):

PR #463, opened 2024-05-13
[MINOR][Documentation] Fix pip install with test dependencies

AlenkaF:
Adding missing double quotes so the install with tests works.


paleolimbot (approved):
Thanks

PR #464, opened 2024-05-13
feat(python): Add column-wise buffer builder

paleolimbot:
This PR implements building columns buffer-wise for the types where this makes sense. It also implements a few other changes:

- `item_size` was renamed to `itemsize` to match the memoryview property name
- The visitor methods are now the `ArrayViewVisitable` mixin such that they are available in both the `Array` and `ArrayView` without duplicating documentation.

Functionally this means that the `Array` and `ArrayStream` now have `to_column()` and `to_column_list()` methods that do something that more closely matches what somebody would expect.

A quick demo:

```python
import nanoarrow as na
import pyarrow as pa

batch = pa.record_batch({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
batch_with_nulls = pa.record_batch({"col1": [1, None, 3], "col2": ["a", "b", None]})

# Either builds a buffer or a list depending on column types
na.Array(batch).to_columns_pysequence()
#> (['col1', 'col2'],
#>  [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']])

# One can inject a null handler (a few experimental ones provided)
na.Array(batch_with_nulls).to_columns_pysequence(handle_nulls=na.nulls_as_sentinel())
#> (['col1', 'col2'], [array([ 1., nan,  3.]), ['a', 'b', None]])

# ...by default you have to choose how to do this or we error
na.Array(batch_with_nulls).to_columns_pysequence()
#> ValueError: Null present with null_handler=nulls_forbid()
```

This will basically get you data frame conversion:

```python
import nanoarrow as na
import pandas as pd

url = "https://github.com/apache/arrow-experiments/raw/main/data/arrow-commits/arrow-commits.arrows"
names, data = na.ArrayStream.from_url(url).to_columns_pysequence(handle_nulls=na.nulls_as_sentinel())
pd.DataFrame({k: v for k, v in zip(names, data)})
#>                                          commit                      time  \
#> 0      49cdb0fe4e98fda19031c864a18e6156c6edbf3c 2024-03-07 02:00:52+00:00   
#> 1      1d966e98e41ce817d1f8c5159c0b9caa4de75816 2024-03-06 21:51:34+00:00   
#> 2      96f26a89bd73997f7532643cdb27d04b70971530 2024-03-06 20:29:15+00:00   
#> 3      ee1a8c39a55f3543a82fed900dadca791f6e9f88 2024-03-06 07:46:45+00:00   
#> 4      3d467ac7bfae03cf2db09807054c5672e1959aec 2024-03-05 16:13:32+00:00   
#> ...                                         ...                       ...   
#> 15482  23c4b08d154f8079806a1f0258d7e4af17bdf5fd 2016-02-17 12:39:03+00:00   
#> 15483  16e44e3d456219c48595142d0a6814c9c950d30c 2016-02-17 12:38:39+00:00   
#> 15484  fa5f0299f046c46e1b2f671e5e3b4f1956522711 2016-02-17 12:38:39+00:00   
#> 15485  cbc56bf8ac423c585c782d5eda5c517ea8df8e3c 2016-02-17 12:38:39+00:00   
#> 15486  d5aa7c46692474376a3c31704cfc4783c86338f2 2016-02-05 20:08:35+00:00   
#> 
#>        files  merge                                            message  
#> 0          2  False  GH-40370: [C++] Define ARROW_FORCE_INLINE for ...  
#> 1          1  False     GH-40386: [Python] Fix except clauses (#40387)  
#> 2          1  False  GH-40227: [R] ensure executable files in `crea...  
#> 3          1  False  GH-40366: [C++] Remove const qualifier from Bu...  
#> 4          1  False  GH-20127: [Python][CI] Remove legacy hdfs test...  
#> ...      ...    ...                                                ...  
#> 15482     73  False  ARROW-4: This provides an partial C++11 implem...  
#> 15483      8  False  ARROW-3: This patch includes a WIP draft speci...  
#> 15484    124  False                 ARROW-1: Initial Arrow Code Commit  
#> 15485      2  False             Update readme and add license in root.  
#> 15486      1  False                                     Initial Commit  
#> 
#> [15487 rows x 5 columns]
```

### DIFF
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -135,8 +135,8 @@ def nulls_as_sentinel(sentinel=None):
 
     A null handler that assigns a sentinel to null values. This is
     done using numpy using the expression ``data[~is_valid] = sentinel``.
-    The default sentinel value of ``None`` will result in ``nan``
-    assigned to null values in numeric and boolean outputs. This
+    The default sentinel value of ``None`` will result in float output and ``nan``
+    assigned to null values for numeric and boolean inputs. This
     corresponds to numpy's handling of ``None`` in ``np.result_type()``
     and ``result[~is_valid] = None``.
 
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -66,12 +66,12 @@ class ArrayViewVisitable:
 
         >>> import nanoarrow as na
         >>> import pyarrow as pa
-        >>> batch = pa.record_batch([pa.array([1, 2, 3])], names=["col1"])
+        >>> batch = pa.record_batch({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
         >>> names, columns = na.Array(batch).to_column_list()
         >>> names
-        ['col1']
+        ['col1', 'col2']
         >>> columns
-        [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3)]
+        [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']]
         """
         return ColumnsBuilder.visit(self, handle_nulls=handle_nulls)
 
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -59,7 +59,8 @@ class ArrayViewVisitable:
         handle_nulls : callable
             A function returning a sequence based on a validity bytemap and a
             contiguous buffer of values (e.g., the callable returned by
-            :meth:`nulls_as_sentinel`).
+            :func:`nulls_as_sentinel`, :func:`nulls_forbid`, or
+            :func:`nulls_separate`). The default value is :func:`nulls_forbid`.
 
         Examples
         --------
@@ -90,7 +91,8 @@ class ArrayViewVisitable:
         handle_nulls : callable
             A function returning a sequence based on a validity bytemap and a
             contiguous buffer of values (e.g., the callable returned by
-            :meth:`nulls_as_sentinel`).
+            :func:`nulls_as_sentinel`, :func:`nulls_forbid`, or
+            :func:`nulls_separate`). The default value is :func:`nulls_forbid`.
 
         Examples
         --------
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -47,11 +47,11 @@ class ArrayViewVisitable:
         """
         return ListConverter.visit(self)
 
-    def convert_columns(self, handle_nulls=None) -> Tuple[List[str], List[Sequence]]:
+    def convert_columns(self, *, handle_nulls=None) -> Tuple[List[str], List[Sequence]]:
         """Convert to a ``list`` of contiguous sequences
 
-        Converts a stream of struct arrays into its column-wise representation
-        according to :meth:`convert`.
+        Experimentally converts a stream of struct arrays into a list of contiguous
+        sequences using the same logic as :meth:`convert`.
 
         Paramters
         ---------
@@ -75,15 +75,17 @@ class ArrayViewVisitable:
         """
         return ColumnListConverter.visit(self, handle_nulls=handle_nulls)
 
-    def convert(self, handle_nulls=None) -> Sequence:
+    def convert(self, *, handle_nulls=None) -> Sequence:
         """Convert to a contiguous sequence
 
-        Converts a stream of arrays into a columnar representation
+        Experimentally converts a stream of arrays into a columnar representation
         such that each column is either a contiguous buffer or a ``list``.
         Integer, float, and interval arrays are currently converted to their
         contiguous buffer representation; other types are returned as a list
         of Python objects. The sequences returned by :meth:`to_column` are
         designed to work as input to ``pandas.Series`` and/or ``numpy.array()``.
+        The default conversions are subject to change based on initial user
+        feedback.
 
         Parameters
         ----------
@@ -112,9 +114,9 @@ def nulls_forbid() -> Callable[[CBuffer, Sequence], Sequence]:
     --------
 
     >>> import nanoarrow as na
-    >>> na.Array([1, 2, 3], na.int32()).convert(na.nulls_forbid())
+    >>> na.Array([1, 2, 3], na.int32()).convert(handle_nulls=na.nulls_forbid())
     nanoarrow.c_lib.CBuffer(int32[12 b] 1 2 3)
-    >>> na.Array([1, None, 3], na.int32()).convert(na.nulls_forbid())
+    >>> na.Array([1, None, 3], na.int32()).convert(handle_nulls=na.nulls_forbid())
     Traceback (most recent call last):
     ...
     ValueError: Null present with null_handler=nulls_forbid()
@@ -149,11 +151,12 @@ def nulls_as_sentinel(sentinel=None):
     --------
 
     >>> import nanoarrow as na
-    >>> na.Array([1, 2, 3], na.int32()).convert(na.nulls_as_sentinel())
+    >>> na_array = na.Array([1, 2, 3], na.int32())
+    >>> na_array.convert(handle_nulls=na.nulls_as_sentinel())
     array([1, 2, 3], dtype=int32)
-    >>> na.Array([1, None, 3], na.int32()).convert(na.nulls_as_sentinel())
+    >>> na_array.convert(handle_nulls=na.nulls_as_sentinel())
     array([ 1., nan,  3.])
-    >>> na.Array([1, None, 3], na.int32()).convert(na.nulls_as_sentinel(-999))
+    >>> na_array.convert(handle_nulls=na.nulls_as_sentinel(-999))
     array([   1, -999,    3], dtype=int32)
     """
     import numpy as np
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -84,6 +84,8 @@ def nulls_forbid() -> Callable[[CBuffer, Sequence], Sequence]:
         if len(is_valid) > 0:
             raise ValueError("Null present with null_handler=nulls_forbid()")
 
+        return data
+
     return handle
 
 
@@ -95,7 +97,7 @@ def nulls_debug() -> Callable[[CBuffer, Sequence], Tuple[CBuffer, Sequence]]:
 
 
 def nulls_as_sentinel(sentinel=None):
-    from numpy import array, nan
+    from numpy import array, nan, result_type
 
     if sentinel is None:
         sentinel = nan
@@ -105,9 +107,12 @@ def nulls_as_sentinel(sentinel=None):
         data = array(data, copy=False)
 
         if len(is_valid) > 0:
+            out_type = result_type(data, sentinel)
+            data = array(data, dtype=out_type, copy=True)
             data[~is_valid] = sentinel
-
-        return data
+            return data
+        else:
+            return data
 
     return handle
 
@@ -123,7 +128,7 @@ def nulls_as_masked_array():
         if len(is_valid) > 0:
             return masked_array(data, ~is_valid)
         else:
-            data
+            return data
 
     return handle
 
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -86,9 +86,7 @@ class ArrayViewVisitable:
         designed to work as input to ``pandas.Series`` and/or ``numpy.array()``.
 
         Parameters
-        ---------
-        schema : schema-like, optional
-            An optional schema, passed to :func:`c_array_stream`.
+        ----------
         handle_nulls : callable
             A function returning a sequence based on a validity bytemap and a
             contiguous buffer of values (e.g., the callable returned by
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -158,10 +158,10 @@ def nulls_as_sentinel(sentinel=None):
     import numpy as np
 
     def handle(is_valid, data):
-        is_valid = np.array(is_valid, copy=False)
         data = np.array(data, copy=False)
 
-        if len(is_valid) > 0:
+        if is_valid is not None:
+            is_valid = np.array(is_valid, copy=False)
             out_type = np.result_type(data, sentinel)
             data = np.array(data, dtype=out_type, copy=True)
             data[~is_valid] = sentinel
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -75,7 +75,7 @@ def to_columns(obj, schema=None, handle_nulls=None) -> Tuple[List[str], List[Seq
     >>> names
     ['col1']
     >>> columns
-    [[1, 2, 3]]
+    [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3)]
     """
     return ColumnsBuilder.visit(obj, schema, handle_nulls=handle_nulls)
 
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -47,11 +47,11 @@ class ArrayViewVisitable:
         """
         return ListConverter.visit(self)
 
-    def to_column_list(self, handle_nulls=None) -> Tuple[List[str], List[Sequence]]:
+    def convert_columns(self, handle_nulls=None) -> Tuple[List[str], List[Sequence]]:
         """Convert to a ``list`` of contiguous sequences
 
         Converts a stream of struct arrays into its column-wise representation
-        according to :meth:`to_column`.
+        according to :meth:`convert`.
 
         Paramters
         ---------
@@ -67,7 +67,7 @@ class ArrayViewVisitable:
         >>> import nanoarrow as na
         >>> import pyarrow as pa
         >>> batch = pa.record_batch({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
-        >>> names, columns = na.Array(batch).to_column_list()
+        >>> names, columns = na.Array(batch).convert_columns()
         >>> names
         ['col1', 'col2']
         >>> columns
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -121,6 +121,7 @@ def nulls_forbid() -> Callable[[CBuffer, Sequence], Sequence]:
     """
 
     def handle(is_valid, data):
+        # the is_valid bytemap is only created if there was at least one null
         if is_valid is not None:
             raise ValueError("Null present with null_handler=nulls_forbid()")
 
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -134,8 +134,10 @@ def nulls_as_sentinel(sentinel=None):
 
     A null handler that assigns a sentinel to null values. This is
     done using numpy using the expression ``data[~is_valid] = sentinel``.
-    The default sentinel value will result in ``nan`` assigned to null
-    values in numeric and boolean outputs.
+    The default sentinel value of ``None`` will result in ``nan``
+    assigned to null values in numeric and boolean outputs. This
+    corresponds to numpy's handling of ``None`` in ``np.result_type()``
+    and ``result[~is_valid] = None``.
 
     Parameters
     ----------
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -75,7 +75,7 @@ class ArrayViewVisitable:
         >>> columns
         [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']]
         """
-        return ToColumnListConverter.visit(self, handle_nulls=handle_nulls)
+        return ToColumnsPysequenceConverter.visit(self, handle_nulls=handle_nulls)
 
     def to_pysequence(self, *, handle_nulls=None) -> Sequence:
         """Convert to a contiguous sequence
@@ -265,7 +265,7 @@ class ToPySequenceConverter(ArrayViewVisitor):
         return self._visitor.finish()
 
 
-class ToColumnListConverter(ArrayViewVisitor):
+class ToColumnsPysequenceConverter(ArrayViewVisitor):
     def __init__(self, schema, handle_nulls=None, *, array_view=None):
         super().__init__(schema, array_view=array_view)
 
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -222,9 +222,11 @@ class BooleanColumnBuilder(ArrayStreamVisitor):
 
 
 class NullableColumnBuilder(ArrayStreamVisitor):
-    def __init__(self, schema, column_builder_cls, *, array_view=None):
+    def __init__(
+        self, schema, column_builder_cls=BufferColumnBuilder, *, array_view=None
+    ):
         super().__init__(schema, array_view=array_view)
-        self._column_builder = column_builder_cls(schema, array_view=None)
+        self._column_builder = column_builder_cls(schema, array_view=self._array_view)
 
     def begin(self, total_elements: Union[int, None]):
         self._builder = CBufferBuilder()
@@ -258,8 +260,7 @@ class NullableColumnBuilder(ArrayStreamVisitor):
         self._column_builder.visit_chunk_view(array_view)
 
     def finish(self) -> Any:
-
-        return self._column_builder
+        return self._builder.finish(), self._column_builder.finish()
 
     def _fill_valid(self, length):
         builder = self._builder
--- a/python/src/nanoarrow/visitor.py
+++ b/python/src/nanoarrow/visitor.py
@@ -25,6 +25,12 @@ from nanoarrow.schema import Type
 
 
 class ArrayViewVisitable:
+    """Mixin class providing conversion methods based on visitors
+
+    Can be used with classes that implement ``__arrow_c_stream__()``
+    or ``__arrow_c_array__()``.
+    """
+
     def to_pylist(self) -> List:
         """Convert to a ``list()`` of Python objects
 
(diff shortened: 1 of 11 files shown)


python/src/nanoarrow/visitor.py
```diff
@@ -49,7 +50,7 @@ def to_pylist(obj, schema=None) -> List:
     return ListBuilder.visit(obj, schema)
 
 
-def to_columns(obj, schema=None) -> Tuple[List[str], List[Sequence]]:
+def to_columns(obj, schema=None, handle_nulls=None) -> Tuple[List[str], List[Sequence]]:
     """Convert ``obj`` to a ``list()` of sequences
 
     Converts a stream of struct arrays into its column-wise representation
```
jorisvandenbossche, line 56: For the line below, but can you then clarify when you get a buffer or when a list?

python/src/nanoarrow/visitor.py
```diff
@@ -74,9 +75,60 @@ def to_columns(obj, schema=None) -> Tuple[List[str], List[Sequence]]:
...
+
+def nulls_debug() -> Callable[[CBuffer, Sequence], Tuple[CBuffer, Sequence]]:
+    def handle(is_valid, data):
+        return is_valid, data
+
+    return handle
+
+
+def nulls_as_sentinel(sentinel=None):
+    from numpy import array, result_type
```
jorisvandenbossche, line 101:
```suggestion
    import numpy as np
```

python/src/nanoarrow/visitor.py
```diff
@@ -74,9 +75,60 @@ def to_columns(obj, schema=None) -> Tuple[List[str], List[Sequence]]:
...
+
+    def handle(is_valid, data):
+        is_valid = array(is_valid, copy=False)
+        data = array(data, copy=False)
```
jorisvandenbossche, line 105:
```suggestion
        is_valid = np.array(is_valid, copy=False)
        data = np.array(data, copy=False)
```

Even though the import is just above, I was already confused about to what kind of array we were converting this, and looking for a `from nanoarrow import array` import at the top of the file ... So I think better to stick with explicit module (alias), which is also the common way numpy is used in Python

jorisvandenbossche: Also, i am not entirely sure the `copy=false` is correct here for `is_valid`, or is that already unpacked at this point?

python/src/nanoarrow/visitor.py
```diff
@@ -74,9 +75,60 @@ def to_columns(obj, schema=None) -> Tuple[List[str], List[Sequence]]:
...
+            data = array(data, dtype=out_type, copy=True)
+            data[~is_valid] = sentinel
+            return data
+        else:
+            return data
+
+    return handle
+
+
+def nulls_as_masked_array():
```
jorisvandenbossche, line 118: I wonder if it would be more useful to actually return each array as a tuple of data and mask (the numpy masked array isn't used that much)

jorisvandenbossche (commented):
A few quick comments!

python/src/nanoarrow/visitor.py
paleolimbot: Good to know! I'll remove that then 🙂 

> I wonder if it would be more useful to actually return each array as a tuple of data and mask

Currently that would be `nulls_debug()`, mostly because I couldn't think of a good name (in a previous version it was `nulls_as_mask_and_data()`). Returning a tuple here sort of breaks the "every column is a sequence" guarantee. There could also be a `MaskedSequence` class that wraps the mask and data but that seems like opening a can of worms.

python/src/nanoarrow/visitor.py
paleolimbot: > is that already unpacked at this point?

Yes, all bitmaps are unpacked before they get here. We could also export a bitmap but that is supported basically nowhere (neither pandas nor numpy seem to support it to mark nulls/missing data or to do some kind of subset-assign to otherwise mark them).

One of the things that's not implemented in buffer implementation exported here is the ability to export a typed writable buffer. Technically they all could be written to (we just allocated all of them) which might make some of these or downstream modifications faster. On the flip side, we could try harder to *not* copy things for the case where we only ever saw one big array.

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,187 @@
...
+from nanoarrow._lib import CArrayView, CArrowType, CBuffer, CBufferBuilder
 from nanoarrow.c_array_stream import c_array_stream
+from nanoarrow.c_schema import c_schema_view
 from nanoarrow.iterator import ArrayViewBaseIterator, PyIterator
 from nanoarrow.schema import Type
 
 
-def to_pylist(obj, schema=None) -> List:
-    """Convert ``obj`` to a ``list()` of Python objects
+class ArrayViewVisitable:
```
danepitkin, line 27: Awesome, love this!

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,187 @@
...
+            A function returning a sequence based on a validity bytemap and a
+            contiguous buffer of values (e.g., the callable returned by
+            :meth:`nulls_as_sentinel`).
+
+        Examples
+        --------
+
+        >>> import nanoarrow as na
+        >>> import pyarrow as pa
+        >>> batch = pa.record_batch([pa.array([1, 2, 3])], names=["col1"])
```
danepitkin, line 69: nit: can we use two columns in the docstrings? I think it would better show the functionality.

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,187 @@
...
+            data = np.array(data, dtype=out_type, copy=True)
+            data[~is_valid] = sentinel
+            return data
+        else:
+            return data
+
+    return handle
+
+
+def nulls_debug() -> Callable[[CBuffer, Sequence], Tuple[CBuffer, Sequence]]:
```
danepitkin, line 176: Should we mark this as an internal API? e.g. `_nulls_debug()`

python/src/nanoarrow/visitor.py
```diff
@@ -156,18 +274,23 @@ def __init__(self, schema, *, array_view=None):
...
+        cls, kwargs = _resolve_column_builder_cls(child_schema, handle_nulls)
+        return cls(child_schema, **kwargs, array_view=child_array_view)
 
     def begin(self, total_elements: Union[int, None] = None) -> None:
         for child_visitor in self._child_visitors:
             child_visitor.begin(total_elements)
 
     def visit_chunk_view(self, array_view: CArrayView) -> Any:
+        if array_view.null_count > 0:
+            raise ValueError("null_count > 0 encountered in ColumnsBuilder")
```
danepitkin, line 292: Why is this an error?

python/tests/test_nanoarrow.py
```diff
@@ -183,6 +183,45 @@ def test_c_array_view_dictionary():
...
+def test_c_array_view_null_count():
+    # With explicit null count == 0
+    array = na.c_array_from_buffers(
+        na.int32(), 3, (None, na.c_buffer([1, 2, 3], na.int32())), null_count=0
+    )
+    assert array.view().null_count == 0
+
+    # Infer null count == 0 because of null data buffer
+    array = na.c_array_from_buffers(
+        na.int32(), 3, (None, na.c_buffer([1, 2, 3], na.int32())), null_count=-1
```
danepitkin, line 195: why do we need to add `null_count=-1` in these tests?

danepitkin (commented):
Very cool! The Column builder code is quite nuanced, I need to do another pass over it.

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
-    """Convert ``obj`` to a ``list()` of Python objects
+class ArrayViewVisitable:
+    """Mixin class providing conversion methods based on visitors
+
+    Can be used with classes that implement ``__arrow_c_stream__()``
+    or ``__arrow_c_array__()``.
+    """
+
+    def to_pylist(self) -> List:
+        """Convert to a ``list()`` of Python objects
```
jorisvandenbossche, line 35:
```suggestion
        """Convert to a ``list`` of Python objects
```

(the type is `list` itself, without calling it)

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        >>> import nanoarrow as na
+        >>> from nanoarrow import visitor
+        >>> array = na.Array([1, 2, 3], na.int32())
+        >>> array.to_pylist()
+        [1, 2, 3]
+        """
+        return ListBuilder.visit(self)
+
+    def to_column_list(self, handle_nulls=None) -> Tuple[List[str], List[Sequence]]:
+        """Convert to a ``list()` of contiguous sequences
```
jorisvandenbossche, line 52:
```suggestion
        """Convert to a ``list`` of contiguous sequences
```
(same as above and mismatching number of `` ` ``)

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        >>> columns
+        [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']]
+        """
+        return ColumnsBuilder.visit(self, handle_nulls=handle_nulls)
+
+    def to_column(self, handle_nulls=None) -> Sequence:
+        """Convert to a contiguous sequence
+
+        Converts a stream of arrays into a columnar representation
+        such that each column is either a contiguous buffer or a ``list()``.
```
jorisvandenbossche, line 82:
```suggestion
        such that each column is either a contiguous buffer or a ``list``.
```

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        ---------
+        obj : array stream-like
+            An array-like or array stream-like object as sanitized by
+            :func:`c_array_stream`.
+        schema : schema-like, optional
+            An optional schema, passed to :func:`c_array_stream`.
+        handle_nulls : callable
+            A function returning a sequence based on a validity bytemap and a
+            contiguous buffer of values (e.g., the callable returned by
+            :meth:`nulls_as_sentinel`).
```
jorisvandenbossche, line 98: Mention that the default is `nulls_forbid()`?

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']]
+        """
+        return ColumnsBuilder.visit(self, handle_nulls=handle_nulls)
+
+    def to_column(self, handle_nulls=None) -> Sequence:
+        """Convert to a contiguous sequence
+
+        Converts a stream of arrays into a columnar representation
+        such that each column is either a contiguous buffer or a ``list()``.
+        Integer, float, and interval arrays are currently converted to their
```
jorisvandenbossche, line 83: What is the reason interval arrays are returned as buffer? (just because there is no obvious python object?)

And why are other primitive fixed-width types like timestamp not returned as buffers? (because they are not a _c_ primitive type? And have an obvious python object to use?) 

I can imagine that for certain purposes, you would actually want the integers behind a timestamp column, because that is much cheaper to work with than a list of python datetime.datetime objects. 

Should we make this user configurable?

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        contiguous buffer representation; other types are returned as a list
+        of Python objects. The sequences returned by :meth:`to_column` are
+        designed to work as input to ``pandas.Series`` and/or ``numpy.array()``.
+
+        Parameters
+        ---------
+        obj : array stream-like
+            An array-like or array stream-like object as sanitized by
+            :func:`c_array_stream`.
```
jorisvandenbossche, line 92:
```suggestion
```

For the methods in this mixin class, this is self?

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        schema : schema-like, optional
+            An optional schema, passed to :func:`c_array_stream`.
```
jorisvandenbossche, line 94: This is not actually present in the signature?

python/src/nanoarrow/visitor.py
jorisvandenbossche: And maybe also list the built-in options

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+    >>> na.Array([1, None, 3], na.int32()).to_column(na.nulls_forbid())
+    Traceback (most recent call last):
+    ...
+    ValueError: Null present with null_handler=nulls_forbid()
     """
-    return ListBuilder.visit(obj, schema)
 
+    def handle(is_valid, data):
+        if len(is_valid) > 0:
+            raise ValueError("Null present with null_handler=nulls_forbid()")
```
jorisvandenbossche, line 128: It's not because there is a validity buffer that it includes any nulls? Or did you ensure that before? (eg not convert the bitmap to bytemap if the bitmap is all set?)

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
-    ---------
-    obj : array stream-like
-        An array-like or array stream-like object as sanitized by
-        :func:`c_array_stream`.
-    schema : schema-like, optional
-        An optional schema, passed to :func:`c_array_stream`.
+    A null handler that assigns a sentinel to null values. This is
+    done using numpy using the expression ``data[~is_valid] = sentinel``.
+    The default sentinel value will result in ``nan`` assigned to null
+    values in numeric and boolean outputs.
```
jorisvandenbossche, line 141: I am not entirely sure this is the case for boolean? (setting None into a boolean array sets False)

jorisvandenbossche: Ah, but you ensure to convert int/bool to float in case there are nulls, so that setting as NaN indeed works correctly. Maybe clarify that here

python/src/nanoarrow/visitor.py
paleolimbot: I changed the name to `nulls_separate()` and exported it...from @jorisvandenbossche's comment it sounds like it would be useful (in the situation where caller wants to handle nulls completely on their own).

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+    """Return nulls as a tuple of is_valid, data
+
+    A null handler that returns its components.
+
+    Examples
+    --------
+
+    >>> import nanoarrow as na
+    >>> na.Array([1, 2, 3], na.int32()).to_column(na.nulls_separate())
+    (nanoarrow.c_lib.CBuffer(uint8[0 b] ), nanoarrow.c_lib.CBuffer(int32[12 b] 1 2 3))
```
jorisvandenbossche, line 186: Should we return `None` here instead of an empty buffer? Because if you convert both outputs to numpy arrays, you can have numpy arrays of different length (of course, it's not that one of checking for None or for empty / len 0 is much harder than the other)

jorisvandenbossche (commented):
First set of comments. And thanks for the nice docstrings!

python/src/nanoarrow/visitor.py
paleolimbot: If there are any nulls here the column results will be bogus! (We have no way to propagate a top-level nulls through into the child columns). At the "record batch" level nulls don't exist (typically they are exported as an explicitly non-nullable struct; however, at least one implementation doesn't do this so we have to check for actual nulls whilst iterating instead of erroring when we see the schema).

python/src/nanoarrow/visitor.py
jorisvandenbossche: `nulls_separate` sounds in any case as a better name! (and yes I think it would be useful)

python/tests/test_nanoarrow.py
paleolimbot: I added comments to each of these...-1 is the sentinel for "not computed yet". Technically it's the default but here it's exactly what we're testing.

python/src/nanoarrow/visitor.py
paleolimbot: Yes...the bool buffer indicating validity is never allocated unless a null is actually present (and is None otherwise). I added this to the `handle_nulls` documentation!

python/src/nanoarrow/visitor.py
paleolimbot: Yes, it's a bit of a side-effect of how numpy handles `None` with `result_type()`. I added a comment clarifying this.

```python
import numpy as np

sentinel = None
arr = np.array([True, False, True, True], np.result_type(sentinel, np.bool_))
arr[2] = sentinel
arr
#> array([ 1.,  0., nan,  1.])
```

python/src/nanoarrow/visitor.py
paleolimbot: Yes! I updated this to return `None`, which is less ambiguous. One could also update this such that `handle_nulls()` is never called unless there are actually nulls?

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,185 @@
...
+    nanoarrow.c_lib.CBuffer(int32[12 b] 1 2 3)
+    >>> na.Array([1, None, 3], na.int32()).to_column(na.nulls_forbid())
+    Traceback (most recent call last):
+    ...
+    ValueError: Null present with null_handler=nulls_forbid()
     """
-    return ListBuilder.visit(obj, schema)
+
+    def handle(is_valid, data):
+        if is_valid is not None:
```
jorisvandenbossche, line 124:
```suggestion
        # the is_valid bytemap is only created if there was at least one null
        if is_valid is not None:
```

python/src/nanoarrow/visitor.py
jorisvandenbossche: Can you add a brief comment about this in the code?

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        designed to work as input to ``pandas.Series`` and/or ``numpy.array()``.
+
+        Parameters
+        ---------
+        obj : array stream-like
+            An array-like or array stream-like object as sanitized by
+            :func:`c_array_stream`.
+        schema : schema-like, optional
+            An optional schema, passed to :func:`c_array_stream`.
+        handle_nulls : callable
```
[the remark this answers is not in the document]
paleolimbot: Yes, and only for non-nullable fields. Maybe there should be a `handle_valid` and `handle_nulls`? Or maybe just a `handle_result()`?

python/src/nanoarrow/visitor.py
```diff
@@ -124,27 +242,26 @@ def finish(self) -> Any:
...
+    def visit_chunk_view(self, array_view: CArrayView) -> None:
+        self._visitor.visit_chunk_view(array_view)
 
-    def finish(self) -> List:
-        return self._lst
+    def finish(self) -> Any:
+        return self._visitor.finish()
 
 
 class ColumnsBuilder(ArrayStreamVisitor):
```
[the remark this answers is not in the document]
paleolimbot: It's been through a few iterations of naming 🙂 . Maybe `sequence` is better than `column`? The gist of it is "something that either numpy or Pandas can understand" and/or "we dealt with all the hard parts of a non-contiguous or non-zero offset array so you didn't have to". I could stick with `Visitor` instead of `Builder`, too (but these are internal for now so at least that part can get renamed).

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        --------
+        >>> import nanoarrow as na
+        >>> from nanoarrow import visitor
+        >>> array = na.Array([1, 2, 3], na.int32())
+        >>> array.to_pylist()
+        [1, 2, 3]
+        """
+        return ListBuilder.visit(self)
+
+    def to_column_list(self, handle_nulls=None) -> Tuple[List[str], List[Sequence]]:
```
danepitkin, line 50: would `to_table` potentially be more descriptive?

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,186 @@
...
+        >>> batch = pa.record_batch({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
+        >>> names, columns = na.Array(batch).to_column_list()
+        >>> names
+        ['col1', 'col2']
+        >>> columns
+        [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']]
+        """
+        return ColumnsBuilder.visit(self, handle_nulls=handle_nulls)
+
+    def to_column(self, handle_nulls=None) -> Sequence:
```
danepitkin, line 78: Am I correct in thinking this is a concatenation function? e.g. chunked array -> array

python/src/nanoarrow/visitor.py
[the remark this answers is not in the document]
danepitkin: I'll add that I also had a bit of trouble comprehending the naming. I think it might be worthwhile to rename "column" to something else. Technically, its just a contiguous array, but `array` is overloaded. Are these technically all concatenation helpers?

python/src/nanoarrow/visitor.py
danepitkin: Maybe not, cause it truly is a pylist of na.arrays. But it is tabular in nature.

python/src/nanoarrow/visitor.py
danepitkin: Maybe `merge_chunks` or `consolidate_chunks` or `unify_chunks` could be a more descriptive name?

jorisvandenbossche: But it does concatenation + conversion to python lists at the same time, so just naming it about the first is also not descriptive enough

danepitkin (commented):
I like the improvements! I did end up adding a more comments on naming. Let me know what you think.

python/src/nanoarrow/visitor.py
```diff
@@ -15,68 +15,192 @@
...
+        >>> batch = pa.record_batch({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
+        >>> names, columns = na.Array(batch).convert_columns()
+        >>> names
+        ['col1', 'col2']
+        >>> columns
+        [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']]
+        """
+        return ColumnListConverter.visit(self, handle_nulls=handle_nulls)
+
+    def convert(self, *, handle_nulls=None) -> Sequence:
```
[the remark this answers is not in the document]
danepitkin: ehh maybe not use `pysequence` because that would mean the object implements the python sequence protocol.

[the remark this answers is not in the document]
danepitkin: although the type hint does say a sequence is returned.. do pybuffer and pylist both implement the sequence protocol?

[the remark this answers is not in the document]
paleolimbot: They do!

python/src/nanoarrow/visitor.py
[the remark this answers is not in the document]
paleolimbot: At Dane's suggestion these are now `to_pysequence()` and `to_columns_pysequence()`. I think these can/should get renamed when we see exactly how/if they are used (it may be that everybody will want custom logic and the high-level convert method will get unused).

python/src/nanoarrow/visitor.py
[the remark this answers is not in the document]
paleolimbot: I kept this as is for now but I think there will eventually need to be some way to inject more customization into the process (including exactly when the null handler gets called).

python/src/nanoarrow/visitor.py
paleolimbot: I went with `to_columns_pysequence()` for now...it's not *quite* a table and I am not sure inventing one here is the solution either. Unfortunately it's also the most useful function! Hopefully its usage will give some guidance as to what it should be named or if it needs to be removed/live somewhere else.

python/src/nanoarrow/visitor.py
paleolimbot: Yes, concatenation is a side effect of the fact that the non-Arrow pydata universe can't handle chunked data. This function is specifically trying to get data *out* of Arrow land. Your suggestion of `to_pysequence()` is much better than `to_column()` so I went with that!

python/src/nanoarrow/visitor.py
```diff
@@ -15,76 +15,201 @@
...
+        >>> import nanoarrow as na
+        >>> import pyarrow as pa
+        >>> batch = pa.record_batch({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
+        >>> names, columns = na.Array(batch).to_columns_pysequence()
+        >>> names
+        ['col1', 'col2']
+        >>> columns
+        [nanoarrow.c_lib.CBuffer(int64[24 b] 1 2 3), ['a', 'b', 'c']]
+        """
+        return ToColumnListConverter.visit(self, handle_nulls=handle_nulls)
```
danepitkin, line 78: optional nit: do we want to update this to be `ToColumnsPySequenceConverter` now?

danepitkin (approved):
Overall LGTM! Thanks for being receptive to all of our feedback.

python/src/nanoarrow/visitor.py
```diff
@@ -124,56 +248,209 @@ def finish(self) -> Any:
         return None
 
 
-class ListBuilder(ArrayStreamVisitor):
-    def __init__(self, schema, *, iterator_cls=PyIterator, array_view=None):
+class DispatchingConverter(ArrayStreamVisitor):
```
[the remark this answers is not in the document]
paleolimbot: I went with `ToPySequenceConverter` (for brevity) but it's a great point that ti's unclear if an `XXXConverter` is converting to or from an `XXX`!

python/src/nanoarrow/visitor.py
```diff
@@ -124,56 +248,209 @@ def finish(self) -> Any:
...
 
-    def finish(self) -> List:
-        return self._lst
+    def finish(self) -> Any:
+        return self._visitor.finish()
 
 
-class ColumnsBuilder(ArrayStreamVisitor):
-    def __init__(self, schema, *, array_view=None):
+class ColumnListConverter(ArrayStreamVisitor):
```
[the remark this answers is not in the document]
paleolimbot: I did keep the "column" terminology" here...the intention with the return value of this function is that the list of sequences will be subcomponents of a table...we're just agnostic here to exactly which "table" they are going to end up in.

PR #471, opened 2024-05-15
Fix meson dependencies in tests

WillAyd:


WillAyd:
Believe these are required with changes in https://github.com/apache/arrow-nanoarrow/pull/404 

Where does the weekly meson build output go to? I haven't looked for anything in particular so guessing the tests fail on the weekly meson build we have set up

paleolimbot (approved):
Thanks!

The results get emailed to me every week if they fail, perhaps because I'm the one who hit "merge". I'm not awesome about fixing the weekly build errors promptly but I will make sure to be more organized about it moving forward. They're available for viewing here: https://github.com/apache/arrow-nanoarrow/actions/workflows/meson-build.yaml

PR #483, opened 2024-05-23
feat: Meson build system for nanoarrow-ipc extension

WillAyd:


WillAyd:
Meson's handling of subprojects is way easier I think than CMake's. If down the road we wanted to replace CMake with Meson there is a ton that we could simplify within these extensions, and _maybe_ could even consider them as separate projects

extensions/nanoarrow_ipc/subprojects/nanoarrow.wrap
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[wrap-git]
+url = ../../..
```
WillAyd, line 19: In the main project right now we ask end users to run `meson wrap install <lib>`, but that doesn't make sense to do with this nanoarrow.wrap, so I ended up putting all the wrap files here

Maybe even in the main project we should just add the wrap files? That would technically pin a version for third party dependencies (if not installed elsewhere on the system) which may or may not be desired

extensions/nanoarrow_ipc/subprojects/flatcc.wrap
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[wrap-git]
+method = cmake
```
WillAyd, line 19: flatcc is working to replace (?) CMake with Meson. That would simplify the wrap / subproject usage for flatcc if and when it goes live

https://github.com/dvidelabs/flatcc/pull/262

WillAyd:
I haven't used these in particular so I'm just kind of guessing, but it feels like it would make sense to combine them. I think for consumers having to potentially install three different libraries for nanoarrow instead of just one library that you can choose the dependencies out of is overkill. And as you've mentioned already, there appears to be a ton of overlap between these projects' dependencies

extensions/nanoarrow_ipc/meson.build
```diff
@@ -0,0 +1,42 @@
...
+    ]
+)
+
+if get_option('NANOARROW_IPC_CODE_COVERAGE')
+  error(
+      '''For meson builds, please use '-Db_coverage=true' flag
+      instead of -DNANOARROW_IPC_CODE_COVERAGE=true''')
+endif
+
+if get_option('NANOARROW_IPC_BUNDLE')
```
WillAyd, line 38: @paleolimbot without being super knowledgeable on the end goal of these bundle options, are these used by consumers downstream or just for internal use? I see the Python build uses the bundle to auto generate a Cython header, but not sure of other uses cases and if its worth trying to replicate in Meson

paleolimbot: I don't think there's any reason to expose this in Meson...this is currently how the files in dist/ are generated (e.g., the single-file bundles for each component). The fact that they are written in CMake is something that initially made it easier to test that the single-file distribution actually worked...I think it might make more sense to rewrite the amalgamation scripts in Python at some point though.

extensions/nanoarrow_ipc/subprojects/flatcc.wrap
paleolimbot: It seems like the intention is to support both, although we vendor flatcc anyway for now. I should be more active contributing to that upstream since the IPC reader 100% relies on it!

extensions/nanoarrow_ipc/subprojects/nanoarrow.wrap
paleolimbot: It seems like this complexity would be solved if the IPC extension just lived with the main sources!

paleolimbot (commented):
> I haven't used these in particular so I'm just kind of guessing, but it feels like it would make sense to combine them.

Great! I will aim for that in a near-term PR. I am happy to work through this and merge it before that refactor or afterward (your call!)

WillAyd:
Let's just wait to merge this until after then - there is no rush for this and I think waiting will make that consolidation easier

meson.build
```diff
@@ -34,3 +34,14 @@ subdir('src/nanoarrow')
 if get_option('benchmarks')
   subdir('dev/benchmarks')
 endif
+
+
+if get_option('apps')
```
WillAyd, line 39: I should really move the `meson.build` from the src/nanoarrow directory up to here to mirror what is done for CMake, but am leaving that for a separate PR. For now just put this apps piece up here since it does not live in src/nanoarrow

src/nanoarrow/nanoarrow_testing_test.cc
```diff
@@ -470,7 +470,7 @@ TEST(NanoarrowTestingTest, NanoarrowTestingTestFieldMetadata) {
         NANOARROW_RETURN_NOT_OK(ArrowSchemaSetMetadata(schema, "\0\0\0\0"));
         return NANOARROW_OK;
       },
-      [](ArrowArray* array) { return NANOARROW_OK; }, &WriteFieldJSON,
+      [](ArrowArray*) { return NANOARROW_OK; }, &WriteFieldJSON,
```
WillAyd, line 473: Not strictly related but it showed as a warning we could control with the Meson warning configuration

WillAyd:
Interesting to see the valgrind error in CI right now. I assume this is also why the windows build is failing, although interesting our CMake setup didn't catch it

https://github.com/apache/arrow-nanoarrow/actions/runs/9455933515/job/26046780100?pr=483#step:7:338

WillAyd:
Seems like the error has to do with the fields being sent to ArrowErrorSet in this function:

https://github.com/apache/arrow-nanoarrow/blob/f443e46163ded43ddb5e024c1a2f1fddd3b43c3f/src/nanoarrow/array.c#L1247

Not familiar with the run end encoded implementation but I think we are missing something with how fields are being initialized and accessed through that code block

src/apps/dump_stream.c
```diff
@@ -23,8 +23,6 @@
 
 void dump_schema_to_stdout(struct ArrowSchema* schema, int level, char* buf,
                            int buf_size) {
-  int n_chars = ArrowSchemaToString(schema, buf, buf_size, 0);
```
paleolimbot, line 26: I think this is still needed (or else the top-level type will never appear in the dump). Maybe just remove `int n_chars = ` (if the unused variable was a problem?)

src/nanoarrow/nanoarrow_ipc_decoder_test.cc
```diff
@@ -602,8 +599,10 @@ TEST(NanoarrowIpcTest, NanoarrowIpcSharedBufferThreadSafeDecode) {
   std::thread threads[10];
   for (int i = 0; i < 10; i++) {
     threads[i] = std::thread([&arrays, i, &one_two_three_le] {
-      memcmp(arrays[i].children[0]->buffers[1], one_two_three_le,
-             sizeof(one_two_three_le));
+      auto result = memcmp(arrays[i].children[0]->buffers[1], one_two_three_le,
+                           sizeof(one_two_three_le));
+      // discard result to silence -Wunused-value
+      (void)result;
```
paleolimbot, line 605: Would `NANOARROW_UNUSED()` work here?

src/nanoarrow/nanoarrow_testing_test.cc
paleolimbot: Makes sense to me!

dev/release/rat_exclude_files.txt
```diff
@@ -18,3 +18,4 @@ python/src/nanoarrow/dlpack_abi.h
 subprojects/google-benchmark.wrap
 subprojects/gtest.wrap
 subprojects/nlohmann_json.wrap
+subprojects/zlib.wrap
```
paleolimbot, line 21: Is there a reason these are being ignored rather than just adding the Apache license header to the files? (It seems like those file types support comments and some of the wrap files already have the license header added)

src/nanoarrow/meson.build
```diff
@@ -64,6 +64,25 @@ incdir = include_directories('..')
 nanoarrow_dep = declare_dependency(include_directories: [curdir, incdir],
                                    link_with: nanoarrow_lib)
 
+if get_option('ipc')
+    cmake = import('cmake')
+    cmake_opts = cmake.subproject_options()
+    cmake_opts.add_cmake_defines({'CMAKE_POSITION_INDEPENDENT_CODE': true})
+    flatcc_subproj = cmake.subproject('flatcc', options: cmake_opts)
+    flatcc_dep = flatcc_subproj.dependency('flatccrt')
```
paleolimbot, line 72: This would mean that CMake is required to build the flatcc runtime? I get that it is more "meson"ic to use the wrap file; however, it might also be a little strange that building nanoarrow_ipc via CMake will give you a slightly different result than using Meson. I think I'd prefer that the default build uses the vendored version (via a pure Meson compile of the four files in thirdparty/flatcc/src/runtime), and perhaps our dependency relationship with flatcc could be improved in both CMake/Meson (perhaps with some contributions to the upstream setup).

src/nanoarrow/meson.build
```diff
@@ -147,4 +166,38 @@ if get_option('tests')
...
+              'deps': [nanoarrow_dep, flatcc_dep, gtest_dep],
+          },
+      }
+
+      foreach name, config : ipc_test_files
+          exc = executable(
+              name + '-test',
+              name.replace('-', '_') + '_test.cc',
+              link_with: nanoarrow_ipc_lib,
+              dependencies: config['deps']
```
paleolimbot, line 198: Would it be possible to factor more common dependencies like `gtest_dep` into `link_with`?

(Also: I don't think that `nanoarrow_dep` or `flatcc_dep` is needed because it should be handled as a transitive dependency of `nanoarrow_ipc_lib`? I might have put that in the CMake due to some weirdness about it being a library built on top of nanoarrow rather than part of nanoarrow itself)

ci/scripts/build-with-meson.sh
```diff
@@ -68,12 +68,12 @@ function main() {
     show_header "Run test suite"
     meson configure -Dtests=true -Db_coverage=true
     meson compile
-    meson test --wrap valgrind
+    meson test --wrap='valgrind --track-origins=yes' --print-errorlogs
```
paleolimbot, line 71: I often use `--tool=memcheck --leak-check=full` (but this seems to give great output so take or leave!)

paleolimbot (commented):
A few comments but generally looking great!

dev/release/rat_exclude_files.txt
WillAyd: The ones I have excluded were auto generated by `meson wrap install <library>` whereas the one that I created from scratch (flatcc.wrap) included the Arrow licensing agreement.

Not sure if that distinction matters. Happy to add license agreement to even the auto generated ones

src/nanoarrow/meson.build
WillAyd: OK cool will take a look. I believe flatcc is trying to switch to Meson as well, so yea can look and see if that makes for a good upstream contribution

ci/scripts/build-with-meson.sh
WillAyd: Sure can definitely add the leak-check=full - its just a tradeoff between performance and visibility. This runs pretty quick in CI as is

src/nanoarrow/meson.build
WillAyd: Not sure I understand the first question - what problem are you trying to solve by making that a direct link argument? The way the dependency is declared (see [wrapdb source](https://github.com/mesonbuild/wrapdb/blob/a67b90d64e63fd52d0026ec24cb9972b831c2564/subprojects/packagefiles/gtest/googletest/meson.build#L14)) I believe Meson will take care of that linkage for you, alongside having the proper include and threads dependency

As far as the second question goes the way I had this previously the dependencies were not transitive, but I've since updated the nanoarrow_ipc_dep to include those:

```python
nanoarrow_ipc_dep = declare_dependency(..., dependencies: [nanoarrow_dep, flatcc_dep])
```

dev/release/rat_exclude_files.txt
paleolimbot: I think it would be best to add the license headers rather than put it in the excluded files (although I think it is fine either way since it was generated for us).

ci/scripts/build-with-meson.sh
paleolimbot: Got it! Let's leave it (the memcheck CI job + verification both already do  the memcheck/full version).

src/nanoarrow/meson.build
paleolimbot: > Not sure I understand the first question

It seemed like the loop was handling linking each test with `nanoarrow_ipc_lib` and I was wondering if the loop could also handle linking the dependencies that were common to every test executable. This is probably just me misunderstanding meson and feel free to ignore if it doesn't make sense 🙂 

> As far as the second question goes the way I had this previously the dependencies were not transitive, but I've since updated the nanoarrow_ipc_dep to include those

Great! I mostly just want to make sure my vauge mental model of what's going on here is correct 🙂

WillAyd: I think when coming from CMake it is helpful to understand that the word "dependency" means something else in Meson. In CMake, dependency refers to "something that needs to be built before this target can be". In Meson, a dependency helps you transitively use include directories / link targets. You declare this via the `declare_dependency` function:

```python
base_dep = declare_dependency(
    include_directories: ['foo_dir'],
    link_with: [threads_lib],
)
```

So when you end up using that dependency on another target, it will transitively carry over the include / link flags for you, i.e.

```python
child_lib = library('child', dependencies: [base_dep])
```

expands into

```python
child_lib = library('child',
    include_directories: ['foo_dir'],
    link_with: [threads_lib],
)
```

(at least as far as I understand things)

src/nanoarrow/meson.build
WillAyd: I think this is solved in https://github.com/apache/arrow-nanoarrow/pull/483/files/82a6b48e4218f24b320cfb9e05f972c59d2cdd37..b878e23bb721f07a058e9f63a4db3aa6588f0faa 

Basically Meson allows you to provide your own "patch" file for systems that don't use Meson. The file(s) go in `subprojects/packagefiles/<patch_dir>` and the wrap system still takes care of fetching / downloading the library for you, rather than requiring you to vendor into your source tree

paleolimbot (approved):
Thank you!

The Meson build failure is a separate issue related to attempting to print very large numbers (unrelated to this PR).

src/nanoarrow/meson.build
WillAyd: Ugh set up the subproject here correctly but forgot to come back and update the code block here. Will push up a follow up PR to correct

PR #484, opened 2024-05-23
feat: Add Meson support in nanoarrow_device

WillAyd:


extensions/nanoarrow_device/src/nanoarrow/meson.build
```diff
@@ -0,0 +1,67 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+nanoarrow_dep = dependency('nanoarrow')
+
+apple_dep = dependency('appleframeworks',
```
WillAyd, line 20: I am not able to test the apple or cuda deps unfortunately

ci/scripts/build-with-meson.sh
```diff
@@ -66,7 +66,7 @@ function main() {
     pushd "${SANDBOX_DIR}"
 
     show_header "Run test suite"
-    meson configure -Dtests=true -Db_coverage=true
+    meson configure -Dtests=true -Db_coverage=true -Dipc=true -Ddevice=true
```
WillAyd, line 69: Missed this on the IPC change. If we want to expand this job in the future would need to accept these options dynamically, but this should work for now

paleolimbot (commented):
This looks great!

When I run:

```
meson configure -Dtests=true -Ddevice=true -Dmetal=true
```

I get

```
../src/nanoarrow/nanoarrow_device_metal_test.cc:22:10: fatal error: 'Metal/Metal.hpp' file not found
#include <Metal/Metal.hpp>
         ^~~~~~~~~~~~~~~~~
../src/nanoarrow/nanoarrow_device_metal_test.cc:22:10: note: did not find header 'Metal.hpp' in framework 'Metal' (loaded from '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks')
```

Metal C++ is its own thing and is a sort of header-only wrapper around the Objective C (not included directly in the framework):

https://github.com/apache/arrow-nanoarrow/blob/d6eb52bb0f18c2b2a49456a95d97d646ab54d3b4/CMakeLists.txt#L389-L400

I also tried

```
meson configure -Dtests=true -Ddevice=true -Dcuda=true
```

...and get 

```
c++  -o src/nanoarrow/nanoarrow-device-cuda-test src/nanoarrow/nanoarrow-device-cuda-test.p/nanoarrow_device_cuda_test.cc.o src/nanoarrow/nanoarrow-device-cuda-test.p/.._.._subprojects_googletest-1.14.0_googletest_src_gtest-all.cc.o src/nanoarrow/nanoarrow-device-cuda-test.p/.._.._subprojects_googletest-1.14.0_googletest_src_gtest_main.cc.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,/home/dewey/rscratch/arrow-nanoarrow/builddir/src/nanoarrow -Wl,--start-group src/nanoarrow/libnanoarrow_device.so src/nanoarrow/libnanoarrow.so -pthread /usr/local/cuda/lib64/libcudart_static.a -lrt -lpthread -ldl /usr/local/cuda/lib64/libcublas.so -Wl,--end-group
/usr/bin/ld: src/nanoarrow/nanoarrow-device-cuda-test.p/nanoarrow_device_cuda_test.cc.o: in function `NanoarrowDeviceCuda_DeviceCudaBufferCopy_Test::TestBody()':
nanoarrow_device_cuda_test.cc:(.text+0x1ec2): undefined reference to `ArrowDeviceCuda'
/usr/bin/ld: nanoarrow_device_cuda_test.cc:(.text+0x1f34): undefined reference to `cuDeviceGet'
```

...and this is maybe because `cublas` is not the right thing to link to (the thing we're using is called the "driver" API, which is *not* the "runtime" API). Also possibly because there needs to be a `-DNANOARROW_WITH_CUDA` compile define to get the `ArrowDeviceCuda` symbol into `nanoarrow_device.so` (still workshopping the final version of that).

In the next week or two we'll have a CUDA runner that will help with this 🙂

WillAyd:
Sweet thanks for giving this a shot

> Metal C++ is its own thing and is a sort of header-only wrapper around the Objective C (not included directly in the framework):

Possibly, though I would still hope Meson could handle that. I found another project in the WrapDB that uses Metal:

https://github.com/mesonbuild/wrapdb/blob/2d9ed3ad5bed879d8aa010de21ae165f6cf5ce4d/subprojects/packagefiles/imgui/meson.build#L49

Maybe we just need to add `required: get_option('metal')` to our dependency call?

> ...and this is maybe because `cublas` is not the right thing to link to (the thing we're using is called the "driver" API, which is _not_ the "runtime" API). Also possibly because there needs to be a `-DNANOARROW_WITH_CUDA` compile define to get the `ArrowDeviceCuda` symbol into `nanoarrow_device.so` (still workshopping the final version of that).

Ah OK...apologies as I know embarassingly little about CUDA. But in the configuration now I have `modules: [cublas]` as I just copy-pasted that from the Meson documentation without much thought. Maybe that needs to be `modules: [cuda-driver]`?

WillAyd:
Ah ignore my comment above about Metal - I confused `required: get_option('metal')` for something else. The only other difference I see in that case is the `AppKit` framework inclusion, but that seems unlikely to be the root cause. 

I'll ask on the Meson IRC channel tomorrow how to resolve that header; they are usually really responsive to questions like this

WillAyd:
Actually was able to borrow a MacBook to test out the metal stuff. Didn't realize metal-cpp was not installed with the system framework :-)

So the metal stuff should be compiling now. There were also some issues with the source files that needed to be corrected. Note that the metal test fails - I haven't looked into why but things at least compile now:

```sh
1/1 nanoarrow-device-metal        FAIL            0.03s   killed by signal 11 SIGSEGV
>>> MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 MALLOC_PERTURB_=159 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 /Users/willayd/clones/arrow-nanoarrow/builddir/src/nanoarrow/nanoarrow-device-metal-test
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Running main() from /tmp/googletest-20230817-5025-15wtgyq/googletest-1.14.0/googletest/src/gtest_main.cc
[==========] Running 10 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 6 tests from NanoarrowDeviceMetal
[ RUN      ] NanoarrowDeviceMetal.DefaultDevice
../src/nanoarrow/nanoarrow_device_metal_test.cc:30: Failure
Expected equality of these values:
  ArrowDeviceMetalInitDefaultDevice(device.get(), nullptr)
    Which is: 22
  0

[  FAILED  ] NanoarrowDeviceMetal.DefaultDevice (15 ms)
[ RUN      ] NanoarrowDeviceMetal.DeviceGpuBufferInit
```

src/nanoarrow/meson.build
```diff
@@ -80,6 +80,34 @@ if get_option('ipc')
...
+  if get_option('metal')
+    metal_dep = dependency('appleframeworks', modules: ['Foundation', 'Metal'])
+    metal_cpp_dep = dependency('metal-cpp')
+    device_deps += metal_dep
+    device_deps += metal_cpp_dep
+    device_srcs += 'nanoarrow_device_metal.cc'
+  endif
+
+  if get_option('cuda')
+    cuda_dep = dependency('cuda', modules: ['cuda_driver'])
```
WillAyd, line 97: Just guessing on the module name. Hoping it matches what CMake has (and what exists in our current CMake config)

https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html

WillAyd:
Yea its an old Intel device - so is the metal-cpp header only required for that?

src/nanoarrow/meson.build
paleolimbot: I tried a number of things here, but ultimately was unable to make the linking work properly. We need `libcuda.so`, which I couldn't find in the Meson CUDA sources. It's the "driver" library/API, which the `cudart` would also need to link; however, I think the Meson CUDA stuff is almost 100% designed for `nvcc`, and we're using regular `gcc` (i.e., we're not actually using NVIDIA's compiler, we're using a regular compiler and linking to the library).

It's fine to error here for now until somebody with both Meson experience and a graphics card feels like debugging this!

src/nanoarrow/meson.build
```diff
@@ -80,6 +80,34 @@ if get_option('ipc')
...
+    device_deps += cuda_dep
+    device_srcs += 'nanoarrow_device_cuda.cc'
```
paleolimbot, line 99:
```suggestion
    device_srcs += 'nanoarrow_device_cuda.c'
```

paleolimbot (approved):
Thank you!

Issue #245, opened 2023-06-22
Device extension should implement async buffer copying for CUDA


paleolimbot:
Currently, the device extension is unreasonably slow when copying any array to a device because it synchronizes for every buffer copy. For string/binary types, an additional copy is needed to find the length of the next buffer. There is no technical limitation preventing these copies from occuring in parallel; however, the initial PR for this didn't get there.

References:

- https://github.com/apache/arrow-nanoarrow/pull/205#discussion_r1235579653
- https://github.com/apache/arrow-nanoarrow/pull/205#discussion_r1237463786
- https://github.com/apache/arrow-nanoarrow/pull/205#discussion_r1226854070
- https://github.com/apache/arrow-nanoarrow/pull/205#discussion_r1224780614

PR #488, opened 2024-05-24
refactor(extensions/nanoarrow_device): Migrate CUDA device implementation to use the driver API

### LINKED ISSUE #246: Device extension should use the CUDA driver library (not the CUDA runtime library)
The initial implementation used the runtime library; however, the driver library is more appropriate (better stability guarantees) plus this is also what Arrow C++ uses in its CUDA implementation.

References:

- https://github.com/apache/arrow-nanoarrow/pull/205#discussion_r1235520762
- https://github.com/apache/arrow-nanoarrow/pull/205#discussion_r1224784183

paleolimbot:
Closes #246.

This PR doesn't change much about the existing implementation (and I think there are some things that need to be changed!), it just eliminates the dependency on the runtime library. The driver API is a better fit here anyway since we're doing very low-level things!

This isn't tested in CI yet (working on that here: https://github.com/apache/arrow-nanoarrow/pull/490 ).

### DIFF
--- a/extensions/nanoarrow_device/CMakeLists.txt
+++ b/extensions/nanoarrow_device/CMakeLists.txt
@@ -119,7 +119,7 @@ else()
   if(NANOARROW_DEVICE_WITH_CUDA)
     find_package(CUDAToolkit REQUIRED)
     set(NANOARROW_DEVICE_SOURCES_CUDA src/nanoarrow/nanoarrow_device_cuda.c)
-    set(NANOARROW_DEVICE_LIBS_CUDA CUDA::cudart_static)
+    set(NANOARROW_DEVICE_LIBS_CUDA CUDA::cuda_driver)
     set(NANOARROW_DEVICE_DEFS_CUDA "NANOARROW_DEVICE_WITH_CUDA")
   endif()
 
--- a/extensions/nanoarrow_device/CMakeLists.txt
+++ b/extensions/nanoarrow_device/CMakeLists.txt
@@ -138,6 +138,7 @@ else()
                                                       ${NANOARROW_DEVICE_DEFS_CUDA})
   target_link_libraries(nanoarrow_device PUBLIC ${NANOARROW_DEVICE_LIBS_METAL}
                                                 ${NANOARROW_DEVICE_LIBS_CUDA})
+  target_compile_definitions(nanoarrow_device PUBLIC "$<$<CONFIG:Debug>:NANOARROW_DEBUG>")
 
   install(TARGETS nanoarrow_device DESTINATION lib)
   install(FILES src/nanoarrow/nanoarrow_device.h DESTINATION include/nanoarrow)
--- a/extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
+++ b/extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
@@ -220,6 +220,9 @@ static ArrowErrorCode ArrowDeviceCudaBufferCopyInternal(struct ArrowDevice* devi
                                                         struct ArrowBufferView dst,
                                                         int* n_pop_context,
                                                         struct ArrowError* error) {
+  // Note: the device_src/sync event must be synchronized before calling these methods,
+  // even though the cuMemcpyXXX() functions may do this automatically in some cases.
+
   if (device_src->device_type == ARROW_DEVICE_CPU &&
       device_dst->device_type == ARROW_DEVICE_CUDA) {
     struct ArrowDeviceCudaPrivate* dst_private =
@@ -275,17 +278,14 @@ static ArrowErrorCode ArrowDeviceCudaBufferCopyInternal(struct ArrowDevice* devi
 
   } else if (device_src->device_type == ARROW_DEVICE_CPU &&
              device_dst->device_type == ARROW_DEVICE_CUDA_HOST) {
-    // TODO: Synchronize device_src?
     memcpy((void*)dst.data.data, src.data.data, (size_t)src.size_bytes);
 
   } else if (device_src->device_type == ARROW_DEVICE_CUDA_HOST &&
              device_dst->device_type == ARROW_DEVICE_CUDA_HOST) {
-    // TODO: Synchronize device_src?
     memcpy((void*)dst.data.data, src.data.data, (size_t)src.size_bytes);
 
   } else if (device_src->device_type == ARROW_DEVICE_CUDA_HOST &&
              device_dst->device_type == ARROW_DEVICE_CPU) {
-    // TODO: Synchronize device_src?
     memcpy((void*)dst.data.data, src.data.data, (size_t)src.size_bytes);
 
   } else {
--- a/extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
+++ b/extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
@@ -53,6 +53,19 @@ static inline void ArrowDeviceCudaSetError(CUresult err, const char* op,
   _NANOARROW_CUDA_RETURN_NOT_OK_IMPL(_NANOARROW_MAKE_NAME(cuda_err_, __COUNTER__), EXPR, \
                                      OP, ERROR)
 
+#if defined(NANOARROW_DEBUG)
+#define _NANOARROW_CUDA_ASSERT_OK_IMPL(NAME, EXPR, EXPR_STR)           \
+  do {                                                                 \
+    const CUresult NAME = (EXPR);                                      \
+    if (NAME != CUDA_SUCCESS) NANOARROW_PRINT_AND_DIE(NAME, EXPR_STR); \
+  } while (0)
+#define NANOARROW_CUDA_ASSERT_OK(EXPR)                                                   \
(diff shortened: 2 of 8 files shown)


extensions/nanoarrow_device/CMakeLists.txt
```diff
@@ -119,7 +119,7 @@ else()
   if(NANOARROW_DEVICE_WITH_CUDA)
     find_package(CUDAToolkit REQUIRED)
     set(NANOARROW_DEVICE_SOURCES_CUDA src/nanoarrow/nanoarrow_device_cuda.c)
-    set(NANOARROW_DEVICE_LIBS_CUDA CUDA::cudart_static)
+    set(NANOARROW_DEVICE_LIBS_CUDA CUDA::cuda_driver)
     set(NANOARROW_DEVICE_DEFS_CUDA "NANOARROW_DEVICE_WITH_CUDA")
   endif()
```
zeroshade, line 124: is there anything on the documentation that needs to be updated accordingly (lists of dependencies, description of options, etc.)?

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -32,67 +71,63 @@ static void ArrowDeviceCudaDeallocator(struct ArrowBufferAllocator* allocator,
...
-      cudaFree(allocator_private->allocated_ptr);
+      cuMemFree((CUdeviceptr)allocator_private->allocated_ptr);
       break;
     case ARROW_DEVICE_CUDA_HOST:
-      cudaFreeHost(allocator_private->allocated_ptr);
+      cuMemFreeHost(allocator_private->allocated_ptr);
       break;
     default:
       break;
   }
```
zeroshade, line 83: I assume we're punting on handling Async scenarios for the time being, right?

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -32,67 +71,63 @@ static void ArrowDeviceCudaDeallocator(struct ArrowBufferAllocator* allocator,
...
 
-  cudaSetDevice(prev_device);
```
zeroshade, line 51: Are we always guaranteed that this has already been called or that we know we're using the correct device?

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -32,67 +71,63 @@ static void ArrowDeviceCudaDeallocator(struct ArrowBufferAllocator* allocator,
...
 }
 
 static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
                                                     struct ArrowBuffer* buffer,
                                                     int64_t size_bytes) {
-  int prev_device = 0;
-  cudaError_t result = cudaGetDevice(&prev_device);
-  if (result != cudaSuccess) {
-    return EINVAL;
-  }
```
zeroshade, line 62: Does the context itself manage the current device id by pushing and popping it?

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -32,67 +71,63 @@ static void ArrowDeviceCudaDeallocator(struct ArrowBufferAllocator* allocator,
...
       return EINVAL;
   }
 
-  if (result != cudaSuccess) {
+  if (err != CUDA_SUCCESS) {
+    cuCtxPopCurrent(&unused);
     ArrowFree(allocator_private);
-    cudaSetDevice(prev_device);
-    return ENOMEM;
+    return EIO;
```
zeroshade, line 130: since this function only returns an error code, but doesn't allow populating an error, is it okay that we're swallowing this error here? Should you leave a TODO comment so that we can remember to improve this?

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -105,19 +140,19 @@ static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
...
-  cudaSetDevice(prev_device);
+  cuCtxPopCurrent(&unused);
   return NANOARROW_OK;
 }
 
 struct ArrowDeviceCudaArrayPrivate {
   struct ArrowArray parent;
-  cudaEvent_t sync_event;
+  CUevent cu_event;
 };
```
zeroshade, line 150: This will need to be exposed somehow so that a producer can get access to this in order to record it on a stream or otherwise manage and use the event so that a consumer can benefit. 

If we're not exposing this event anywhere yet (since you're creating it privately and not accepting a user provided event) then we should probably just leave it null for now and not bother trying to create and destroy an event until we are also exposing it

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -161,108 +190,142 @@ static ArrowErrorCode ArrowDeviceCudaArrayInit(struct ArrowDevice* device,
...
+                     (size_t)src.size_bytes),
+        "cuMemcpyDtoH", error);
 
   } else if (device_src->device_type == ARROW_DEVICE_CPU &&
              device_dst->device_type == ARROW_DEVICE_CUDA_HOST) {
-    NANOARROW_RETURN_NOT_OK(
-        ArrowDeviceCudaAllocateBuffer(device_dst, &tmp, src.size_bytes));
-    memcpy_kind = cudaMemcpyHostToHost;
+    // TODO: Synchronize device_src?
+    memcpy((void*)dst.data.data, src.data.data, (size_t)src.size_bytes);
```
zeroshade, line 263: synchronizing wouldn't be limited to the CPU/CUDA_HOST cases. if we need to synchronize, we'd need to synchronize for *all* cases. 

But as I mentioned in the comment above, since we don't expose the event currently, you'd create a deadlock if you try to synchronize since nothing can mark the event as recorded and completed. 

Also since the event is at the top level of the `ArrowDeviceArray` i'd say that if we *are* going to synchronize, we shouldn't synchronize at this level but rather above this on the call stack. And until we start using the cuMemcpyAsync or other Async calls, we don't need to bother attempting to manage synchronization yet. we can punt on that for now

zeroshade: You could potentially use `cuCtxSynchronize` though...?

zeroshade (commented):
overall this looks fine to me, just a bunch of nitpicks and questions.

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: That's next! https://github.com/apache/arrow-nanoarrow/issues/245

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: I believe so...it's created using the `CUdevice` and I stole this from Arrow C++ (the push/pop context always surrounds the `cuMemAlloc()` there). I also don't have a multi GPU system to test on 🙂

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: Yes, device setting is now managed with the push/pop context (the notion of a "current device" is only available in the runtime API and I don't think we have another option).

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: That's a great point (synchronization of the source must have happened before this function is called). I was/am worried that `cudaMemcpy()` might have been flushing something from the device to the page-locked memory that a straight `memcpy()` wouldn't be doing. I'll look into `cuCtxSynchronize()` to see if that's doing what I think it is (or whether it should be called before any of this happens anyway).

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: This can now be specified from `ArrowDeviceArrayInit()`! I have a feeling I will be needing it shortly in the async buffer copying.

extensions/nanoarrow_device/src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: I looked into this and I am almost positive that calling `cuCtxSynchronize()` before memcpy is the right thing to do; however, it should also be done *once* before copying a bunch of buffers (as you suggested). We really just need to test this properly (which will happen naturally when we implement async buffer copying, since that should result in an not-fully-synchronized array with a non-null sync event to properly test on).

PR #490, opened 2024-05-24
chore(ci): Add CUDA runner for testing device extension

paleolimbot:
This PR adds the INFRA-provided runner `["self-hosted", "cuda"]` so that PRs that modify CUDA code get better CI feedback (before this, the extension pretty much was only ever built on my local setup).

The self-hosted runner is running an older linux than the usual `ubuntu-latest`, so a few modifications were needed to reflect compiler warning policies. The cache key we were using for Arrow C++ is also no longer sufficient for the device workflow because a cache restore results in an unusable Arrow C++ static library on the self-hosted runner. This shouldn't be necessary for much longer (just until Arrow C++-dependent tests are factored out and run separately).

### DIFF
--- a/src/nanoarrow/device/cuda_test.cc
+++ b/src/nanoarrow/device/cuda_test.cc
@@ -21,7 +21,6 @@
 #include <tuple>
 
 #include "nanoarrow/nanoarrow_device.hpp"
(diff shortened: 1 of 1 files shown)


lidavidm:
Might need INFRA to enable the CUDA runners on this repo? Not sure who controls that

Issue #495, opened 2024-05-28
Export symbols from C library when compiling as a shared library on Windows


paleolimbot:
We will have to check that this is actually the problem, but it may be that the lack of any exported symbols causes a problem when creating a shared library on Windows ( https://github.com/mesonbuild/wrapdb/pull/1536#issuecomment-2135839166 ). adbc.h has:

```
// Storage class macros for Windows
// Allow overriding/aliasing with application-defined macros
#if !defined(ADBC_EXPORT)
#if defined(_WIN32)
#if defined(ADBC_EXPORTING)
#define ADBC_EXPORT __declspec(dllexport)
#else
#define ADBC_EXPORT __declspec(dllimport)
#endif  // defined(ADBC_EXPORTING)
#else
#define ADBC_EXPORT
#endif  // defined(_WIN32)
#endif  // !defined(ADBC_EXPORT)
```

https://github.com/apache/arrow-adbc/blob/main/adbc.h#L152-L164

(Although I think this is slightly incorrect in that `_WIN32` should maybe be `_MSC_VER` to better support gcc on Windows via msys2).

lidavidm:
> (Although I think this is slightly incorrect in that `_WIN32` should maybe be `_MSC_VER` to better support gcc on Windows via msys2).

Ah, probably. I derived this from some (by now old) copy of the Arrow-C++ equivalents.

WillAyd:
> (Although I think this is slightly incorrect in that `_WIN32` should maybe be `_MSC_VER` to better support gcc on Windows via msys2).

May also be worth checking for __CYGWIN__

https://stackoverflow.com/a/9026009/621736

WillAyd:
Also https://www.msys2.org/wiki/Porting/#platform-checks

WillAyd:
Not to try and expand the scope too much but this gcc article might even help get these to work with GNU compilers:

https://gcc.gnu.org/wiki/Visibility

PR #496, opened 2024-05-28
fix: Force static library build on Windows when building with Meson

WillAyd:
May not necessarily fix https://github.com/apache/arrow-nanoarrow/issues/495 but reasonably defers trying to solve

Basically the suggestion of the great @eli-schwartz


paleolimbot (approved):
Thanks!

PR #499, opened 2024-05-30
feat: add support for appending values to half float `ArrowArray`

cocoa-xu:
Hi, thanks for this great library! I found that `ArrowArrayAppendXX` functions won't append a value to a half float `ArrowArray` while I was playing with this library. So I decided to add this support, which I believe it should make this library easier to use.

The conversion function is adapted from Arrow Go, https://github.com/apache/arrow/blob/main/go/arrow/float16/float16.go, and the tests cases are taken from https://github.com/apache/arrow/blob/main/go/arrow/float16/float16_test.go.

Although I'm not sure if I should also update the files in `dist` though (I guess it might be automated when doing a release but I didn't check). I'll be happy to update them if these files are not updated automatically, and I'm happy to do any other changes. :)

### DIFF
--- a/src/nanoarrow/array_test.cc
+++ b/src/nanoarrow/array_test.cc
@@ -816,7 +816,8 @@ TEST(ArrayTest, ArrayTestAppendToHalfFloatArray) {
   EXPECT_EQ(data_buffer[2], 0);
   EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[3]), 3.0);
   EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[4]), 3.138672);
-  EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[5]), std::numeric_limits<float>::max());
+  EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[5]),
+                  std::numeric_limits<float>::max());
   EXPECT_TRUE(std::isnan(ArrowHalfFloatToFloat(data_buffer[6])));
   EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[7]), INFINITY);
   EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[8]), -INFINITY);
--- a/src/nanoarrow/array_test.cc
+++ b/src/nanoarrow/array_test.cc
@@ -811,17 +811,17 @@ TEST(ArrayTest, ArrayTestAppendToHalfFloatArray) {
   auto data_buffer = reinterpret_cast<const uint16_t*>(array.buffers[1]);
   EXPECT_EQ(validity_buffer[0], 0b11111001);
   EXPECT_EQ(validity_buffer[1], 0b00000111);
-  EXPECT_EQ(data_buffer[0], 15360);
+  EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[0]), 1);
   EXPECT_EQ(data_buffer[1], 0);
   EXPECT_EQ(data_buffer[2], 0);
-  EXPECT_EQ(data_buffer[3], 16896);
-  EXPECT_EQ(data_buffer[4], 16967);
-  EXPECT_EQ(data_buffer[5], 31744);
-  EXPECT_EQ(data_buffer[6], 32256);
-  EXPECT_EQ(data_buffer[7], 31744);
-  EXPECT_EQ(data_buffer[8], 64512);
-  EXPECT_EQ(data_buffer[9], 48128);
-  EXPECT_EQ(data_buffer[10], 0);
+  EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[3]), 3.0);
+  EXPECT_FLOAT_EQ(ArrowHalfFloatToFloat(data_buffer[4]), 3.138672);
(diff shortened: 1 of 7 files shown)


src/nanoarrow/array_test.cc
```diff
@@ -787,6 +787,43 @@ TEST(ArrayTest, ArrayTestAppendToFloatArray) {
...
+  EXPECT_EQ(data_buffer[1], 0);
+  EXPECT_EQ(data_buffer[2], 0);
+  EXPECT_EQ(data_buffer[3], 16896);
+  EXPECT_EQ(data_buffer[4], 16967);
+  EXPECT_EQ(data_buffer[5], 31744);
+  EXPECT_EQ(data_buffer[6], 32256);
+  EXPECT_EQ(data_buffer[7], 31744);
+  EXPECT_EQ(data_buffer[8], 64512);
+  EXPECT_EQ(data_buffer[9], 48128);
+  EXPECT_EQ(data_buffer[10], 0);
```
paleolimbot, line 824: Can we add a note about how to generate these values? (Perhaps numpy can do this?)

paleolimbot (commented):
This is great! Thank you! Just a few notes, and I think you may need to check that this will work on big endian (which you can do locally using `export NANOARROW_ARCH=s390x && docker compose run --rm verify`).

No need to add it in this PR if you'd prefer not to, but a conversion in the other direction would also be very useful (and perhaps make it easier to write the tests).

src/nanoarrow/utils_test.cc
```diff
@@ -543,6 +543,36 @@ TEST(DecimalTest, DecimalRoundtripBitshiftTest) {
   ArrowBufferReset(&buffer);
 }
 
+// test case adapted from
+// https://github.com/apache/arrow/blob/main/go/arrow/float16/float16_test.go
+TEST(HalfFloatTest, FloatToHalfFloat) {
```
benibus: We should specifically test that NaNs are preserved as well. It's a simple check, fortunately: https://github.com/apache/arrow/blob/6800be9331d88024bf550c77865a06c592a22699/go/arrow/float16/float16.go#L176

src/nanoarrow/utils_test.cc
```diff
@@ -543,6 +543,36 @@ TEST(DecimalTest, DecimalRoundtripBitshiftTest) {
...
+    0x3c00,
+    0x4000,
+    0xc000,
+    0x0000,
+    0x5b8f,
+    0xdb8f,
+    0x48c8,
+    0xc8c8,
+  };
+  float cases_float[] = {
```
benibus, line 559: A few more test cases that would be easy to add here:
- `-0` -> `0x8000`
- `+inf` -> `0x7c00`
- `-inf` -> `0xfc00`

benibus (commented):
Thanks for this!

src/nanoarrow/array_test.cc
cocoa-xu: I found that it wasn't that easy to do a bit cast in numpy (compare to C), and since the users will probably also need to convert half floats to floats, I decide to add an `ArrowHalfFloatToFloat` function. 

In this way, we can simply do a round trip test between half floats <=> floats and no need to worry about how these weird numbers were generated. :)

cocoa-xu:
@paleolimbot @benibus Many thanks for the code review!

src/nanoarrow/array_test.cc
paleolimbot: Having a one-way test in at least one direction is also probably a good idea (in the unlikely event that the conversion is wrong in both directions). I think Python can do this:

```python
import struct
import math

float_vals = [-12.3, 0, 12.3, math.nan]
packed = b"".join(struct.pack("e", val) for val in float_vals)
list(memoryview(packed).cast("H"))
#> [51750, 0, 18982, 32256]
```

paleolimbot: (Although the link to the test where you got these is also more than enough 🙂 )

cocoa-xu: > Having a one-way test in at least one direction is also probably a good idea (in the unlikely event that the conversion is wrong in both directions). I think Python can do this:
> 
> ```python
> import struct
> import math
> 
> float_vals = [-12.3, 0, 12.3, math.nan]
> packed = b"".join(struct.pack("e", val) for val in float_vals)
> list(memoryview(packed).cast("H"))
> #> [51750, 0, 18982, 32256]
> ```

Ah thanks for the numpy code! Shall I also add these one-way tests? It should be easy though/

benibus: Also worth noting that, unlike arrow-go, numpy does some rounding in its half-float conversions. Some values may differ if you attempt to generate test cases with it.

cocoa-xu: Oh I see. Maybe we can use the Go implementation to generate some test cases if needed? (And in the most unlikely event that the Go implementation for float16 is wrong, we still would have the same encoding and decoding function and compatible with arrow-go.)

cocoa-xu:
Hi, I've formatted the code and fixed the tests. Just let me know if we should add any generated test cases for these one-way conversions, and I'll happy to add a few cases!

paleolimbot (approved):
Sorry for the delay, I wanted to make sure I had time to check this out and run some tests locally. This looks great...thank you!

cocoa-xu:
> Sorry for the delay, I wanted to make sure I had time to check this out and run some tests locally.

Thank you very much for the kind help and code review!

Issue #478, opened 2024-05-19
Can't build array of type timestamp from iterable


aosingh:
Thanks to the Arrow community for developing this lightweight wrapper. 

I am planning to add support for Apache Arrow in one of the projects I am working on. The aim is to leverage nanoarrow to support exporting tabular data in arrow format. 

Users will have access to a function `to_arrow()`:

```python
import nanoarrow as na

def gen_name():
    for i in range(100):
        yield "John Doe"

def gen_age():
    for i in range(100):
        yield 34


def to_arrow():
    results = [na.c_array(gen_name(), na.string()), na.c_array(gen_age(), na.int64())]
    return results

```

Users of the library can optionally install `pyarrow` and `pandas` to work with the exported data. And the export works fine!

```python
import pyarrow as pa
parray = pa.Table.from_arrays(to_arrow(), names=["name", "age"])
print(parray.to_pandas())

```

```bash
        name  age
0   John Doe   34
1   John Doe   34
2   John Doe   34
3   John Doe   34
4   John Doe   34
..       ...  ...
95  John Doe   34
96  John Doe   34
97  John Doe   34
98  John Doe   34
99  John Doe   34

[100 rows x 2 columns]
```

Adding a third field  `timestamp` to the above list raises an error:

```python
def gen_timestamp():
    for i in range(100):
        yield datetime.datetime.now().timestamp()

result = [na.c_array(gen_name(), na.string()),
          na.c_array(gen_age(), na.int64()),
          na.c_array(gen_timestamp(), na.timestamp("s"))]

parray = pa.Table.from_arrays(result, names=["name", "age", "timestamp"])

print(parray.to_pandas())

``` 

Error:
```
Traceback (most recent call last):
  File "/Users/as/nanoarrow/simple.py", line 28, in <module>
    na.c_array(gen_timestamp(), na.timestamp("s"))]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/as/nanoarrow/arrow-nanoarrow/python/src/nanoarrow/c_array.py", line 131, in c_array
    raise ValueError(
ValueError: An error occurred whilst converting generator to nanoarrow.c_array: 
 Can't build array of type timestamp from iterable
```

I understand the source of error is the [mapping](https://github.com/apache/arrow-nanoarrow/blob/main/python/src/nanoarrow/c_array.py#L531C1-L552) maintained for each datatype. 

[...]

PR #502, opened 2024-05-31
feat(python) : Added support to create timestamp/date32/date64 Array from iterable

aosingh:
Addresses https://github.com/apache/arrow-nanoarrow/issues/478

Currently, if you create a `c_array` from an iterable of timestamps an error is raised

Code:
```python
from datetime import datetime

import nanoarrow as na
import pyarrow as pa


def gen_timestamp():
    for i in range(100):
        yield int(round(datetime.now().timestamp()))

timestamp_array = na.c_array(gen_timestamp(), na.timestamp("s"))

parray = pa.Table.from_arrays([timestamp_array], names=["timestamp"])

print(parray.to_pandas())
```
Error:
```
Traceback (most recent call last):
  File "/Users/as/nanoarrow_poc/.venv/lib/python3.12/site-packages/nanoarrow/c_array.py", line 128, in c_array
    builder = builder_cls(schema)
              ^^^^^^^^^^^^^^^^^^^
  File "/Users/as/nanoarrow_poc/.venv/lib/python3.12/site-packages/nanoarrow/c_array.py", line 457, in __init__
    raise ValueError(
ValueError: Can't build array of type timestamp from iterable
```

Expected output:
```
             timestamp
0  2024-05-31 18:43:10
1  2024-05-31 18:43:10
2  2024-05-31 18:43:10
3  2024-05-31 18:43:10
4  2024-05-31 18:43:10
..                 ...
95 2024-05-31 18:43:10
96 2024-05-31 18:43:10
97 2024-05-31 18:43:10
98 2024-05-31 18:43:10
99 2024-05-31 18:43:10

[100 rows x 1 columns]
```

Benefits
- Pyarrow Table and Pandas Dataframe can use the corresponding nanoarrow Array Type to cast the values when using `to_pandas`


python/src/nanoarrow/_lib.pyx
```diff
@@ -483,13 +483,13 @@ cdef int c_format_from_arrow_type(ArrowType type_id, int element_size_bits, size
     elif type_id == NANOARROW_TYPE_UINT16:
         format_const = "H"
         element_size_bits_calc = 16
-    elif type_id in (NANOARROW_TYPE_INT32, NANOARROW_TYPE_INTERVAL_MONTHS):
+    elif type_id in (NANOARROW_TYPE_INT32, NANOARROW_TYPE_INTERVAL_MONTHS, NANOARROW_TYPE_DATE32):
```
paleolimbot, line 486: We should add a comment here to clarify what this function is doing, but I believe that `NANOARROW_TYPE_DATE32` does not *quite* belong here (arguably `NANOARROW_TYPE_INTERVAL_MONTHS` should not be here either, which has come up in previous reviews, and may change). See below for an alternative (I think adding a property to the schema view might be a better fit).

python/src/nanoarrow/_lib.pyx
```diff
@@ -1069,6 +1069,12 @@ cdef class CSchemaView:
         NANOARROW_TYPE_SPARSE_UNION
     )
 
+    _buffer_protocol_supported_types = (
+        NANOARROW_TYPE_DATE32,
+        NANOARROW_TYPE_DATE64,
+        NANOARROW_TYPE_TIMESTAMP
+    )
```
paleolimbot, line 1076: How about adding a `CSchemaView.storage_buffer_format`? I think these are perhaps more accurately types whose storage can be represented by the Python buffer protocol (however, the types themselves do not have a dedicated buffer protocol format string).

python/tests/test_array.py
```diff
@@ -354,3 +354,86 @@ def test_array_inspect(capsys):
     array.inspect()
     captured = capsys.readouterr()
     assert captured.out.startswith("<ArrowArray struct<col0: int32")
+
+
+def test_array_using_struct(capsys):
```
paleolimbot, line 359: I am a little unclear what this is trying to test...would it be sufficient to test `na.c_array([d1, d2], na.timestamp('ms'))` and its corresponding `to_pysequence()`? (Same question for the tests below)

python/src/nanoarrow/c_array.py
```diff
@@ -549,6 +549,9 @@ def _append_using_buffer_builder(self, obj: Iterable) -> None:
     CArrowType.UINT64: "_append_using_array",
     CArrowType.FLOAT: "_append_using_array",
     CArrowType.DOUBLE: "_append_using_array",
+    CArrowType.TIMESTAMP: "_append_using_array",
+    CArrowType.DATE32: "_append_using_array",
+    CArrowType.DATE64: "_append_using_array"
```
paleolimbot, line 554: Does `CArrowType.DURATION` belong in this list?

python/tests/test_c_array.py
```diff
@@ -514,3 +516,36 @@ def test_c_array_from_buffers_validation():
...
+
+def test_c_array_timestamp_milliseconds():    
+    d1 = int(round(datetime(1970, 1, 1).timestamp() * 1e3))
+    d2 = int(round(datetime(1985, 12, 31).timestamp() * 1e3))
+    d3 = int(round(datetime(2005, 3, 4).timestamp() * 1e3))
+    c_array = na.c_array([d1, d2, d3], na.timestamp('ms'))
+    c_array_from_c_array = na.c_array(c_array)
+    assert c_array_from_c_array.length == c_array.length
+    assert c_array_from_c_array.buffers == c_array.buffers
+    assert list(c_array.view().buffer(1)) == [28800e3, 504864000e3, 1109923200e3]
```
paleolimbot, line 541: Are the numbers here `[d1, d2, d3]`? (and in the test below)

python/tests/test_c_array.py
```diff
@@ -514,3 +516,36 @@ def test_c_array_from_buffers_validation():
...
+
+def test_c_array_timestamp_microseconds():    
+    d1 = int(round(datetime(1970, 1, 1).timestamp() * 1e6))
+    d2 = int(round(datetime(1985, 12, 31).timestamp() * 1e6))
+    d3 = int(round(datetime(2005, 3, 4).timestamp() * 1e6))
+    c_array = na.c_array([d1, d2, d3], na.timestamp('us'))
+    c_array_from_c_array = na.c_array(c_array)
+    assert c_array_from_c_array.length == c_array.length
+    assert c_array_from_c_array.buffers == c_array.buffers
+    assert list(c_array.view().buffer(1)) == [28800e6, 504864000e6, 1109923200e6]
```
paleolimbot, line 551: Are nanoseconds supported by this code path?

paleolimbot (commented):
Thank you for taking this on! Just a few initial comments 🙂 

(Sorry about the CI not running for new contributors...I will try to stay up-to-date and keep the workflows running but feel free to ping me if I forget)

aosingh:
Thank you, I will go through the review comments and also check why some unit tests failed

python/src/nanoarrow/_lib.pyx
aosingh: Ok. will this new property be in addition to the existing `CSchemaView.buffer_format` ?  I am trying to understand how will both these properties be used.

python/src/nanoarrow/c_array.py
aosingh: Yes, I think we can include Duration too. Let me add it and some tests.

python/tests/test_c_array.py
aosingh: If I understand correctly Timestamp is a 64-bit signed integer.  Time unit can be either of four units: seconds, milliseconds, microseconds or nanoseconds,  

Did you mean if we have to do something different for nanoseconds ?

aosingh:
@paleolimbot 

Could you please help with the GitHub CI workflow ?  I wanted to verify the latest changes and unit tests. Thanks.

python/src/nanoarrow/_lib.pyx
paleolimbot: It's a great question (and should be better documented): `buffer_format` is for when a Python buffer protocol format string/struct module format string is sufficient to describe a type. `storage_buffer_format` would also include types whose underlying data has a valid format string (but that have extra logical type interpretation that is not captured solely by the format string).

python/tests/test_c_array.py
paleolimbot: I think that I just didn't spot a test for nanoseconds on the first round (but I may have missed it).

aosingh: I see, correct, I will add a testcase for nanoseconds.

aosingh:
There are a few style guide check failures and 1 unit test failed. I am checking this.

aosingh:
Thanks, `pre-commit run --all-files` formatted the code as per style guide. I have also modified a unit testcase to use UTC timezone in timestamps.

python/tests/test_array.py
aosingh: I have modified this testcase to compare the timestamp array values with its corresponding` to_pysequence()`

I guess, in the remaining 2 tests, I wanted to verify if this fix works for the problem which is the motivation behind this PR. i.e. successfully convert to Pyarrow Table and Pandas Dataframe. (Maybe it is not needed?)

python/src/nanoarrow/_lib.pyx
aosingh: @paleolimbot thanks, I will spend some time thinking about `storage_buffer_format` and how we can incorporate the change.

paleolimbot (commented):
Thank you! (And thank for bearing with me regarding the CI 😬 only running when I remember to click).

I am also happy to take a stab at the "storage_buffer_format" thing (or defer to you if you're into it!)...you've done all the hard work here and that particular detail is a low-level thing whose role I am the first to admit is unclear in our internals.

python/tests/test_array.py
```diff
@@ -354,3 +355,87 @@ def test_array_inspect(capsys):
...
+        {"creation_timestamp": datetime(2005, 3, 4, 0, 0)},
+    ]
+    array.inspect()
+    captured = capsys.readouterr()
+    assert captured.out.startswith(
+        "<ArrowArray struct<creation_timestamp: timestamp('ms'"
+    )
+
+
+def test_pyarrow_table_using_array():
```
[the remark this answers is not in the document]
aosingh: Understood, makes sense. I will remove these 2 for now.

aosingh:
@paleolimbot 

Please review the updates. I have simplified unit tests. Hopefully, the test cases are clear now. Let me know your feedback.

python/tests/test_c_buffer.py
```diff
@@ -259,8 +260,8 @@ def test_c_buffer_from_iterable():
 
     # An Arrow type whose storage type is not the same as its top-level
     # type will error.
-    with pytest.raises(ValueError, match="Can't create buffer"):
-        na.c_buffer([1, 2, 3], na.date32())
+    # with pytest.raises(ValueError, match="Can't create buffer"):
+    #     na.c_buffer([1, 2, 3], na.duration("s"))
```
paleolimbot, line 264:
```suggestion
    with pytest.raises(ValueError, match="Can't create buffer"):
        na.c_buffer([1, 2, 3], na.dictionary(na.int32(), na.string()))
```

(I think this should trigger the intended code path without requiring the new construction features!)

python/tests/test_c_array.py
```diff
@@ -514,3 +516,71 @@ def test_c_array_from_buffers_validation():
             validation_level=validation_level,
         )
         assert c_array.length == 2
+
+
+def test_c_array_timestamp_seconds():
+    d1 = int(round(datetime(1970, 1, 1).timestamp()))
+    d2 = int(round(datetime(1985, 12, 31).timestamp()))
+    d3 = int(round(datetime(2005, 3, 4).timestamp()))
+    c_array = na.c_array([d1, d2, d3], na.timestamp("s"))
```
paleolimbot, line 525: In addition to `na.c_array([d1, d2, d3], na.timestamp("s"))`, I believe that creating this from a pybuffer (e.g., numpy) should work too. Perhaps we should also check `na.c_array(array.array("i", [d1, d2, d3]), na.timestamp("s"))`?

python/tests/test_c_buffer.py
```diff
@@ -362,3 +363,47 @@ def test_c_buffer_bitmap_from_iterable():
     builder.write_elements([True, False])
     with pytest.raises(NotImplementedError, match="Append to bitmap"):
         builder.write_elements([True])
+
+
+def test_c_buffer_from_timestamp_iterable():
```
paleolimbot, line 368: I am not sure that one should be able to do `na.c_buffer([1, 2, 3], na.timestamp("ms"))` without getting an error (a timestamp is a valid *array* type but not really a valid *buffer* type). I think that the `buffer_format` vs. `storage_buffer_format` change will take care of this.

paleolimbot (commented):
Thank you for these changes and for bearing with me!

I do think that a separate `storage_buffer_format` will help clarify the distinctions between types that losslessly communicated via a pybuffer format string and the types whose physical representation happens to be a single buffer. That would require updating the logic in `ArrayFromIterableBuilder.__init__` to ensure that those types are routed through the `_append_using_array` method but I think it would be worth it to minimize the chance that type information is lost by accident 🙂 .

aosingh:
I have added `CSchemaView.storage_buffer_format` and updated unit tests.  I will further review it tomorrow.

python/src/nanoarrow/_lib.pyx
```diff
@@ -1095,6 +1095,15 @@ cdef class CSchemaView:
     def storage_type_id(self):
         return self._schema_view.storage_type
 
+    @property
+    def storage_buffer_format(self):
+        if self._schema_view.type == NANOARROW_TYPE_DATE32:
```
paleolimbot, line 1100:
```suggestion
        if self.buffer_format is not None:
            return self.buffer_format
        elif self._schema_view.type == NANOARROW_TYPE_DATE32:
```

Possibly?

aosingh:
@paleolimbot 

I have updated the PR based on the review comments. 

To summarize:

- Creation of c_array depends on schema's `storage_buffer_format`
- The property `storage_buffer_format` either returns the existing `buffer_format`  or the underlying buffer format corresponding to the  logical data type
- Added a couple of test cases to create c_array of timestamps from PyBuffer

Please review and let me know your feedback

paleolimbot (approved):
Thank you for bearing with me on this!

I am still not sure that `na.c_buffer([d1, d2, d3], na.timestamp("ms"))` should run without error. I'm happy to look into that in a later PR if you'd prefer or feel free to tackle it here (I don't know remember off the top of my head why it works).

aosingh:
@paleolimbot 

Sure, no problem, thank you for your patience. 

`na.c_buffer([d1, d2, d3], na.timestamp("ms"))` raises ValueError. The unit tests are modified to verify that. 

```python
with pytest.raises(ValueError):
     na.c_buffer([d1, d2, d3], na.timestamp("ms"))
```

is this not expected ?

aosingh:
@paleolimbot 

I see some unit tests for Windows has failed after merge. Let me try to understand the reason

PR #501, opened 2024-05-31
feat: Add float16 support for `ArrowArrayViewGet{Double,Int,UInt}Unsafe()`

cocoa-xu:
Hi this PR adds float16 support for `ArrowArrayViewGet{Double,Int,UInt}Unsafe()`. Related issue: #500

PR #503, opened 2024-06-01
test: test with the `HalfFloatType` from arrow

cocoa-xu:
Hi this PR adds an extra test case that tests `ArrowFloatToHalfFloat` and `ArrowArrayViewGet{Double,Int,UInt}Unsafe` with the `HalfFloatType` from arrow.

https://github.com/apache/arrow/blob/255dbf990c3d3e5fb1270a2a11efe0af2be195ab/cpp/src/arrow/type.h#L704-L713

And sorry that I didn't know there's a `HalfFloatType` in arrow-cpp when I was doing #501 🥲.

Sadly that we cannot simply append `TestGetFromNumericArrayView<HalfFloatType>();` at the end of the `ArrayViewTestGetNumeric` test suite because we have to convert floats to half-floats using `ArrowFloatToHalfFloat` before calling `builder.Append` (otherwise we'll get weird values back in the subsequent `ArrowArrayViewGet{Double,Int,UInt}Unsafe` calls).

### DIFF
--- a/src/nanoarrow/array_test.cc
+++ b/src/nanoarrow/array_test.cc
@@ -2493,6 +2493,76 @@ void TestGetFromNumericArrayView() {
   ArrowSchemaRelease(&schema);
 }
 
+void TestGetFromNumericArrayViewWithHalfFloatType() {
+  struct ArrowArray array;
+  struct ArrowSchema schema;
+  struct ArrowArrayView array_view;
+  struct ArrowError error;
+
+  auto type = TypeTraits<HalfFloatType>::type_singleton();
+
+  // Array with nulls
+  auto builder = NumericBuilder<HalfFloatType>();
+  ARROW_EXPECT_OK(builder.Append(ArrowFloatToHalfFloat(1.0)));
+  ARROW_EXPECT_OK(builder.AppendNulls(2));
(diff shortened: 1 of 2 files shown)


paleolimbot (approved):
Thank you! 

I think the repetition is OK...we have a lot of repetition in our tests and while it would be good to go through and reduce that where possible, I prefer getting the test coverage there first.

There is also a few places this could be enabled in the integration tests if you have the bandwidth (no pressure!). (This can also be another PR if you prefer)

https://github.com/apache/arrow-nanoarrow/blob/5dc4aee07b4caa1a378951105658c7e93722ac72/src/nanoarrow/nanoarrow_testing.hpp#L807-L819

https://github.com/apache/arrow-nanoarrow/blob/5dc4aee07b4caa1a378951105658c7e93722ac72/src/nanoarrow/nanoarrow_testing.hpp#L807-L819

https://github.com/apache/arrow-nanoarrow/blob/5dc4aee07b4caa1a378951105658c7e93722ac72/src/nanoarrow/nanoarrow_testing.hpp#L2149-L2152

cocoa-xu:
> and reduce that where possible

Not a problem, I added a helper function and it should look better now. :)

> There is also a few places this could be enabled in the integration tests if you have the bandwidth (no pressure!)

It's done now ;)

paleolimbot (approved):
Thank you!

PR #504, opened 2024-06-03
refactor!: More idiomatic meson usage

WillAyd:
Two main changes here:

1. Changed options like `NANOARROW_BUILD_TESTS` to just `tests`
2. Included wrap files into repository, rather than requiring user to install

Point 1 was inspired by looking at other projects in the wrap database. Note that if multiple projects were being compiled at once and there was some ambiguity between options, you can namespace providing them on the command line by using the form `<project>:<option>`, ex: `meson configure -Darrow:tests=true`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ CMakeUserPresets.json
 __pycache__
 
 # meson subprojects
-subprojects
+subprojects/*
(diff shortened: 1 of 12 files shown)


paleolimbot (approved):
Thanks!

PR #507, opened 2024-06-04
feat: Add support for run-end encoded array

cocoa-xu:
Hi this PR tries to add support for run-end encoded array based on the arrow spec here, https://arrow.apache.org/docs/format/Columnar.html#run-end-encoded-layout.

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -69,6 +69,7 @@ static ArrowErrorCode ArrowArraySetStorageType(struct ArrowArray* array,
   switch (storage_type) {
     case NANOARROW_TYPE_UNINITIALIZED:
     case NANOARROW_TYPE_NA:
+    case NANOARROW_TYPE_RUN_END_ENCODED:
       array->n_buffers = 0;
       break;
 
@@ -811,6 +812,15 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
                       (long)array_view->n_children);
         return EINVAL;
       }
+      break;
+    case NANOARROW_TYPE_RUN_END_ENCODED:
+      if (array_view->n_children != 2) {
+        ArrowErrorSet(error, "Expected 2 child of %s array but found %ld child arrays",
+                      ArrowTypeString(array_view->storage_type),
+                      (long)array_view->n_children);
+        return EINVAL;
+      }
+      break;
     default:
       break;
   }
@@ -843,6 +853,22 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
                       "Expected child of fixed_size_list array to have length >= %ld but "
                       "found array with length %ld",
                       (long)child_min_length, (long)array_view->children[0]->length);
+        return EINVAL;
+      }
+      break;
+
+    case NANOARROW_TYPE_RUN_END_ENCODED:
+      if (array_view->children[0]->null_count != 0) {
+        ArrowErrorSet(error, "Run Ends cannot be null but found null_count value %ld",
+                      (long)array_view->children[0]->null_count);
+        return EINVAL;
+      }
+      if (array_view->children[0]->length < array_view->children[1]->length) {
+        ArrowErrorSet(
+            error,
+            "Expected the 2 children of run-end encoded array to have equal length "
+            "but found mismatched lengths: run_ends->length=%ld, values->length=%ld",
+            (long)array_view->children[0]->length, (long)array_view->children[1]->length);
         return EINVAL;
       }
       break;
@@ -995,6 +1021,31 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
         }
       }
       break;
+
+    case NANOARROW_TYPE_RUN_END_ENCODED: {
+      struct ArrowArrayView* run_ends_view;
+      run_ends_view = array_view->children[0];
+      if (run_ends_view->null_count != 0) {
+        ArrowErrorSet(error, "Run Ends cannot be null but found null_count value %ld",
+                      (long)run_ends_view->null_count);
+        return EINVAL;
+      }
+      for (int64_t i = 0, prev_run_end = 0; i < run_ends_view->length; i++) {
+        int64_t run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, i);
+        if (run_end < 0 || run_end < prev_run_end) {
+          return EINVAL;
+        }
+        prev_run_end = run_end;
+        if (i == run_ends_view->length - 1 && run_end != array_view->length) {
+          ArrowErrorSet(error,
+                        "Run End value %ld at index %ld exceeds the logical "
+                        "length of the run-end encoded array %ld",
+                        (long)run_end, (long)i, (long)array_view->length);
+          return EINVAL;
+        }
+      }
+      break;
+    }
     default:
       break;
   }
@@ -1160,6 +1211,40 @@ static int ArrowArrayViewValidateFull(struct ArrowArrayView* array_view,
             (long)i, (int)child_id, (long)child_length, (long)offset);
         return EINVAL;
       }
+    }
+  }
+
+  if (array_view->storage_type == NANOARROW_TYPE_RUN_END_ENCODED) {
+    struct ArrowArrayView* run_ends_view = array_view->children[0];
+    switch (run_ends_view->storage_type) {
+      case NANOARROW_TYPE_INT16:
+      case NANOARROW_TYPE_INT32:
+      case NANOARROW_TYPE_INT64:
+        break;
+      default:
+        ArrowErrorSet(
+            error,
+            "Run-end encoded array only supports INT16, INT32 or INT64 run-ends "
+            "but found run-ends type %s",
+            ArrowTypeString(run_ends_view->storage_type));
+        return EINVAL;
+    }
+    int64_t prev_run_end = 0;
+    for (int64_t i = 0; i < run_ends_view->length; i++) {
+      int64_t run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, i);
+      if (run_end < 0 || run_end < prev_run_end) {
+        return EINVAL;
+      }
+      prev_run_end = run_end;
+    }
+    if (prev_run_end != array_view->length) {
+      ArrowErrorSet(error,
+                    "The last run end value of a run-end encoded array must be equal to "
+                    "the logical length"
+                    "of the values array but found the last run end value %ld while the "
+                    "logical length is %ld",
+                    (long)prev_run_end, (long)array_view->length);
+      return EINVAL;
     }
   }
 
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -815,9 +815,9 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
       break;
     case NANOARROW_TYPE_RUN_END_ENCODED:
       if (array_view->n_children != 2) {
-        ArrowErrorSet(error, "Expected 2 children of %s array but found %ld child arrays",
-                      ArrowTypeString(array_view->storage_type),
-                      (long)array_view->n_children);
+        ArrowErrorSet(
+            error, "Expected 2 children for %s array but found %ld child arrays",
+            ArrowTypeString(array_view->storage_type), (long)array_view->n_children);
         return EINVAL;
       }
       break;
@@ -907,7 +907,7 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
       if (run_ends_view->length == 0 && values_view->length != 0) {
         ArrowErrorSet(error,
                       "Run-end encoded array has zero length %ld, but values array has "
-                      "non-zero length ",
+                      "non-zero length",
                       (long)values_view->length);
         return EINVAL;
       }
@@ -1257,7 +1257,7 @@ static int ArrowArrayViewValidateFull(struct ArrowArrayView* array_view,
       last_run_end = run_end;
     }
     last_run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, run_ends_view->length - 1);
-    if (last_run_end < array_view->offset + array_view->length) {
+    if (last_run_end < (array_view->offset + array_view->length)) {
       ArrowErrorSet(error,
                     "Last run end is %ld but it should >= %ld (offset: %ld, length: %ld)",
                     (long)last_run_end, (long)(array_view->offset + array_view->length),
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -858,17 +858,70 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
       break;
 
     case NANOARROW_TYPE_RUN_END_ENCODED:
-      if (array_view->children[0]->null_count != 0) {
-        ArrowErrorSet(error, "Run Ends cannot be null but found null_count value %ld",
-                      (long)array_view->children[0]->null_count);
+      if (array_view->n_children != 2) {
+        ArrowErrorSet(error,
+                      "Expected 2 children for run-end encoded array but found %ld",
+                      (long)array_view->n_children);
         return EINVAL;
       }
-      if (array_view->children[0]->length < array_view->children[1]->length) {
+      struct ArrowArrayView* run_ends_view;
+      struct ArrowArrayView* values_view;
+      run_ends_view = array_view->children[0];
+      values_view = array_view->children[1];
+      if (run_ends_view == NULL) {
+        ArrowErrorSet(error, "Run ends array is null pointer");
+        return EINVAL;
+      }
+      if (values_view == NULL) {
+        ArrowErrorSet(error, "Values array is null pointer");
+        return EINVAL;
+      }
+      uint64_t max_length;
+      switch (run_ends_view->storage_type) {
+        case NANOARROW_TYPE_INT16:
+          max_length = INT16_MAX;
+          break;
+        case NANOARROW_TYPE_INT32:
+          max_length = INT32_MAX;
+          break;
+        case NANOARROW_TYPE_INT64:
+          max_length = INT64_MAX;
+          break;
+        default:
+          ArrowErrorSet(
+              error,
+              "Run-end encoded array only supports INT16, INT32 or INT64 run-ends "
+              "but found run-ends type %s",
+              ArrowTypeString(run_ends_view->storage_type));
+          return EINVAL;
+      }
+      if ((uint64_t)array_view->offset + (uint64_t)array_view->length > max_length) {
         ArrowErrorSet(
             error,
-            "Expected the 2 children of run-end encoded array to have equal length "
-            "but found mismatched lengths: run_ends->length=%ld, values->length=%ld",
-            (long)array_view->children[0]->length, (long)array_view->children[1]->length);
+            "Offset + length of a run-end encoded array must fit in a value"
+            " of the run end type %s, but offset + length is %lu while the "
+            "allowed maximum is %lu",
+            ArrowTypeString(run_ends_view->storage_type),
+            (unsigned long)array_view->offset + (unsigned long)array_view->length,
+            (unsigned long)max_length);
+        return EINVAL;
+      }
+      if (run_ends_view->length > values_view->length) {
+        ArrowErrorSet(
+            error, "Length of run_ends is greater than the length of values: %ld > %ld",
+            (long)run_ends_view->length, (long)values_view->length);
+        return EINVAL;
+      }
+      if (run_ends_view->length == 0 && values_view->length != 0) {
+        ArrowErrorSet(error,
+                      "Run-end encoded array has zero length %ld, but values array has "
+                      "non-zero length ",
+                      (long)values_view->length);
+        return EINVAL;
+      }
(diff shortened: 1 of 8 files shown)


paleolimbot (commented):
This is very clean and I look forward to giving it a closer look tomorrow!

@felipecrv if you have the bandwidth, would you mind taking a look through this for the details of the run-end encoding?

src/nanoarrow/array.c
```diff
@@ -846,6 +856,22 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
...
+        return EINVAL;
+      }
+      if (array_view->children[0]->length < array_view->children[1]->length) {
+        ArrowErrorSet(
+            error,
+            "Expected the 2 children of run-end encoded array to have equal length "
+            "but found mismatched lengths: run_ends->length=%ld, values->length=%ld",
+            (long)array_view->children[0]->length, (long)array_view->children[1]->length);
+        return EINVAL;
+      }
```
felipecrv, line 873: It's OK to have less run-ends than values. It's the opposite that's wrong because you always start the indexing of values by checking the run-ends first:

1. find run-end index according to logical index
2. use that index with the values array

https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/ree_util.cc#L199

src/nanoarrow/array.c
```diff
@@ -995,6 +1021,31 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
...
+        }
+        prev_run_end = run_end;
+        if (i == run_ends_view->length - 1 && run_end != array_view->length) {
+          ArrowErrorSet(error,
+                        "Run End value %ld at index %ld exceeds the logical "
+                        "length of the run-end encoded array %ld",
+                        (long)run_end, (long)i, (long)array_view->length);
+          return EINVAL;
+        }
+      }
```
felipecrv, line 1046: This is how I wrote this loop with a single branch inside the loop:

```cpp
      if (run_ends[0] < 1) {
        return Status::Invalid(
            "All run ends must be greater than 0 but the first run end is ", run_ends[0]);
      }
      int64_t last_run_end = run_ends[0];
      for (int64_t index = 1; index < run_ends_length; index++) {
        const int64_t run_end = run_ends[index];
        if (run_end <= last_run_end) {
          return Status::Invalid(
              "Every run end must be strictly greater than the previous run end, "
              "but run_ends[",
              index, "] is ", run_end, " and run_ends[", index - 1, "] is ",
              last_run_end);
        }
        last_run_end = run_end;
      }
```

https://github.com/apache/arrow/blob/e4baf6be2167eb6ccbda90275304336f49998eac/cpp/src/arrow/array/validate.cc#L753

src/nanoarrow/nanoarrow_types.h
```diff
@@ -953,6 +956,16 @@ static inline void ArrowDecimalSetBytes(struct ArrowDecimal* decimal,
...
+/// \brief A representation of a run of the run end encoded array
+/// \ingroup nanoarrow-utils
+///
+struct ArrowRunEndEncoded {
+  /// \brief The length of this run
+  int64_t length;
+
+  /// \brief The value of this run
+};
+
```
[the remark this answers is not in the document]
cocoa-xu: Oh sorry! I forgot to remove these lines -- originally thought that we might need something to represent a run but it turned out there's no need for that :)

src/nanoarrow/array.c
```diff
@@ -995,6 +1021,31 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
...
+                      (long)run_ends_view->null_count);
+        return EINVAL;
+      }
+      for (int64_t i = 0, prev_run_end = 0; i < run_ends_view->length; i++) {
+        int64_t run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, i);
+        if (run_end < 0 || run_end < prev_run_end) {
+          return EINVAL;
+        }
+        prev_run_end = run_end;
+        if (i == run_ends_view->length - 1 && run_end != array_view->length) {
```
felipecrv, line 1039: This "violation" is not really a violation. The parent array can change it's length without mutating the immutable run ends children.

Also keep in mind that if the parent has an offset, that last run-end would be affected and not match the logical length of the parent.

src/nanoarrow/array.c
```diff
@@ -1163,6 +1214,40 @@ static int ArrowArrayViewValidateFull(struct ArrowArrayView* array_view,
...
+      int64_t run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, i);
+      if (run_end < 0 || run_end < prev_run_end) {
+        return EINVAL;
+      }
+      prev_run_end = run_end;
+    }
+    if (prev_run_end != array_view->length) {
+      ArrowErrorSet(error,
+                    "The last run end value of a run-end encoded array must be equal to "
+                    "the logical length"
```
felipecrv, line 1243: Nope. :-) And this is what make run-end encoded arrays so mind-bending.

Think of the children run-ends and values arrays as forming a virtual big array starting from zero and ending at the last run-end. Then the parent offset/length pair are just a projection onto this big array.

The offset on the parent can land on any of the runs. And that offset plus the length can land on any of the runs and the run doesn't necessarily end when the parent's offset+length says.

For example, in this REE

```
offset: 1
length: 3
  run_ends: [2, 5, 6]
  values:   [A, B, C]
```

the uncompressed "big array" is `[A A B B B C]` but the entire REE is `[A B B]`.

felipecrv (changes requested):
Initial review with some quick thoughts.

cocoa-xu:
Hi @felipecrv, many thanks for the code review and these examples! I've adapted some code from the code snippets in the review, and it should handle the validation properly now. :)

src/nanoarrow/array_test.cc
```diff
@@ -1440,6 +1441,86 @@ TEST(ArrayTest, ArrayTestAppendToStructArray) {
...
+  EXPECT_STREQ(ArrowErrorMessage(&error),
+               "Expected 2 child of run_end_encoded array but found 0 child arrays");
+  array.n_children = 2;
+
+  // Make sure final child size is checked at finish
+  array.children[0]->length = array.children[0]->length - 1;
+  EXPECT_EQ(ArrowArrayFinishBuilding(&array, NANOARROW_VALIDATION_LEVEL_FULL, &error),
+            EINVAL);
+  EXPECT_STREQ(ArrowErrorMessage(&error),
+               "Last run end is 6 but it should match 7 (offset: 0, length: 7)");
```
paleolimbot, line 1477: Maybe `>= 7`? (i.e., the last run end must be greater than or equal to the length of the parent array?)

src/nanoarrow/array_test.cc
```diff
@@ -1440,6 +1441,86 @@ TEST(ArrayTest, ArrayTestAppendToStructArray) {
...
+  EXPECT_EQ(ArrowArrayFinishBuildingDefault(&array, nullptr), NANOARROW_OK);
+
+  // run-end encoded array with offsets
+  struct ArrowArrayView array_view;
+  ASSERT_EQ(ArrowArrayViewInitFromSchema(&array_view, &schema, NULL), NANOARROW_OK);
+  ASSERT_EQ(ArrowArrayViewSetArray(&array_view, &array, NULL), NANOARROW_OK);
+  array_view.length = 4;
+  array_view.offset = 1;
+  EXPECT_EQ(ArrowArrayViewValidate(&array_view, NANOARROW_VALIDATION_LEVEL_FULL, &error),
+            NANOARROW_OK);
```
paleolimbot, line 1489: It may be helpful to add a comment to each of these cases to explain what they are testing? For this one in particular (non-zero offset), I wonder if `length = 6` and `offset = 1` might be a better choice (to check that the last run end -- parent array math is correct).

src/nanoarrow/array_test.cc
```diff
@@ -1440,6 +1441,86 @@ TEST(ArrayTest, ArrayTestAppendToStructArray) {
...
+
+  array_view.length = 10;
+  array_view.offset = 1;
+  EXPECT_EQ(ArrowArrayViewValidate(&array_view, NANOARROW_VALIDATION_LEVEL_FULL, &error),
+            EINVAL);
+  EXPECT_STREQ(ArrowErrorMessage(&error),
+               "Last run end is 7 but it should match 11 (offset: 1, length: 10)");
```
paleolimbot, line 1496: Similarly, I wonder if `length = 7`, `offset = 1` would be a better check (only off by one)

src/nanoarrow/array_inline.h
```diff
@@ -661,6 +661,44 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+      if (array->children[0]->length != array->children[1]->length) {
+        return EINVAL;
+      }
+      if (array->children[0]->null_count != 0) {
+        return EINVAL;
+      }
+      array->length = 0;
+      struct ArrowBuffer* data_buffer;
+      data_buffer = ArrowArrayBuffer(array->children[0], 1);
+      for (int64_t i = 0; i < array->children[0]->length; i++) {
```
paleolimbot, line 675: I am not sure that you need a loop here? (i.e., is it possible to just extract the last run end?)

I wonder if this would be a better fit for `ArrowArrayFinishBuilding()`? Or maybe we need `FinishAppending()`? I am not sure I would have thought of `FinishElement` to do these checks/updates since they really only need to happen once per array.

src/nanoarrow/array_inline.h
```diff
@@ -661,6 +661,44 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+        }
+        if (run_end < 0) {
+          return EINVAL;
+        }
+        array->length = run_end;
+      }
+
+      if (private_data->bitmap.buffer.data != NULL) {
+        NANOARROW_RETURN_NOT_OK(ArrowBitmapAppend(ArrowArrayValidityBitmap(array), 1, 1));
+      }
```
paleolimbot, line 699: I am not sure that the outer array here is allowed to have a validity buffer?

src/nanoarrow/nanoarrow.h
```diff
@@ -372,6 +374,15 @@ ArrowErrorCode ArrowSchemaSetTypeDecimal(struct ArrowSchema* schema, enum ArrowT
...
                                          int32_t decimal_scale);
 
+/// \brief Set the format field of a run-end encoded schema
+///
+/// Returns EINVAL for scale <= 0 or for run_end_type that is not
+/// NANOARROW_TYPE_INT16, NANOARROW_TYPE_INT32 or NANOARROW_TYPE_INT64.
+/// Schema must have been initialized using ArrowSchemaInit() or ArrowSchemaDeepCopy().
+ArrowErrorCode ArrowSchemaSetTypeRunEndEncoded(struct ArrowSchema* schema,
+                                               enum ArrowType run_end_type,
+                                               enum ArrowType value_type);
```
paleolimbot, line 384: I might leave this as `struct ArrowSchema* schema, enum ArrowType run_end_type` and require that the caller do `ArrowSchemaSetType(schema.children[1], ...)`. The `run_end_type` can always be represented by a simple `ArrowType` value but the `value_type` could be a nested or parameterized type for which a simple enum value won't be able to construct it.

paleolimbot (commented):
Thank you again (and thanks to Felipe for doing a first pass on the details)! I have a few comments here for discussion 🙂

src/nanoarrow/array.c
```diff
@@ -846,6 +856,75 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
...
+      if (array_view->n_children != 2) {
+        ArrowErrorSet(error,
+                      "Expected 2 children for run-end encoded array but found %ld",
+                      (long)array_view->n_children);
+        return EINVAL;
+      }
+      struct ArrowArrayView* run_ends_view;
+      struct ArrowArrayView* values_view;
+      run_ends_view = array_view->children[0];
+      values_view = array_view->children[1];
```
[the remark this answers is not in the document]
cocoa-xu: Sadly we cannot do this because it will cause compile errors... https://github.com/apache/arrow-nanoarrow/actions/runs/9382594421/job/25834344055

```
[  3%] Building CXX object CMakeFiles/nanoarrow.dir/amalgamation/nanoarrow/nanoarrow.cc.o
/home/runner/work/arrow-nanoarrow/arrow-nanoarrow/build/amalgamation/nanoarrow/nanoarrow.cc: In function ‘int ArrowArrayViewValidateMinimal(ArrowArrayView*, ArrowError*)’:
/home/runner/work/arrow-nanoarrow/arrow-nanoarrow/build/amalgamation/nanoarrow/nanoarrow.cc:3022:5: error: jump to case label
 3022 |     default:
      |     ^~~~~~~
/home/runner/work/arrow-nanoarrow/arrow-nanoarrow/build/amalgamation/nanoarrow/nanoarrow.cc:2973:16: note:   crosses initialization of ‘uint64_t max_length’
 2973 |       uint64_t max_length = 0;
      |                ^~~~~~~~~~
/home/runner/work/arrow-nanoarrow/arrow-nanoarrow/build/amalgamation/nanoarrow/nanoarrow.cc:2964:30: note:   crosses initialization of ‘ArrowArrayView* values_view’
 2964 |       struct ArrowArrayView* values_view = array_view->children[1];
      |                              ^~~~~~~~~~~
/home/runner/work/arrow-nanoarrow/arrow-nanoarrow/build/amalgamation/nanoarrow/nanoarrow.cc:2963:30: note:   crosses initialization of ‘ArrowArrayView* run_ends_view’
 2963 |       struct ArrowArrayView* run_ends_view = array_view->children[0];
      |                              ^~~~~~~~~~~~~
gmake[2]: *** [CMakeFiles/nanoarrow.dir/build.make:76: CMakeFiles/nanoarrow.dir/amalgamation/nanoarrow/nanoarrow.cc.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:230: CMakeFiles/nanoarrow.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
```

src/nanoarrow/array.c
```diff
@@ -846,6 +856,75 @@ static int ArrowArrayViewValidateMinimal(struct ArrowArrayView* array_view,
...
+      if (run_ends_view == NULL) {
+        ArrowErrorSet(error, "Run ends array is null pointer");
+        return EINVAL;
+      }
+      if (values_view == NULL) {
+        ArrowErrorSet(error, "Values array is null pointer");
+        return EINVAL;
+      }
+      uint64_t max_length;
```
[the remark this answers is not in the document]
cocoa-xu: But if the malformed array_view has an `offset = INT64_MAX - 1` and `length = 2`, that would cause an overflow and we wouldn't be able to test that scenario (after this `switch` block).

src/nanoarrow/array.c
```diff
@@ -995,6 +1074,40 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
         }
       }
       break;
+
+    case NANOARROW_TYPE_RUN_END_ENCODED: {
+      struct ArrowArrayView* run_ends_view;
+      run_ends_view = array_view->children[0];
```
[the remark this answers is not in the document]
cocoa-xu: This would cause the same compilation errors, `error: jump to case label`

src/nanoarrow/array.c
```diff
@@ -995,6 +1074,40 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
...
+        if (run_end <= last_run_end) {
+          ArrowErrorSet(
+              error,
+              "Every run end must be strictly greater than the previous run end, "
+              "but run_ends[%ld] is %ld and run_ends[%ld] is %ld",
+              (long)i, (long)run_end, (long)i - 1, (long)last_run_end);
+          return EINVAL;
+        }
+        last_run_end = run_end;
+      }
```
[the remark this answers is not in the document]
cocoa-xu: No problem, I'll move it there

src/nanoarrow/nanoarrow.h
cocoa-xu: OK got it!

src/nanoarrow/array.c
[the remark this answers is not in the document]
paleolimbot: Perhaps something like:

```c
case NANOARROW_TYPE_RUN_END_ENCODED: {

}
```

(i.e., add the braces?)

src/nanoarrow/array.c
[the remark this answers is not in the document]
paleolimbot: Great point! (Perhaps add a comment so that nobody else tries to "fix" this?)

src/nanoarrow/array.c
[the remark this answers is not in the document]
cocoa-xu: Ah I see!

src/nanoarrow/array_inline.h
cocoa-xu: Maybe `ArrowArrayFinishRunEndEncoded` might be better as it's more suggestive from its name and less confusion?

paleolimbot: Perhaps, but it is also very similar to the piece of code we have that updates the `null_count` based on the validity buffer, which I think happens in `ArrowArrayFinishBuilding()` (although that "feature" has caused me some personal confusion because I sometimes forget that it happens and wonder why the `null_count` isn't what I set it to be).

The most backward compatible thing to do would be to not do *any* updating of lengths (i.e., force the caller to set the length of the parent run-end-encoded array). This would not necessarily be difficult to do because they would have to be keeping track of that length to append to the run-ends child (so they would have to keep a counter somewhere anyway). Perhaps for this PR the helper could be omitted and we can learn from the experience of implementing this elsewhere what the best intervention would be?

cocoa-xu: > Perhaps for this PR the helper could be omitted and we can learn from the experience of implementing this elsewhere what the best intervention would be?

No problem, I've removed it :)

cocoa-xu:
Just added more tests to cover more lines and the memory issue in the schema test should be fixed.

src/nanoarrow/array.c
```diff
@@ -1163,6 +1236,36 @@ static int ArrowArrayViewValidateFull(struct ArrowArrayView* array_view,
...
 
+  if (array_view->storage_type == NANOARROW_TYPE_RUN_END_ENCODED) {
+    struct ArrowArrayView* run_ends_view = array_view->children[0];
+    int64_t last_run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, 0);
+    if (last_run_end < 1) {
+      ArrowErrorSet(error,
+                    "All run ends must be greater than 0 but the first run end is %ld",
+                    (long)last_run_end);
+      return EINVAL;
+    }
```
paleolimbot, line 1247: This check might be good to keep in `ArrowArrayViewValidateDefault()` (because it's a check that only happens once per array regardless of the array length).

paleolimbot (approved):
Just a few nits, but this looks good from my end!

@felipecrv are there any more REE-specific details relevant to this PR (or follow-ups that we need to open issues for and tackle before the next release?)

src/nanoarrow/array.c
cocoa-xu: Got it!

felipecrv:
> Just a few nits, but this looks good from my end!
> 
> @felipecrv are there any more REE-specific details relevant to this PR (or follow-ups that we need to open issues for and tackle before the next release?)

Not that I'm aware right now. I recommend that you and @cocoa-xu take a look at the C++ code around REEs to see all the traps. I don't remember everything that I did, but I left many comments in the C++ implementation.

https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/ree_util.h
and
https://github.com/apache/arrow/blob/main/cpp/src/arrow/util/ree_util.cc
are good sources to look at.

felipecrv (approved):

WillAyd:
I think this PR introduced a regression that valgrind is complaining about. You can see it started back in the weekly Meson build:

https://github.com/apache/arrow-nanoarrow/actions/runs/9432588616

A more detailed error description can be found here:

https://github.com/apache/arrow-nanoarrow/actions/runs/9457996609/job/26052769002?pr=483#step:7:319

Apparently Valgrind does not like something that ultimately hits this codeblock from the RunEndEncoded tests:

https://github.com/apache/arrow-nanoarrow/blob/f443e46163ded43ddb5e024c1a2f1fddd3b43c3f/src/nanoarrow/array.c#L1247

src/nanoarrow/array.c
```diff
@@ -1163,6 +1247,30 @@ static int ArrowArrayViewValidateFull(struct ArrowArrayView* array_view,
...
+      if (run_end <= last_run_end) {
+        ArrowErrorSet(error,
+                      "Every run end must be strictly greater than the previous run end, "
+                      "but run_ends[%ld] is %ld and run_ends[%ld] is %ld",
+                      (long)i, (long)run_end, (long)i - 1, (long)last_run_end);
+        return EINVAL;
+      }
+      last_run_end = run_end;
+    }
+    last_run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, run_ends_view->length - 1);
```
WillAyd, line 1264:
```cpp
last_run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, run_ends_view->length - 1);
```

Is there any chance of run_ends_view->length being 0 here? If so I _think_ this might invoke UB

cocoa-xu: Good catch! I'll send a PR

PR #508, opened 2024-06-06
chore: Move bundling from CMake to Python

### LINKED ISSUE #485: #include "nanoarrow/nanoarrow.h" or "nanoarrow.h"?
Since we are getting closer to having nanoarrow set up in the meson wrap database, I wanted to check if it was intended to support both `#include "nanoarrow/nanoarrow.h"` and `#include "nanoarrow.h"`. Even the documentation seems to mix the two styles.

Its not a big deal, but we are explicitly adding both include directories to the meson dependency. If the long term goal was to only use one it might be easier to standardize that before we create the wrap database entry

paleolimbot:
Some of the work with the Meson build system has highlighted that the bundling is currently very inflexible because it's implemented in CMake. This was initially done because it made it easy to use `configure_file()` in CMake to keep the version information consistent, and because it allowed a straightforward path to testing the bundled files in the same way as the normal build.

This had the unfortunate side effect of requiring CMake to "build" the R or Python packages as is generally confusing, and used the CMake "install" in a confusing way (it installed sources). Instead, we can do this with a Python script (and call that from CMake, R, and Python). The testing of the namespaced builds is still done in CMake, which is essential to ensure that both build types are tested in the same way.

This also closes #485 by prefixing all nanoarrow includes with "nanoarrow/" in the source files. Many of them had to not use this prefix because when the files were bundled they weren't always bundled into a "nanoarrow/" directory. This feature has been moved to the bundling script which can now namespace or un-namespace the nanoarrow includes.


ci/scripts/bundle.py
```diff
@@ -0,0 +1,232 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+import io
+import os
+import pathlib
+import re
+
+
+def read_content(path_or_content):
```
WillAyd, line 24: Of course up to you but I find the behavior of this function surprising. More often than not, when you accept a path argument then accepting a string usually still represents the filepath/name, not the contents itself.

If you wanted you could refactor this to just be:

```python
def read_content(path_or_content: io.StringIO | pathlib.Path | str):
    with open(path_or_content) as f:
        return f.read()
```

For the caller this would require them to pass `io.StringIO("some_contents")` instead of just `"some_contents"`, but I think makes it less ambiguous in case someone expects a string argument here to still represent a path

assignUser (approved):
LGTM +1 on python > bash for the bundling ^^

ci/scripts/bundle.py
paleolimbot: I agree it's a little confusing...if this utility gets more widespread use than just this bundler it should definitely get revisited!

PR #509, opened 2024-06-07
feat(extensions/nanoarrow_device): Implement asynchronous buffer copying

### LINKED ISSUE #245: Device extension should implement async buffer copying for CUDA

paleolimbot:
This PR implements asychronous buffer copying when copying CUDA buffers. Before this, we had basically been issuing `cuMemCopyDtoH/HtoD()` a lot of times in a row with a synchronize up front and a synchronize at the end. This was probably not great for performance. Additionally, for copying String/Binary/Large String/Large Binary arrays from CUDA to the CPU, we were issuing very tiny copies on the offsets buffer and synchronizing with the CPU to get the number of bytes to copy for the data buffer.

After this PR, when copying from CPU to CUDA, we will be able to return before the copy is necessarily completed by setting the output `sync_event`.

When copying from CUDA to CPU, the copy is done in one pass if there are no string/binary arrays, or two passes if there are. When copying string/binary arrays, the implementation walks the entire tree of arrays and issues asynchronous copies for the last offset value. Then, the stream is synchronized with the CPU, and a second set of asynchronous copies are issued for the buffers whose size we now know.

I don't have much experience with CUDA async programming to know if this approach could be simplified (e.g., I do this in two streams, but it might be that one stream is sufficient since perhaps all of the device -> host copies are getting queued against eachother regardless of what stream they're on).

This will be easier to test when (e.g., bigger, non-trivial data) it is wired up to Python.

TODO:

- Implement `sync_event` integration (both for source and destination)
- Test more than just a few string arrays

Closes #245.


src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,247 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+    case NANOARROW_TYPE_BINARY:
+      if (array_view->buffer_views[2].size_bytes == -1) {
+        src_view.data.as_int32 =
+            array_view->buffer_views[1].data.as_int32 + offset_plus_length;
+        src_view.size_bytes = sizeof(int32_t);
+
+        dst_view.data.data = &(array_view->buffer_views[2].size_bytes);
+        dst_view.size_bytes = sizeof(int32_t);
+
+        // Note that this strategy (copying four bytes into an int64_t)
```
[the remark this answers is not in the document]
jorisvandenbossche: Ah, because essentially you are copying this into `array_view->buffer_views[2].size_bytes` and `size_bytes` is an int64?

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,247 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+      (*unknown_buffer_size_count)++;
+      continue;
+    }
+
+    struct ArrowBuffer* buffer_dst = ArrowArrayBuffer(dst, i);
+
+    // Don't initialize a buffer that has already been initialized
+    if (buffer_dst->size_bytes > 0) {
+      continue;
+    }
```
jorisvandenbossche, line 518: This ensures a buffer (that was already copied) doesn't get copied a second time when calling `ArrowDeviceCudaArrayViewCopyBuffers` a second time?

jorisvandenbossche (commented):
Can't review the CUDA async details, but just took a look if I could follow the general logic, and added a few questions

src/nanoarrow/nanoarrow_device.h
```diff
@@ -189,6 +194,11 @@ struct ArrowDevice {
                                struct ArrowDevice* device_dst,
                                struct ArrowDeviceArray* dst);
 
+  /// \brief Copy an ArrowDeviceArray to a device
+  ArrowErrorCode (*array_copy)(struct ArrowDeviceArrayView* src,
+                               struct ArrowDevice* device_dst,
+                               struct ArrowDeviceArray* dst);
```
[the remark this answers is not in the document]
paleolimbot: The `ArrowDeviceArrayViewCopy()` function does the dispatch here...it tries `src->device->array_copy()`, then `dst->array_copy()`, and then falls back to an implementation based on just buffer copying if both of those return `ENOTSUP`. Basically, either the source *or* the destination has to know how to do the copy (and the CPU device returns ENOTSUP for everything except cpu--cpu).

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,247 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+        src_view.data.as_int32 =
+            array_view->buffer_views[1].data.as_int32 + offset_plus_length;
+        src_view.size_bytes = sizeof(int32_t);
+
+        dst_view.data.data = &(array_view->buffer_views[2].size_bytes);
+        dst_view.size_bytes = sizeof(int32_t);
+
+        // Note that this strategy (copying four bytes into an int64_t)
+        // only works on little-endian (should be checked at a higher level).
+        array_view->buffer_views[2].size_bytes = 0;
```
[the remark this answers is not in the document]
paleolimbot: I added a comment, it's because we need the high bits of the int64 to be zero in order for the positive little endian int32 that's about to be copied to the same address to be interpreted correctly. It's a bit of a hack but I'm not sure I have a better solution 🤷

src/nanoarrow/nanoarrow_device_cuda.c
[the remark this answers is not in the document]
paleolimbot: I added a comment that hopefully clarifies this!

src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: Yes, it's so that we can just call this function again (instead of keeping track of a sorted list of flattened buffer indexes or something to know exactly which buffers to initialize a copy for on the second pass). I added a comment that hopefully clarifies this!

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -157,6 +160,13 @@ static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
   return NANOARROW_OK;
 }
 
+static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
+                                                    struct ArrowBuffer* buffer,
+                                                    int64_t size_bytes) {
+  return ArrowDeviceCudaAllocateBufferAsync(device, buffer, size_bytes,
+                                            NANOARROW_CUDA_DEFAULT_STREAM);
+}
```
kkraus14, line 168: I don't think this is safe to do. CUDA Streams can be created that are non-blocking with respect to the default stream where this could lead to races if someone assumes that `ArrowDeviceCudaAllocateBuffer` is synchronous and safe to use from any stream.

Additionally, the default stream can be controlled at compile time where it can be set to a per thread stream which would make this extra problematic.

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -246,8 +252,8 @@ static ArrowErrorCode ArrowDeviceCudaBufferCopyInternal(struct ArrowDevice* devi
     (*n_pop_context)++;
 
     NANOARROW_CUDA_RETURN_NOT_OK(
-        cuMemcpyDtoD((CUdeviceptr)dst.data.data, (CUdeviceptr)src.data.data,
-                     (size_t)src.size_bytes),
+        cuMemcpyDtoDAsync((CUdeviceptr)dst.data.data, (CUdeviceptr)src.data.data,
+                          (size_t)src.size_bytes, hstream),
         "cuMemcpytoD", error);
```
kkraus14, line 257: May want to update to make this say `Async` as well?
```suggestion
        "cuMemcpyDtoD", error);
```

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -312,17 +319,26 @@ static ArrowErrorCode ArrowDeviceCudaBufferCopy(struct ArrowDevice* device_src,
...
 
-static ArrowErrorCode ArrowDeviceCudaBufferInit(struct ArrowDevice* device_src,
+static ArrowErrorCode ArrowDeviceCudaBufferCopy(struct ArrowDevice* device_src,
                                                 struct ArrowBufferView src,
                                                 struct ArrowDevice* device_dst,
-                                                struct ArrowBuffer* dst) {
+                                                struct ArrowBufferView dst) {
+  return ArrowDeviceCudaBufferCopyAsync(device_src, src, device_dst, dst,
+                                        NANOARROW_CUDA_DEFAULT_STREAM);
+}
```
kkraus14, line 328: Same issue here where if this is assumed to be synchronous then this can cause races for other non-blocking streams

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -345,6 +362,14 @@ static ArrowErrorCode ArrowDeviceCudaBufferInit(struct ArrowDevice* device_src,
   return NANOARROW_OK;
 }
 
+static ArrowErrorCode ArrowDeviceCudaBufferInit(struct ArrowDevice* device_src,
+                                                struct ArrowBufferView src,
+                                                struct ArrowDevice* device_dst,
+                                                struct ArrowBuffer* dst) {
+  return ArrowDeviceCudaBufferInitAsync(device_src, src, device_dst, dst,
+                                        NANOARROW_CUDA_DEFAULT_STREAM);
+}
```
kkraus14, line 371: +1

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,251 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
   return ENOTSUP;
 }
 
+static ArrowErrorCode ArrowDeviceCudaResolveBufferSizesAsync(
+    struct ArrowDevice* device, struct ArrowArrayView* array_view, CUstream hstream) {
```
kkraus14, line 421: `array_view` is host memory here, correct? If so, if the allocated memory is CUDA pinned host memory, then the copies will run truly asynchronously here and return before the memory is actually set to the resolved sizes, but nothing will be blocking the CPU. Is that a user's responsibility in this situation?

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,251 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+  NANOARROW_CUDA_ASSERT_OK(cuCtxPopCurrent(&unused));
+  NANOARROW_CUDA_RETURN_NOT_OK(err, "cuStreamCreate", error);
+
+  return NANOARROW_OK;
+}
+
+static void ArrowDeviceCudaDestroyStream(CUstream hstream) {
+  if (hstream != NANOARROW_CUDA_DEFAULT_STREAM) {
+    NANOARROW_CUDA_ASSERT_OK(cuStreamDestroy(hstream));
+  }
```
kkraus14, line 572: In addition to the default stream there's also the legacy default stream and per thread default stream (see `CU_STREAM_LEGACY` and `CU_STREAM_PER_THREAD` [here](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES))

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,251 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+
+  // Create a stream to do the first (and maybe only) round of buffer copying.
+  // The ArrowDeviceArrayView should probably have a sync_event so that we can
+  // force the stream we just created to wait on the event.
+  CUstream hstream = NANOARROW_CUDA_DEFAULT_STREAM;
+  NANOARROW_RETURN_NOT_OK(
+      ArrowDeviceCudaCreateStream(device_src, device_dst, &hstream, error));
+
+  code = ArrowDeviceCudaArrayViewCopyBuffers(device_src, &src->array_view, device_dst,
+                                             dst, hstream, &unknown_buffer_size_count);
```
kkraus14, line 590: If `src` has a sync event then we need to wait on that event in `hstream` otherwise we have a possible race here

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,251 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+    return code;
+  }
+
+  CUstream hstream_data = NANOARROW_CUDA_DEFAULT_STREAM;
+
+  if (unknown_buffer_size_count > 0) {
+    // If buffers with unknown size were encountered, we need to copy all of their
+    // sizes to the CPU, synchronize, then copy all of the buffers once the CPU
+    // knows how many bytes from each buffer need to be copied.
+    code = ArrowDeviceCudaCreateStream(device_src, device_dst, &hstream_data, error);
```
kkraus14, line 603: +1 here, need to wait on the sync event possibly provided by `src` on `hstream_data`

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +417,251 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+    }
+  }
+
+  if (hstream_data != NANOARROW_CUDA_DEFAULT_STREAM) {
+    err = cuStreamSynchronize(hstream_data);
+    ArrowDeviceCudaDestroyStream(hstream_data);
+    if (err != CUDA_SUCCESS) {
+      NANOARROW_CUDA_RETURN_NOT_OK(err, "cuStreamSynchronize(hstream_data)", error);
+    }
+  }
```
kkraus14, line 660: If we're not populating the sync event we probably want to synchronize the stream regardless of whether it's the default stream or not

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -400,6 +670,58 @@ static void ArrowDeviceCudaRelease(struct ArrowDevice* device) {
   device->release = NULL;
 }
 
+static ArrowErrorCode ArrowDeviceCudaArrayViewCopy(struct ArrowDeviceArrayView* src,
+                                                   struct ArrowDevice* device_dst,
+                                                   struct ArrowDeviceArray* dst) {
+  // For CUDA_HOST, the default implementation is sufficient
+  // TODO: Make sure we've synchronized here when CUDA_HOST is the source.
+  // cuCtxSynchronize() is probably necessary there (since the implementation
+  // calls raw memcpys)
```
kkraus14, line 679: Should only need to synchronize the relevant stream still. `cuCtxSynchronize()` synchronizes the entire device I believe.

src/nanoarrow/nanoarrow_device_cuda.c
kkraus14: In hindsight, I think this is an issue across most of the functions in this PR where we always need to be checking the sync event in source arrays and waiting on them in the stream used for the function

src/nanoarrow/nanoarrow_device.c
```diff
@@ -173,13 +183,13 @@ ArrowErrorCode ArrowDeviceArrayInit(struct ArrowDevice* device,
   return NANOARROW_OK;
 }
 
-ArrowErrorCode ArrowDeviceBufferInit(struct ArrowDevice* device_src,
-                                     struct ArrowBufferView src,
-                                     struct ArrowDevice* device_dst,
-                                     struct ArrowBuffer* dst) {
-  int result = device_dst->buffer_init(device_src, src, device_dst, dst);
+ArrowErrorCode ArrowDeviceBufferInitAsync(struct ArrowDevice* device_src,
```
zeroshade, line 186: Are we sure we want to remove the non-async versions?

src/nanoarrow/nanoarrow_device.c
```diff
@@ -435,9 +451,9 @@ static ArrowErrorCode ArrowDeviceArrayViewCopyInternal(struct ArrowDevice* devic
   return NANOARROW_OK;
 }
 
-ArrowErrorCode ArrowDeviceArrayViewCopy(struct ArrowDeviceArrayView* src,
-                                        struct ArrowDevice* device_dst,
-                                        struct ArrowDeviceArray* dst) {
+static ArrowErrorCode ArrowDeviceArrayViewCopyDefault(struct ArrowDeviceArrayView* src,
```
zeroshade, line 454: why this one specifically getting `static` added?

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +404,202 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+    struct ArrowDevice* device_dst, struct ArrowArray* dst, CUstream* stream,
+    struct ArrowError* error) {
+  CUresult err;
+  int code;
+  int64_t unknown_buffer_size_count = 0;
+
+  NANOARROW_RETURN_NOT_OK_WITH_ERROR(
+      ArrowDeviceCudaArrayViewCopyBuffers(device_src, &src->array_view, device_dst, dst,
+                                          stream, &unknown_buffer_size_count),
+      error);
```
zeroshade, line 567: You probably need to call `cuStreamWaitEvent` if there's a source event before you pass it to these functions

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -392,6 +404,202 @@ static ArrowErrorCode ArrowDeviceCudaArrayMove(struct ArrowDevice* device_src,
...
+        error);
+
+    // There should not have been any buffers with unknown size in this round of copying
+    NANOARROW_DCHECK(unknown_buffer_size_count == 0);
+  }
+
+  // TODO: Populate the sync event instead of synchornizing
+
+  NANOARROW_CUDA_RETURN_NOT_OK(cuStreamSynchronize(*stream), "cuStreamSynchronize",
+                               error);
```
zeroshade, line 598: As long as you call `cuStreamWaitEvent` you don't have to manually synchronize the stream, you can leave that to the caller/consumer. We don't need the CPU to synchronize and wait, we just need the stream to wait for the event before it gets there.

src/nanoarrow/nanoarrow_device.c
paleolimbot: They're still there! (they're just `static inline` now and pass `NULL` as the sync event). It's not a perfect system since streams don't really apply to any other implementation but basically users of the CUDA version will be required to use the async versions (maybe we can do something smarter in the future).

src/nanoarrow/nanoarrow_device.c
paleolimbot: This is currently only referred to from within the file (as a fallback if neither src or dst provides an implementation). I'm still workshopping this (with the `stream` now available to pass through everywhere, we might be able to make this implementation-agnostic again).

src/nanoarrow/nanoarrow_device_cuda.c
kkraus14: If we're not populating a sync event then if call `cuStreamWaitEvent` with the given stream and someone tries to touch `dst` from a different stream we end up with a race. We either need to synchronize the stream or populate the sync event.

src/nanoarrow/nanoarrow_device_cuda.c
```diff
@@ -122,7 +123,7 @@ static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
   switch (device->device_type) {
     case ARROW_DEVICE_CUDA: {
       CUdeviceptr dptr = 0;
-      err = cuMemAlloc(&dptr, (size_t)size_bytes);
+      err = cuMemAllocAsync(&dptr, (size_t)size_bytes, *stream);
```
shwina, line 126: It's probably worth checking if the finding in https://github.com/apache/arrow-nanoarrow/pull/534 is relevant here. `cuMemAlloc()` is documented not to like `0` input for the `byte_size` parameter, but `cuMemAllocAsync()` has no such limitation AFAI can see.

paleolimbot: Thanks! I merged that and rebased for the extra quality-of-life improvements in the tests 🙂

src/nanoarrow/nanoarrow_device_cuda.c
paleolimbot: I *think* the sync event is now handled (i.e., we always call `cuStreamWaitEvent()` when consuming, and we always `cuEventRecord()` when copying to the GPU (and `cuStreamSynchronize()` when copying back to the CPU).

src/nanoarrow/nanoarrow_device.c
```diff
@@ -175,10 +175,12 @@ ArrowErrorCode ArrowDeviceArrayInitAsync(struct ArrowDevice* device,
...
-  // Handling a sync event or stream is not supported in the default constructor
-  if (sync_event != NULL || stream != NULL) {
-    return EINVAL;
-  }
+  // TODO: Right now we have to let these events through and ignore them
+  // because we don't have a good way to separate the source and destination
+  // streams.
+  // if (sync_event != NULL || stream != NULL) {
+  //   return EINVAL;
+  // }
```
zeroshade: The purpose of using a sync event is so that we don't need to care about the source and destination streams. We just need to tell the destination stream to wait for the event (if it exists) before passing it to the memcpy calls. If there's no sync_event, then we don't need to be concerned about the source stream at all

paleolimbot: I think I fixed this one! This "init" is done by the producer, who in this case has a stream that needs to be propagated to the consumer with as an event. That implementation (create an event and have the stream record it) lives in the "array init".

This implementation is for the CPU device, so if a stream or event get passed here something has gone wrong (that commented code has been updated). In this case, the issue was that the array copier takes a `stream` argument, but it wasn't clear if this was for the source or destination so it was getting passed to both.

This should now have been updated such that the stream is synchronized before the call to this function (and a NULL stream and sync event is passed here).

zeroshade: That sounds like a good solution and handling to me, though as far as taking the stream argument is concerned. The stream should always be considered to be "the stream that any operations should be put on", i.e. the destination. By using the whole sync-event handling that we do, we bypass the need to logically separate a source stream and destination stream: either we need to sync (there should be a sync event and the stream should be given a `cuStreamWaitEvent` on it) or there is no sync-event and thus no need to sync. The stream is always the stream to be used for any memcpy operations. Just mentioning for future reference to hopefully clear this up and make it easier to understand. 

obviously @kkraus14 can correct me if i'm wrong here... lol

zeroshade:
from @kkraus14's comment: 

> There could be numerous options as far as buffer copying strategies that could be employed here. I.E. copy different buffers on different streams, using a copy kernel, some mix, etc. I don't think it's feasible to capture / implement all of these different options here.
> * Should we allow folks to provide their own implementations for copies? Again, I'm not clear if this is already accomplished here.

I think we should definitely allow folks to provide their own implementations for copies. @paleolimbot how difficult is it for a user to inject their own implementations of the ArrowDevice object or callbacks to use for the copying (instead of using the defaults we are providing in this PR)?

src/nanoarrow/nanoarrow_device.c
paleolimbot: I think we're all on the same page! The `stream` argument in `ArrowDeviceArrayInit()` is "the stream that any operations *were* put on", which is handy for the likely case where a consumer just built a bunch of buffers using a stream, and is ready to hand a `ArrowDeviceArray` off to a consumer.

There is plenty of flexibility to refactor/rename/document exactly how/where this happens (e.g., perhaps the `ArrowDevice` abstraction is just not needed since the idioms for Metal mostly don't apply here). For now, I needed to implement the `stream` -> `event` logic somewhere and was a not unreasonable place to put it.

src/nanoarrow/device/cuda.c
```diff
@@ -100,9 +100,10 @@ static void ArrowDeviceCudaDeallocator(struct ArrowBufferAllocator* allocator,
   ArrowFree(allocator_private);
 }
 
-static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
-                                                    struct ArrowBuffer* buffer,
-                                                    int64_t size_bytes) {
+static ArrowErrorCode ArrowDeviceCudaAllocateBufferAsync(struct ArrowDevice* device,
+                                                         struct ArrowBuffer* buffer,
+                                                         int64_t size_bytes,
+                                                         CUstream* stream) {
```
zeroshade, line 106: `CUstream` is actually a typedef'd pointer, so you can probably pass it by value instead of via pointer

src/nanoarrow/device/cuda.c
```diff
@@ -181,10 +185,9 @@ static void ArrowDeviceCudaArrayRelease(struct ArrowArray* array) {
   array->release = NULL;
 }
 
-static ArrowErrorCode ArrowDeviceCudaArrayInit(struct ArrowDevice* device,
-                                               struct ArrowDeviceArray* device_array,
-                                               struct ArrowArray* array,
-                                               void* sync_event) {
+static ArrowErrorCode ArrowDeviceCudaArrayInitInternal(
+    struct ArrowDevice* device, struct ArrowDeviceArray* device_array,
+    struct ArrowArray* array, CUevent* cu_event) {
```
zeroshade, line 190: Same thing as with the stream, you can probably just pass this by value instead of pointer since you never need the actual address of this and `CUevent` is a typedef'd pointer anyway

src/nanoarrow/device/cuda.c
```diff
@@ -217,15 +220,63 @@ static ArrowErrorCode ArrowDeviceCudaArrayInit(struct ArrowDevice* device,
...
-                                                        int* n_pop_context,
-                                                        struct ArrowError* error) {
+  NANOARROW_CUDA_RETURN_NOT_OK(cuCtxPushCurrent(private_data->cu_context),
+                               "cuCtxPushCurrent", NULL);
+  CUcontext unused;  // needed for cuCtxPopCurrent()
+
+  CUevent* cu_event = (CUevent*)sync_event;
+
+  // If the stream was passed, we must have an event to work with
+  CUevent cu_event_tmp = NULL;
```
zeroshade, line 237: I don't think this is necessarily true. If there's no event then that just means that the producer said we don't need to worry about synchronizing.

src/nanoarrow/device/cuda_test.cc
```diff
@@ -55,6 +55,35 @@ class CudaTemporaryContext {
...
+    }
+
+    if (cuStreamCreate(&hstream_, CU_STREAM_DEFAULT) != CUDA_SUCCESS) {
+      return EINVAL;
+    }
+
+    return NANOARROW_OK;
+  }
+
+  CUstream* get() { return &hstream_; }
```
zeroshade, line 75: We should probably just return `CUstream` not a pointer to it

src/nanoarrow/device/metal.cc
```diff
@@ -292,7 +297,7 @@ static int ArrowDeviceMetalCopyRequiredCpuToMetal(MTL::Device* mtl_device,
 }
 
 static ArrowErrorCode ArrowDeviceMetalSynchronize(struct ArrowDevice* device,
-                                                  void* sync_event,
+                                                  void* sync_event, void* stream,
```
zeroshade, line 300: should we return a different status than `NANOARROW_OK` since it's not implemented?

paleolimbot: Definitely!

src/nanoarrow/device/cuda_test.cc
paleolimbot: I just tried this, but because the user-facing device-agnostic functions have to take a `void*`, it's slightly more convenient to have this return the `CUstream*`. (For consistency with how the `CUevent` is exposed int the device array (as a `CUevent*`, aka `struct something**`), I'm assuming that `void* stream` for CUDA is a `CUstream*`).

src/nanoarrow/device/cuda.c
paleolimbot: This is a self-imposed truth (I added the ability for `ArrowDeviceArray()` to accept a `void* stream`, and made this the behaviour). A user is also free to not provide an event OR a stream, in which case we don't have to worry about synchronizing.

I added a comment here and will add a test, too.

src/nanoarrow/device/cuda.c
paleolimbot: I updated all the internal functions to take a `CUevent/CUstream` instead of a `CUevent*/CUstream*`!

src/nanoarrow/device/cuda_test.cc
zeroshade: Your assumption is correct, that makes sense.

Issue #407, opened 2024-03-20
Make extensions available as CMake components


vyasr:
Currently the extensions to the API are built completely separately. It would be helpful to support building them as components of the library. This would also facilitate transitioning them out of an experimental state.

paleolimbot:
I agree! The pattern should be more like `-DNANOARROW_WITH_IPC=ON`.

WillAyd:
I think this is closable now after https://github.com/apache/arrow-nanoarrow/pull/517

paleolimbot:
I think the last piece is to ensure that the targets are installed (I don't think using the IPC or device components will work from an "installed" version of nanoarrow)!

WillAyd:
I think these are already being installed?

https://github.com/apache/arrow-nanoarrow/blob/2d5769bbc30c5ea3814ffcd9e098993ea83a8ec0/CMakeLists.txt#L229

https://github.com/apache/arrow-nanoarrow/blob/2d5769bbc30c5ea3814ffcd9e098993ea83a8ec0/CMakeLists.txt#L298

paleolimbot:
Possibly! Adding a test to the cmake build scenarios would ensure that we know for sure. If I'm remembering from the last time I went through the CMake, there are some other things that happen for each target that never made it to the device/IPC extensions.

PR #511, opened 2024-06-08
Consolidated IPC extension into main project

WillAyd:


WillAyd:
I don't believe the two test failures are related to this change - seems like they showed up in the weekly verification run

https://github.com/apache/arrow-nanoarrow/actions/runs/9432588619

.gitattributes
```diff
@@ -16,5 +16,4 @@
 # under the License.
 
 dist/* linguist-generated
-extensions/nanoarrow_ipc/thirdparty/* linguist-vendored
```
paleolimbot, line 19: Should this be changed to `thirdparty/*`?

CMakeLists.txt
```diff
@@ -310,10 +433,82 @@ if(NANOARROW_BUILD_TESTS)
...
+    target_link_libraries(nanoarrow_ipc_decoder_test
+                          nanoarrow_ipc
+                          nanoarrow
+                          flatccrt
+                          ${NANOARROW_ARROW_TARGET}
+                          gtest_main
+                          ipc_coverage_config)
+    target_include_directories(nanoarrow_ipc_decoder_test
+                               PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+                                      $<INSTALL_INTERFACE:include>)
```
paleolimbot, line 470: Other tests defined in this CMakeLists.txt don't use these generator expressions...are they needed?

ci/scripts/coverage.sh
```diff
@@ -81,11 +81,12 @@ function main() {
...
-    mkdir "${SANDBOX_DIR}/nanoarrow_ipc"
-    pushd "${SANDBOX_DIR}/nanoarrow_ipc"
+    rm -rf "${SANDBOX_DIR}/nanoarrow"
+    mkdir "${SANDBOX_DIR}/nanoarrow"
+    pushd "${SANDBOX_DIR}/nanoarrow"
 
-    cmake "${TARGET_NANOARROW_DIR}/extensions/nanoarrow_ipc" \
-        -DNANOARROW_IPC_BUILD_TESTS=ON -DNANOARROW_IPC_CODE_COVERAGE=ON
+    cmake "${TARGET_NANOARROW_DIR}" \
+        -DNANOARROW_IPC=ON -DNANOARROW_BUILD_TESTS=ON -DNANOARROW_CODE_COVERAGE=ON
```
paleolimbot, line 89: Does this need to be separate (i.e, can we just add `NANOARROW_IPC=ON` to the main nanoarrow coverage call?)

dev/release/verify-release-candidate.sh
```diff
@@ -246,17 +246,15 @@ test_c() {
   test_cmake_project build . -DNANOARROW_BUILD_TESTS=ON
 
   show_header "Build and test C IPC extension"
-  test_cmake_project build-ipc extensions/nanoarrow_ipc -DNANOARROW_IPC_BUILD_TESTS=ON
+  test_cmake_project build . -DNANOARROW_BUILD_TESTS=ON -DNANOARROW_IPC=ON
```
paleolimbot, line 249: Does this need to be separate (i.e, can we just add `NANOARROW_IPC=ON` to the line above it?)

paleolimbot (commented):
A few small comments! I'll also re-echo the heroicness of taking this on...it is definitely how those taking on nanoarrow as a dependency via meson or CMake expect this to work.

Linking https://github.com/apache/arrow-nanoarrow/issues/407 since this is approximately half of that issue. Before the next release we'll need to check the build/install interfaces and add examples (in a later PR).

I think that the benchmarking will need some changes to pass for both dev and previous versions. For now, I think that you can limit this change to the `if(IS_DIRECTORY "${NANOARROW_BENCHMARK_SOURCE_URL}")` case (which is what tests the current HEAD)...after this is in a released version we'll perhaps need to revise the strategy for building previous versions of the `nanoarrow_ipc` target.

https://github.com/apache/arrow-nanoarrow/blob/9f96d2459bea480d2aaff23210541d2ae0548839/dev/benchmarks/CMakeLists.txt#L55-L75

Moving the bundling to Python will make the root CMakeLists.txt a bit less verbose (as well as removing the need to invoke cmake from the R or Python packaging steps.

https://github.com/apache/arrow-nanoarrow/pull/514 should take care of the two verification failures!

CMakeLists.txt
WillAyd: Thanks for catching this - mistake on my end. I got confused while trying to get tests to find the proper include paths when bundled, but the actual fix needed was elsewhere.

ci/scripts/coverage.sh
WillAyd: Not at all - just kept as separate steps owing to the history, but happy to consolidate this into the existing "core" check

WillAyd:
Still think the windows CI failure might be unrelated - I don't see that  as having run in #514

paleolimbot (approved):
Thanks!

PR #512, opened 2024-06-09
fix(Python) Use timezone-aware timestamps in some unit tests on Windows

aosingh:
Using `datetime.timestamp()` for timestamps close to EPOCH fails on Windows crashing Python. 

Please check this run https://github.com/apache/arrow-nanoarrow/actions/runs/9428246621/job/25973355558

Following snippet

```python
 def test_c_array_timestamp_microseconds():
     d1 = int(round(datetime(1970, 1, 1).timestamp() * 1e6))
```
raises error

```bash
    OSError: [Errno 22] Invalid argument
```

The crux of the problem is explained in this [CPython issue](https://github.com/python/cpython/issues/81708). Solution is to use timezone aware datetime objects


paleolimbot (approved):
Thanks!

(We should add Windows to our Python CI to catch issues like this 😬 )

PR #517, opened 2024-06-10
refactor: Consolidate device extension into main project

WillAyd:


CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 # under the License.
 
 message(STATUS "Building using CMake version: ${CMAKE_VERSION}")
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.22)
```
WillAyd, line 19: The version bump is required to support the DEPENDS condition in the dependent option:

https://cmake.org/cmake/help/v3.29/module/CMakeDependentOption.html

Might be a creative way to work around this if 3.14 is a hard requirement

CMakeLists.txt
```diff
@@ -493,6 +606,55 @@ if(NANOARROW_BUILD_TESTS)
     gtest_discover_tests(nanoarrow_ipc_files_test)
     gtest_discover_tests(nanoarrow_ipc_hpp_test)
   endif()
+
+  if(NANOARROW_DEVICE)
+    enable_testing()
```
WillAyd, line 611: There are a few things here that are duplicative of what happens in the IPC branch directly preceding it, but didn't try to solve that as part of this PR. Figured a copy/paste from the old structure into the new would be easiest to review

docs/source/conf.py
```diff
@@ -66,8 +66,6 @@ def get_version():
 # Breathe configuration
 breathe_projects = {
```
WillAyd, line 67: Not super sure on this change - what should I be looking for with the documentation as part of this?

paleolimbot (commented):
Just a few comments! I agree that there's some repetition here that could be cleaned up now that all the projects are in the same CMakeLists.txt; however, this is a vast improvement and I think getting the bundling out of CMake will help simplify things considerably.

I'm working on the device extension over the next few weeks and will circle back to this CMake setup/file structure (e.g., I think we might want `src/nanoarrow/device|ipc/xxx.c` and perhaps to move some of the CMake into the subdirectories).

CMakeLists.txt
```diff
@@ -73,7 +92,8 @@ endif()
 
 configure_file(src/nanoarrow/nanoarrow_config.h.in generated/nanoarrow_config.h)
 
-if(NANOARROW_IPC AND (NANOARROW_BUILD_TESTS OR NOT NANOARROW_BUNDLE))
+if((NANOARROW_DEVICE OR NANOARROW_IPC) AND (NANOARROW_BUILD_TESTS OR NOT NANOARROW_BUNDLE
+                                           ))
```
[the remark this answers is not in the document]
WillAyd: I actually think it was a mistake to include NANOARROW_DEVICE here - the nanoarrow_device target on main includes the flatcc headers as part of the build interface but I don't think actually requires it. Just removed that instead to keep this section untouched

paleolimbot (approved):
Thanks!

PR #518, opened 2024-06-11
fix: check `run_ends_view->length` before accessing its values

cocoa-xu:
This PR should fix the issue where `run_ends_view->length` is not checked if equals to 0 before attempting to access `run_ends_view`'s values. Many thanks to @WillAyd.

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -1067,6 +1067,7 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
 
     case NANOARROW_TYPE_RUN_END_ENCODED: {
       struct ArrowArrayView* run_ends_view = array_view->children[0];
+      if (run_ends_view->length == 0) break;
       int64_t last_run_end = ArrowArrayViewGetIntUnsafe(run_ends_view, 0);
       if (last_run_end < 1) {
         ArrowErrorSet(error,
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
Thank you both!

WillAyd:
Wow that was fast - great work @cocoa-xu !

Unfortunately Valgrind still doesn't like something with how this is set up:

https://github.com/apache/arrow-nanoarrow/actions/runs/9468487196/job/26084953378?pr=483#step:7:308

I haven't been able to reproduce locally or boil it down to an MRE. Just sharing in case you have any other ideas what might have gone awry.

Thanks again for the contributions

WillAyd:
I think I was able to find some more information when running meson with `-Db_sanitize=address,undefined`:

```sh
../src/nanoarrow/array.c:753:11: runtime error: signed integer overflow: 9223372036854775807 + 7 cannot be represented in type 'long int'
../src/nanoarrow/array.c:892:9: runtime error: signed integer overflow: 9223372036854775807 + 7 cannot be represented in type 'long int'
```

It looks like the offsets value for the RUN_END_ENCODED parent array_view is a junk value, but `ArrowArrayViewValidateMinimal` does not expect this 

https://github.com/apache/arrow-nanoarrow/blob/e92c364f65f9d6fb029d918b9ff31a8a39b3a1df/src/nanoarrow/array.c#L752

So that is definitely problematic. Haven't quite pieced together the link between that and what valgrind is giving us, but maybe there is one

PR #520, opened 2024-06-11
refactor: Use inttypes.h macros instead of casts to print fixed-width integers

WillAyd:
Very pedantic change to the great work done in https://github.com/apache/arrow-nanoarrow/pull/507 just to prevent any issues on LLP64 systems

### DIFF
--- a/src/nanoarrow/array.c
+++ b/src/nanoarrow/array.c
@@ -16,6 +16,7 @@
 // under the License.
 
 #include <errno.h>
+#include <inttypes.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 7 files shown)


paleolimbot (commented):
This is a great idea!

I think that if we do this, we should do it for all the places where we are trying to printf an int64_t, which is quite a lot of places.

I also recall that I ran into some problems on mingw when packaging ADBC for R, which is one of the reasons that the "blind cast to `long`" + `%ld` was used when I started out.

I wonder if a utility function or macro + `NANOARROW_PRID64` define in `nanoarrow_types.h` could be used to isolate this change in case we (or somebody else) needs to work around this? (Dragging inttypes.h into nanoarrow.h is maybe not great, though, and there are definitely some places in buffer_inline/array_inline that need to print int64s).

WillAyd:
Sure happy to take a larger look. Do you know what the R issues were with ADBC? Would have to double check but I am fairly certain this is a C99 macro, so surprised that something wouldn't support it at this point

WillAyd:
Are you thinking of just redefining PRId64 via macros or wrapping a new printf function? I'm a little wary of trying to make something not standards compliant become standards compliant. Doesn't feel like a great value add to nanoarrow, but of course open to whatever direction you want to take this in

src/nanoarrow/nanoarrow_device.c
```diff
@@ -365,8 +366,9 @@ ArrowErrorCode ArrowDeviceArrayViewSetArrayMinimal(
   struct ArrowDevice* device =
       ArrowDeviceResolve(device_array->device_type, device_array->device_id);
   if (device == NULL) {
-    ArrowErrorSet(error, "Can't resolve device with type %d and identifier %ld",
-                  (int)device_array->device_type, (long)device_array->device_id);
+    ArrowErrorSet(error,
+                  "Can't resolve device with type %" PRId32 " and identifier %l" PRId64,
```
paleolimbot, line 370:
```suggestion
                  "Can't resolve device with type %" PRId32 " and identifier %" PRId64,
```

src/nanoarrow/array.c
```diff
@@ -1145,7 +1152,7 @@ static int ArrowAssertIncreasingInt64(struct ArrowBufferView view,
 
   for (int64_t i = 1; i < view.size_bytes / (int64_t)sizeof(int64_t); i++) {
     if (view.data.as_int64[i] < view.data.as_int64[i - 1]) {
-      ArrowErrorSet(error, "[%ld] Expected element size >= 0", (long)i);
+      ArrowErrorSet(error, "[%" PRId64 "] Expected element size >= 0", (long)i);
```
paleolimbot, line 1155:
```suggestion
      ArrowErrorSet(error, "[%" PRId64 "] Expected element size >= 0", i);
```

paleolimbot (approved):
Epic!

Just two typos that my local build complained about, possibly because it uses `-DNANOARROW_DEBUG`.

PR #522, opened 2024-06-12
refactor: Remove CMake requirement from Meson IPC config

WillAyd:
Follow up to https://github.com/apache/arrow-nanoarrow/pull/483


paleolimbot (approved):
Thank you!

PR #523, opened 2024-06-12
chore(c): Add UBSAN / ASAN to Meson CI job

WillAyd:
Figured this is a pretty easy addition, and might help catch more mistakes up front if this becomes a standard CI job

### DIFF
--- a/ci/scripts/build-with-meson.sh
+++ b/ci/scripts/build-with-meson.sh
@@ -66,12 +66,21 @@ function main() {
     pushd "${SANDBOX_DIR}"
 
     show_header "Run test suite"
-    meson configure -Dtests=true -Db_coverage=true -Dipc=true
+    meson configure \
+          -Dbuildtype=debugoptimized
+          -Db_sanitize="address,undefined"
+          -Dtests=true \
+          -Db_coverage=true \
+          -Dipc=true
     meson compile
     meson test --wrap='valgrind --track-origins=yes --leak-check=full' --print-errorlogs
 
     show_header "Run benchmarks"
-    meson configure -Dbenchmarks=true
+    meson configure \
+          -Dbuildtype=release \
+          -Db_sanitize=none \
+          -Db_coverage=false \
+          -Dbenchmarks=true
     meson compile
     meson test --benchmark --print-errorlogs
 
--- a/ci/scripts/build-with-meson.sh
+++ b/ci/scripts/build-with-meson.sh
@@ -65,16 +65,34 @@ function main() {
 
     pushd "${SANDBOX_DIR}"
 
-    show_header "Run test suite"
+    show_header "Run ASAN/UBSAN test suite"
     meson configure \
           -Dbuildtype=debugoptimized \
           -Db_sanitize="address,undefined" \
           -Dtests=true \
-          -Db_coverage=true \
           -Dipc=true
     meson compile
     export ASAN_OPTIONS=allocator_may_return_null=1  # allow ENOMEM tests
+    meson test --print-errorlogs
+
+    show_header "Run valgrind test suite"
+    meson configure \
+          -Dbuildtype=debugoptimized \
+          -Db_sanitize=none \
+          -Dtests=true \
+          -Dipc=true
+    meson compile
     meson test --wrap='valgrind --track-origins=yes --leak-check=full' --print-errorlogs
+
+    show_header "Run coverage test suite"
+    meson configure \
+          -Dbuildtype=release \
+          -Db_sanitize=none \
+          -Db_coverage=true \
+          -Dtests=true \
+          -Dipc=true
+    meson compile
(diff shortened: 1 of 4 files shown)


src/nanoarrow/array_test.cc
```diff
@@ -87,12 +87,6 @@ TEST(ArrayTest, ArrayTestAllocateChildren) {
   EXPECT_EQ(array.n_children, 0);
   ArrowArrayRelease(&array);
 
-  ASSERT_EQ(ArrowArrayInitFromType(&array, NANOARROW_TYPE_STRUCT), NANOARROW_OK);
```
WillAyd, line 90: Unfortunately ASAN does not like this code block:

```
==897112==ERROR: AddressSanitizer: requested allocation size 0x7ffffffffffffff8 (0x8000000000000ff8 after adjustments for alignment, red zones etc.) exceeds maximum supported size of 0x10000000000 (thread T0)
```

It might be possible to add a function attribute to disable instrumentation:

https://github.com/google/sanitizers/wiki/AddressSanitizer#turning-off-instrumentation

But couldn't find a reasonable way to make that work with gtest

WillAyd: actually solved via `ASAN_OPTIONS=allocator_may_return_null=1`

paleolimbot: Thanks (it's tricky to get some of those errors to fire and ensuring memory doesn't leak on error is nice to test if possible!)

WillAyd: Meson has allows you to pick either `address,undefined` or `leak` (amongst other options) but not both at the same time. We may want to consider a second compilation for leak checking in that case

paleolimbot (approved):
Thanks!

I think https://github.com/apache/arrow-nanoarrow/pull/524 will fix the overflow issue 🙂

src/nanoarrow/nanoarrow_ipc_decoder_test.cc
```diff
@@ -252,7 +252,7 @@ TEST(NanoarrowIpcTest, NanoarrowIpcVerifyInvalid) {
 
     memcpy(simple_schema_invalid, kSimpleSchema, i);
     memcpy(simple_schema_invalid + i, kSimpleSchema + (i + 1),
-           (sizeof(simple_schema_invalid) - i));
+           (sizeof(simple_schema_invalid) - i - 1));
```
WillAyd, line 255: Fixes an off by one in the tests:

```sh
==980248==ERROR: AddressSanitizer: global-buffer-overflow on address 0x5622387a2880 at pc 0x7f1435c3a397 bp 0x7ffe37c04940 sp 0x7ffe37c040e8
READ of size 159 at 0x5622387a2880 thread T0
    #0 0x7f1435c3a396 in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:827
    #1 0x5622385468f4 in NanoarrowIpcTest_StreamReaderInvalidBuffer_Test::TestBody() ../src/nanoarrow/nanoarrow_ipc_reader_test.cc:384
    #2 0x56223869b6a1 in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ../subprojects/googletest-1.14.0/googletest/src/gtest.cc:2612
    #3 0x5622386710d6 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ../subprojects/googletest-1.14.0/googletest/src/gtest.cc:2648

...
0x5622387a2880 is located 0 bytes to the right of global variable 'kSimpleRecordBatch' defined in '../src/nanoarrow/nanoarrow_ipc_reader_test.cc:46:16' (0x5622387a27e0) of size 160
0x5622387a2880 is located 32 bytes to the left of global variable 'kEndOfStream' defined in '../src/nanoarrow/nanoarrow_ipc_reader_test.cc:60:16' (0x5622387a28a0) of size 8
```

ci/scripts/build-with-meson.sh
```diff
@@ -65,16 +65,52 @@ function main() {
 
     pushd "${SANDBOX_DIR}"
 
-    show_header "Run test suite"
-    meson configure -Dtests=true -Db_coverage=true -Dipc=true
+    show_header "Run ASAN/UBSAN test suite"
```
WillAyd, line 68: We might want to split these into separate jobs but punting that for now. I do think it would be nice to make the Meson job run as part of regular CI in a follow up if that's something you are open to.

Figure it doesn't hurt to have another tool to analyze the codebase, and I think Meson is a good deal easier to mess with than CMake. The downside of course is CMake isn't going anywhere anytime soon, so we do have active support for both

paleolimbot (approved):
Thank you!

PR #524, opened 2024-06-12
fix: Check for offset + length > int64_max before using the value to calculate buffer sizes

paleolimbot:
See https://github.com/apache/arrow-nanoarrow/pull/518#issuecomment-2161063096 . The overflow wasn't the one I'd suspected and it wasn't related to that PR (that PR/the one before it just added a test covering a case that didn't exist in our previous tests).

I also fixed some compiler warnings exposed by the meson build on linux/arm64 (where character is unsigned).


WillAyd (approved):
nice work

PR #527, opened 2024-06-14
fix: Ensure ArrowDeviceArray implementation for AppleMetal passes tests on newer MacOS

paleolimbot:
This PR fixes the Metal build (which was broken after a previous change), adds it to CI (thanks to the new GitHub M1 runners!) and fixes a failure I see locally (but not on CI) whereby the Metal runtime on some (newer? different?) MacOS environments can wrap arbitrary bytes as an `MTL::Buffer()`.

If true, this is awesome! But it does mean that in some cases we can "move" CPU arrays to the GPU without a copy and in some cases we have to copy into page-aligned memory (and that this can only be tested at runtime).

### DIFF
--- a/src/nanoarrow/nanoarrow_device_metal_test.cc
+++ b/src/nanoarrow/nanoarrow_device_metal_test.cc
@@ -106,7 +106,7 @@ TEST(NanoarrowDeviceMetal, DeviceGpuBufferMove) {
   old_ptr = buffer.data;
   ASSERT_EQ(ArrowDeviceBufferMove(cpu, &buffer, gpu, &buffer2), NANOARROW_OK);
   EXPECT_EQ(buffer2.size_bytes, 5);
-  EXPECT_NE(buffer2.data, old_ptr);
+  // EXPECT_NE(buffer2.data, old_ptr);
   EXPECT_EQ(memcmp(buffer2.data, data, sizeof(data)), 0);
   EXPECT_EQ(buffer.data, nullptr);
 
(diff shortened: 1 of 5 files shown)


src/nanoarrow/nanoarrow_device_metal_test.cc
```diff
@@ -237,15 +246,25 @@ TEST_P(StringTypeParameterizedTestFixture, ArrowDeviceMetalArrayViewString) {
 
   EXPECT_EQ(device_array_view.array_view.buffer_views[2].size_bytes, 7);
 
-  // We should be able to move arrays to Metal (this may depend on MacOS version
-  // ...wrapping an arbitrary sequence of bytes as an MTL::Buffer failed under
-  // previous versions of MacOS but now seems to work)
+  // In some MacOS environments, ArrowDeviceArrayMoveToDevice() with buffers allocated
+  // using ArrowMalloc() will work (i.e., apparently MTL::Buffer can wrap
+  // arbitrary bytes, but only sometimes)
```
raulcd, line 251: is this depending on macOS version or just randomly on the same version?

paleolimbot: I actually have no clue! I don't think it's MacOS version (becuase both me and the GitHub runner are on 14.5), but it's also not hardware (because I wrote the initial version on the same computer where the test is now failing).

Metal is sort of a mystery to most people, including me...I did this initially to ensure that the framework I'd come up with was going to work for something other than CUDA. Eventually I'll have to check with https://github.com/ml-explore/mlx to get a better handle on what's possible here.

PR #529, opened 2024-06-14
refactor(python): Extract utility functions into _utils.pyx

paleolimbot:
This PR moves a number of utility functions that were relatively self-contained into a new module, _utils.pyx.

TODO:

- dereplicate the error handling utility
- Ensure "exported" functions are documented with a docstring


jorisvandenbossche (approved):
Assuming the code changes in `_lib.pyx` and `_utils.pyx` are just copying from one place to another, this looks all good to me

(note that if it is only for code organization purposes, you could also include the utils file in `_lib.pyx` to keep it as a single module in practice)

jorisvandenbossche:
> Is keeping the single module helpful? (I'd assumed that since pyarrow was trying to split up its `lib.pyx` that it might be good to proactively split up this one before any more features were added; however, I might have misunderstood pyarrow's motivation).

Compilation time is a motivation for pyarrow AFAIK, which might not really be a concern (yet?) for nanoarrow. 

I was wondering a bit to what extent making this into two separate modules would increase the library size of both including nanoarrow c parts, but not entirely sure how this works

jorisvandenbossche:
>  I ran a check and it does seem like this PR adds about 100KB to the binary size. I think that is OK for now (and we can revisit some strategies for making that smaller if after separating the buffer, schema, array, and array stream portions it becomes unwieldy).

Sounds good

PR #530, opened 2024-06-17
refactor: Move Meson build targets to top level directory

WillAyd:
This makes the Meson build generate targets more closely to the CMake build. Instead of building targets in src/nanoarrow this makes it so it builds things in the build directory root

The only exception is the "nanoarrow_config.h" file which gets built to `<build_directory>/src`. The CMake installation creates a separate `generated` directory, but Meson practically limits the structure of your build folder to how the `meson.build` files are laid out in source. We could create a `generated` directory in the source if we so cared, or alternately could try to change the include to `#include <nanoarrow/nanoarrow_config.>` for consistency with all other includes, placing that generated target in the same directory as the rest of the headers


paleolimbot (approved):
Thank you!

PR #533, opened 2024-06-20
refactor(r): Use JSON in experimental R vctrs extension type

paleolimbot:
This PR updates the "vctrs extension type" to use JSON as its metadata serialization. JSON is use for most other extension types and using it here provides some workaround for non-R consumers that encounter this type elsewhere.

The serializer and deserializer use the same format as `jsonlite::serializeJSON()`, but restrict the types of objects that it is willing to serialize or deserialize.


amoeba:
This looks good @paleolimbot. I am getting one test failure locally though:

```
══ Results ═══════════════════════════════════════════════════════════════════════════
Duration: 4.6 s

── Failed tests ──────────────────────────────────────────────────────────────────────
Failure (test-extension.R:83:3): as_nanoarrow_array() dispatches on registered extension spec
`as_nanoarrow_array(...)` did not throw the expected error.

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 1426 ]
```

amoeba:
Yep, that did it. All tests pass now.

PR #534, opened 2024-06-20
fix(Python): Ensure we don't call cuMemAlloc with 0 bytesize

shwina:
According to the CUDA docs for [cuMemAlloc](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1gb82d2a09844a58dd9e744dc31e8aa467):

> If bytesize is 0, [cuMemAlloc()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1gb82d2a09844a58dd9e744dc31e8aa467) returns [CUDA_ERROR_INVALID_VALUE](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggc6c391505e117393cc2558fff6bfc2e990696c86fcee1f536a1ec7d25867feeb).

We end up calling `cuMemAlloc()` with `0` bytesize when allocating device buffers with no null mask. Thus, the following change was causing `nanoarrow_device_cuda_test` to fail:

```diff
@@ -207,7 +207,8 @@ TEST_P(StringTypeParameterizedTestFixture, ArrowDeviceCudaArrayViewString) {
   ASSERT_EQ(ArrowArrayStartAppending(&array), NANOARROW_OK);
   ASSERT_EQ(ArrowArrayAppendString(&array, ArrowCharView("abc")), NANOARROW_OK);
   ASSERT_EQ(ArrowArrayAppendString(&array, ArrowCharView("defg")), NANOARROW_OK);
-  ASSERT_EQ(ArrowArrayAppendNull(&array, 1), NANOARROW_OK);
+  ASSERT_EQ(ArrowArrayAppendString(&array, ArrowCharView("defg")), NANOARROW_OK);
+  // ASSERT_EQ(ArrowArrayAppendNull(&array, 1), NANOARROW_OK);
   ASSERT_EQ(ArrowArrayFinishBuildingDefault(&array, nullptr), NANOARROW_OK);
 
   ASSERT_EQ(ArrowDeviceArrayInit(cpu, &device_array, &array, nullptr), NANOARROW_OK);
```

In this PR, I've fixed this by simply skipping the call to `cuMemAlloc`. The resulting buffer will have `nullptr` as its `data` member and `0` as its `size_bytes`, which I believe is the desired outcome. 

I also modified the test above to include cases with no nulls.

### DIFF
--- a/src/nanoarrow/nanoarrow_device_cuda.c
+++ b/src/nanoarrow/nanoarrow_device_cuda.c
@@ -126,8 +126,7 @@ static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
       CUdeviceptr dptr = 0;
       if (size_bytes > 0) {  // cuMemalloc requires non-zero size_bytes
         err = cuMemAlloc(&dptr, (size_t)size_bytes);
-      }
-      else {
+      } else {
         err = CUDA_SUCCESS;
       }
       ptr = (void*)dptr;
--- a/src/nanoarrow/nanoarrow_device_cuda.c
+++ b/src/nanoarrow/nanoarrow_device_cuda.c
@@ -124,7 +124,12 @@ static ArrowErrorCode ArrowDeviceCudaAllocateBuffer(struct ArrowDevice* device,
   switch (device->device_type) {
     case ARROW_DEVICE_CUDA: {
       CUdeviceptr dptr = 0;
-      err = cuMemAlloc(&dptr, (size_t)size_bytes);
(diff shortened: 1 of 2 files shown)


shwina:
I see that this logic is being overhauled in https://github.com/apache/arrow-nanoarrow/pull/509/. If the maintainers want to close this PR out, that's fine by me! 

It's probably worth including the changes to the tests though.

Issue #309, opened 2023-09-26
Fix CMake deprecation warning


paleolimbot:
I am not sure exactly which `CMakeLists.txt` this should be fixed in, but many cmake configure calls now result in:

```
CMake Deprecation Warning at D:/a/arrow-nanoarrow/arrow-nanoarrow/nanoarrow-verify-tmp/build-ipc/_deps/googletest-src/CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.
  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
CMake Deprecation Warning at D:/a/arrow-nanoarrow/arrow-nanoarrow/nanoarrow-verify-tmp/build-ipc/_deps/googletest-src/googlemock/CMakeLists.txt:45 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.
  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
CMake Deprecation Warning at D:/a/arrow-nanoarrow/arrow-nanoarrow/nanoarrow-verify-tmp/build-ipc/_deps/googletest-src/googletest/CMakeLists.txt:56 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.
  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

paleolimbot:
These warnings are occurring only when building tests because we pin a rather old version of googletest. I don't think we need to update that version quite yet.

PR #535, opened 2024-06-21
Fix: CMake deprecation warnings from subprojects

WillAyd:
closes https://github.com/apache/arrow-nanoarrow/issues/309


paleolimbot (approved):
Sorry for the delay!

My earlier comment wasn't correct since this is correctly guarded by the GCC version check. This means that we still have one CI job checking C++11 compliance, so this is all good. Thank you!

PR #536, opened 2024-06-21
refactor: Separate components into folders under src/nanoarrow

paleolimbot:
The number of files in src/nanoarrow had accumulated quite a bit, which wasn't exactly a problem, but the files did have confusingly verbose names for historical reasons (mostly that the IPC and Device libraries originated elsewhere). This PR moves the `nanoarrow` target sources to `src/nanoarrow/common`, the `nanoarrow_ipc` target sources to `src/nanoarrow/ipc`, the `nanoarrow_device` target sources to `src/nanoarrow/device`, and the tests for the testing helpers to `src/nanoarrow/testing` (in the future this will/should be a `nanoarrow_testing` target).

I did leave the user-facing headers in src/nanoarrow, which makes a slightly cleaner transition between the "bundled" and CMake/Meson subproject builds (i.e., one always just does `#include "nanoarrow/nanoarrow.h"`)


assignUser (approved):
Very nice, especially that the user facing headers don't change :+1:

PR #541, opened 2024-06-25
chore(ci): Consolidate bundling and docs workflows into the packaging workflow

paleolimbot:
This PR consolidates a few existing workflows into `packaging.yaml` and removes the dist/ folder that had previously contained a cached version of the bundled files (updated nightly). This was confusing for a number of tools (e.g., codecov, languageserver) and is no longer really needed since the bundling is now very easy (I added the documentation to the README).

I also generally cleaned up the packaging workflow, which had duplicated the documentation build and had some things that ADBC needed but that nanoarrow doesn't.


README.md
```diff
@@ -62,6 +62,46 @@ The nanoarrow R package is available from [CRAN](https://cran.r-project.org):
...
+```
+
+...and declared as a dependency with:
+
+```
+nanoarrow_dep = dependency('nanoarrow')
+example_exec = executable('example_meson_minimal_app',
+                          'src/app.cc',
+                          dependencies: [nanoarrow_dep])
+```
```
paleolimbot, line 103: @WillAyd Did I get this part right?

WillAyd: Yep that's exactly it

WillAyd (approved):
Nice work - this is a huge simplification

.gitattributes
```diff
@@ -15,6 +15,5 @@
 # specific language governing permissions and limitations
 # under the License.
 
-dist/* linguist-generated
```
assignUser, line 18: :bulb: TIL about https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

.github/workflows/packaging.yaml
```diff
@@ -114,120 +105,87 @@ jobs:
...
-          echo "ref: ${{ github.ref }}"
-
       - name: Build documentation
         run: |
-          pushd nanoarrow
-          docker compose run --rm docs
-          popd
+          echo "::group::Docker Pull"
+          cd nanoarrow
+          docker compose run --rm -e GITHUB_ACTIONS docs
```
assignUser, line 113: You are never closing the group, is that intentional?
```suggestion
          echo "::group::Docker Pull"
          cd nanoarrow
          docker compose run --rm -e GITHUB_ACTIONS docs
          echo "::endgroup::"
```

.github/workflows/packaging.yaml
assignUser, line 1: Please explicitly limit/add permissions per job via a general workflow-wide
```yaml
permissions:
  contents:read
```
and then elevate where needed via https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions

assignUser (approved):
Some minor comments. LGTM

PR #542, opened 2024-06-25
fix: Meson install header files and pkgconfig

WillAyd:


meson.build
```diff
@@ -60,6 +73,12 @@ nanoarrow_dep = declare_dependency(include_directories: [incdir],
 if get_option('ipc')
     flatcc_dep = dependency('flatcc')
 
+    install_headers(
+        'src/nanoarrow/nanoarrow_ipc.h',
+        # 'src/nanoarrnow/nanoarrow_ipc_flatcc_generated.h',
```
WillAyd, line 78: I don't know a ton about flatbuffers but not sure if this file is required? I didn't see any internal usage

If we wanted it probably better to add a custom target to meson that runs the `./dev/update_fbs.sh` script, although Meson most likely will not want those updates to be done in the source tree

paleolimbot: > I don't know a ton about flatbuffers but not sure if this file is required? 

I think it is called `src/nanoarrow/ipc/flatcc_generated.h` now, and is included from the files in `src/nanoarrow/ipc`.

That is a generated file, but it doesn't need to be regenerated very often (whenever we update the flatcc dependency and whenever the Arrow Schema.fbs changes). I think it should not be regenerated every build (i.e., it should be done carefully and intentionally).

paleolimbot (approved):
Thank you! (And sorry for the massive refactor that caused the issue 😬 )

meson.build
paleolimbot: It may be missing there, too (currently no tests for the "installed" configuration for the IPC extension)

WillAyd: Ah OK - yea it looks like the CMake install fails

```sh
CMake Error at cmake_install.cmake:78 (file):
  file INSTALL cannot find
  "/home/willayd/clones/arrow-nanoarrow/src/nanoarrow/nanoarrow_ipc_flatcc_generated.h":
  No such file or directory.
```

WillAyd:
I'll take a look at setting up a ci job for installed components. Think that should be simple

PR #545, opened 2024-06-27
refactor(python): Split type identifier utilities into thier own module

paleolimbot:
This is another chapter in the saga to split up _lib.pyx to make it easier to find the piece of code one is looking for. Another advantage of this is that VSCode's Cython plugin does a much better job when `cimport`s are explicit: for something like `from nanoarrow_c cimport *`, this means that every use of a nanoarrow symbol (including `NANOARROW_TYPE_XXX`) would get a yellow squiggly underline (which normally highlight things that should be fixed). The setup here enables both Python and Cython modules to use `_types.UINT8`. In Cython this compiles into a C symbol reference (i.e., not an attribute lookup, which might be slow).


python/src/nanoarrow/_lib.pyx
```diff
@@ -830,63 +646,63 @@ cdef class CSchemaView:
...
-        if self._schema_view.type == NANOARROW_TYPE_TIMESTAMP:
+        if _types.equal(self._schema_view.type, _types.TIMESTAMP):
             return self._schema_view.timezone.decode()
         else:
             return None
 
     @property
     def union_type_ids(self):
-        if self._schema_view.type in CSchemaView._union_types:
+        if _types.has_time_unit(self._schema_view.type):
```
danepitkin, line 705: Copy/paste error? It should be `is_union`?

python/src/nanoarrow/_types.pyx
```diff
@@ -0,0 +1,244 @@
...
+            return True
+
+    return False
+
+
+cpdef bint is_unsigned_integer(int type_id):
+    """Check if type_id is an unsigned integral type"""
+    return type_id in (
+        _types.UINT8,
+        _types.UINT8,
```
danepitkin, line 53: duplicate entries for UINT8

danepitkin (commented):
Nice refactor! I really like it. Just a few small comments

PR #547, opened 2024-06-27
feat(python): Implement CUDA build in Python bindings

paleolimbot:
This PR updates the build system such that nanoarrow's CUDA capability can be exposed in Python (it doesn't quite expose any of the functionality yet). I'm not sure this is the best long-term strategy...I think we might be able to get CUDA plugged in at runtime because we don't use much of the API...but I think this would at least enable distributing a cuda-enabled build on conda-forge.

To build, one has to do something like:

```shell
export NANOARROW_PYTHON_CUDA_HOME=/usr/local/cuda
pip install .
```

On Windows, this would be something like:

```powershell
$Env:NANOARROW_PYTHON_CUDA_HOME = 'C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.5'
python -m pip install .
```

Then, you should be able to resolve a CUDA device:

```python
from nanoarrow import device

device.resolve(device.DeviceType.CUDA.value, 0)
#> <nanoarrow.device.Device>
#> - device_type: CUDA <2>
#> - device_id: 0
```


python/bootstrap.py
```diff
@@ -167,11 +167,58 @@ def _pxd_header(self):
...
+        output.write(b"    cdef int NANOARROW_MAX_FIXED_BUFFERS\n")
+        output.write(b"    cdef int ARROW_FLAG_DICTIONARY_ORDERED\n")
+        output.write(b"    cdef int ARROW_FLAG_NULLABLE\n")
+        output.write(b"    cdef int ARROW_FLAG_MAP_KEYS_SORTED\n")
+        output.write(b"\n")
+
+
+class NanoarrowDevicePxdGenerator(PxdGenerator):
+    def _preprocess_content(self, content):
+        self.device_names = re.findall("#define (ARROW_DEVICE_[A-Z0-9_]+)", content)
```
shwina, line 198: Neat!

shwina (approved):
Thanks! This looks like a very reasonable approach to me!

PR #550, opened 2024-06-28
feat(python): Implement array from buffer for non-CPU arrays

paleolimbot:
Requires building with (e.g.) `export NANOARROW_PYTHON_CUDA=/usr/local/cuda` and a `cupy` install:

```python
import nanoarrow as na
from nanoarrow import device
import cupy as cp


device.c_device_array(cp.array([1, 2, 3]))
#> <nanoarrow.device.CDeviceArray>
#> - device_type: CUDA <2>
#> - device_id: 0
#> - array: <nanoarrow.c_array.CArray int64>
#>   - length: 3
#>   - offset: 0
#>   - null_count: 0
#>   - buffers: (0, 133980798058496)
#>   - dictionary: NULL
#>   - children[0]:

# Also roundtrips
darray = device.c_device_array(cp.array([1, 2, 3]))
cp.from_dlpack(darray.array.view().buffer(1))
#> array([1, 2, 3])
```


danepitkin (commented):
LGTM! But I'll let someone with more non-CPU experience a chance to review.

danepitkin (approved):

python/src/nanoarrow/_utils.pyx
```diff
@@ -493,3 +493,29 @@ cdef object c_buffer_set_pybuffer(object obj, ArrowBuffer** c_buffer):
...
+cdef ArrowBufferAllocator c_pyobject_deallocator(object obj):
+    """ArrowBufferAllocator implementation wrapping a PyObject"""
+    Py_INCREF(obj)
+    return ArrowBufferDeallocator(
+        <ArrowBufferDeallocatorCallback>&c_deallocate_pyobject,
+        <void*>obj
+    )
+
+cdef void c_buffer_set_pyobject(object base, uint8_t* data, int64_t size_bytes, ArrowBuffer** c_buffer):
+    """Manage a Py_Buffer reference as an ArrowBuffer
```
jorisvandenbossche, line 512: This is copy pasted from above and not fully correct here?

python/src/nanoarrow/c_array.py
```diff
@@ -201,6 +202,7 @@ def c_array_from_buffers(
     children: Iterable[Any] = (),
     validation_level: Literal[None, "full", "default", "minimal", "none"] = None,
     move: bool = False,
+    device: Union[Device, None] = None,
```
jorisvandenbossche, line 205: Can you add this keyword to the docstring?

python/src/nanoarrow/c_array.py
```diff
@@ -298,7 +303,10 @@ def c_array_from_buffers(
     builder.resolve_null_count()
 
     # Validate + finish
-    return builder.finish(validation_level=validation_level)
+    if device is DEVICE_CPU:
+        return builder.finish(validation_level=validation_level)
+    else:
+        return builder.finish_device()
```
jorisvandenbossche, line 309: This means it returns a ArrowArray vs ArrowDeviceArray depending on the device? (but you might also want a device array for cpu device?)

jorisvandenbossche (commented):
Nice work! Added a few minor comments, and a few questions

python/src/nanoarrow/_array.pyx
```diff
@@ -107,7 +106,7 @@ cdef class CArrayView:
     def __cinit__(self, object base, uintptr_t addr):
         self._base = base
         self._ptr = <ArrowArrayView*>addr
-        self._device = DEVICE_CPU
+        self._event = CSharedSyncEvent(DEVICE_CPU)
```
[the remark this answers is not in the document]
paleolimbot: We still have the device here (it's one of the two things stored by the shared sync event), but we definitely need the ability to synchronize before buffer access. The other option would be to synchronize before constructing the array view but I gather that synchronizing is expensive (hence trying to avoid here).

> Should we make event an option param in the init here?

The constructors for all the classes are still a bit of a mess...pyarrow has a better system for this (e.g., most constructors have zero parameters and/or error, and internal constructors + static methods are used to construct).

python/src/nanoarrow/_buffer.pyx
```diff
@@ -569,6 +618,36 @@ cdef class CBuffer:
         out._populate_view()
         return out
 
+    @staticmethod
+    def from_dlpack(obj):
+        capsule = obj.__dlpack__()
```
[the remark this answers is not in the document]
paleolimbot: Probably, and we could probably do something with `cuStreamRecordEvent()` to create the `CSharedSyncEvent`!

python/src/nanoarrow/_buffer.pyx
```diff
@@ -163,15 +193,31 @@ cdef DLDevice view_to_dlpack_device(CBufferView view):
...
+    if arrow_device.device_type is ARROW_DEVICE_CPU:
+        # DLPack uses 0 for the CPU device id where Arrow uses -1
         device.device_type = kDLCPU
         device.device_id =  0
     else:
-        raise ValueError('Only CPU device is currently supported.')
+        # Otherwise, Arrow's device identifiers and types are intentionally
+        # identical to DLPack
+        device.device_type = <DLDeviceType>arrow_device.device_type
+        device.device_id = arrow_device.device_id
```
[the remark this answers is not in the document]
paleolimbot: Good call! I think we have to add something to the device support library to handle this since it's always an integer and has device-type specific meanings. I pushed something here but I don't think `stream` is supposed to be a `CUstream*` as a `uintptr_t` as I'd initially assumed.

python/tests/test_c_array.py
```diff
@@ -610,3 +622,91 @@ def test_c_array_duration():
...
     ]
+
+
+def test_device_array_errors():
+    from nanoarrow.device import DeviceType, resolve
+
+    try:
+        cuda_device = resolve(DeviceType.CUDA, 0)
+    except ValueError:
+        pytest.skip("CUDA device not available")
```
danepitkin, line 633: It could be nice to put this in a pytest.fixture if you want to make it reusable e.g.
```
@pytest.fixture
def cuda_device():
    from nanoarrow.device import resolve
    try:
        return resolve(DeviceType.CUDA, 0)
    except ValueError:
        pytest.skip("CUDA device not available")

def test_device_array_errors(cuda_device):
    ...
```

This should still skip the tests that require the pytest fixture. Add it to conftest.py to use it across test files.

danepitkin (approved):
Overall LGTM, with the caveat that I don't have much CUDA experience. Nice work!!

python/tests/test_c_array.py
danepitkin: This is a nit by the way. Consider it optional or for a future enhancement if you'd like!

paleolimbot: That is magic!

VSCode's testing integration gives bright red Xs for skips that happen in fixtures so I moved the actual skip to the test...but still a huge improvement!

python/tests/test_c_array.py
```diff
@@ -610,3 +622,84 @@ def test_c_array_duration():
         d2_duration_in_ms,
         d3_duration_in_ms,
     ]
+
+
+def test_device_array_errors(cuda_device):
+    if not cuda_device:
+        pytest.skip("CUDA device not available")
```
danepitkin, line 629: Awesome! If you wanted to, I _think_ you could add this check in the pytest fixture. Either way, looks great to me!

python/tests/test_c_array.py
danepitkin: Ahh that makes sense! Ignore my latest comment.

python/tests/test_c_array.py
paleolimbot: That works with `pytest` but not with VSCode's testing (which gives a bright red X for the entire test suite when tests are skipped in this way). Probably workaroundable!

danepitkin: I've add this to tests in pyarrow already 😬 haha.

danepitkin: Oops!

paleolimbot: I am not sure pyarrow works with VSCode's built-in testing integration for other reasons (but I haven't tried in a while!)

PR #555, opened 2024-07-10
feat: add ipc RecordBatch encoding

bkietz:
- added `ArrowIpcEncoderEncodeRecordBatch()`
- added `ArrowIpcEncoderBuildContiguousBodyBuffer()` to provide a default buffer encoder implementation which simply concatenates a batch's buffers into one contiguous (properly aligned and padded) body buffer
- testing uses the decoder tests, replacing arrow C++'s encoder with the new nanoarrow encoder

### DIFF
--- a/.github/workflows/r-check.yaml
+++ b/.github/workflows/r-check.yaml
@@ -58,6 +58,7 @@ jobs:
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
+          install-quarto: false
 
       # Check package install with extra warning flags except on Windows,
       # where the R headers cause some problems
--- a/.github/workflows/r-check.yaml
+++ b/.github/workflows/r-check.yaml
@@ -58,7 +58,6 @@ jobs:
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
-          install-quarto: 'false'
 
       # Check package install with extra warning flags except on Windows,
       # where the R headers cause some problems
--- a/.github/workflows/r-check.yaml
+++ b/.github/workflows/r-check.yaml
@@ -58,7 +58,7 @@ jobs:
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
-          install-quarto: false
+          install-quarto: 'false'
 
       # Check package install with extra warning flags except on Windows,
       # where the R headers cause some problems
--- a/src/nanoarrow/ipc/decoder_test.cc
+++ b/src/nanoarrow/ipc/decoder_test.cc
@@ -762,77 +762,28 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcArrowArrayRoundtrip) {
   ArrowIpcDecoderReset(&decoder);
 }
 
-struct ArrowArrayViewEqualTo {
-  const struct ArrowArrayView* expected;
+void AssertArrayViewIdentical(const struct ArrowArrayView* actual,
+                              const struct ArrowArrayView* expected) {
+  NANOARROW_DCHECK(actual->dictionary == nullptr);
+  NANOARROW_DCHECK(expected->dictionary == nullptr);
 
-  using is_gtest_matcher = void;
-
-  bool MatchAndExplain(const struct ArrowArrayView* actual, std::ostream* os) const {
-    return MatchAndExplain({}, actual, expected, os);
+  ASSERT_EQ(actual->storage_type, expected->storage_type);
+  ASSERT_EQ(actual->offset, expected->offset);
+  ASSERT_EQ(actual->length, expected->length);
+  for (int i = 0; i < 3; i++) {
+    auto a_buf = actual->buffer_views[i];
+    auto e_buf = expected->buffer_views[i];
+    ASSERT_EQ(a_buf.size_bytes, e_buf.size_bytes);
+    if (a_buf.size_bytes != 0) {
+      ASSERT_EQ(memcmp(a_buf.data.data, e_buf.data.data, a_buf.size_bytes), 0);
+    }
   }
 
-  static bool MatchAndExplain(std::vector<int> field_path,
-                              const struct ArrowArrayView* actual,
-                              const struct ArrowArrayView* expected, std::ostream* os) {
-    auto prefixed = [&]() -> std::ostream& {
-      if (!field_path.empty()) {
-        for (int i : field_path) {
-          *os << "." << i;
-        }
-        *os << ":";
-      }
-      return *os;
-    };
-
-    NANOARROW_DCHECK(actual->offset == 0);
-    NANOARROW_DCHECK(expected->offset == 0);
-
-    if (actual->length != expected->length) {
-      prefixed() << "expected length=" << expected->length << "\n";
-      prefixed() << "  actual length=" << actual->length << "\n";
-      return false;
-    }
-
-    auto null_count = [](const struct ArrowArrayView* a) {
-      return a->null_count != -1 ? a->null_count : ArrowArrayViewComputeNullCount(a);
-    };
-    if (null_count(actual) != null_count(expected)) {
-      prefixed() << "expected null_count=" << null_count(expected) << "\n";
-      prefixed() << "  actual null_count=" << null_count(actual) << "\n";
-      return false;
-    }
-
-    for (int64_t i = 0; actual->layout.buffer_type[i] != NANOARROW_BUFFER_TYPE_NONE &&
-                        i < NANOARROW_MAX_FIXED_BUFFERS;
-         ++i) {
-      auto a_buf = actual->buffer_views[i];
-      auto e_buf = expected->buffer_views[i];
-      if (a_buf.size_bytes != e_buf.size_bytes) {
-        prefixed() << "expected buffer[" << i << "].size=" << e_buf.size_bytes << "\n";
-        prefixed() << "  actual buffer[" << i << "].size=" << a_buf.size_bytes << "\n";
-        return false;
-      }
-      if (memcmp(a_buf.data.data, e_buf.data.data, a_buf.size_bytes) != 0) {
-        prefixed() << "expected buffer[" << i << "]'s data to match\n";
-        return false;
-      }
-    }
-
-    field_path.push_back(0);
-    for (int64_t i = 0; i < actual->n_children; i++) {
-      field_path.back() = i;
-      if (!MatchAndExplain(field_path, actual->children[i], expected->children[i], os)) {
-        return false;
-      }
-    }
-    return true;
+  ASSERT_EQ(actual->n_children, expected->n_children);
+  for (int i = 0; i < actual->n_children; i++) {
+    AssertArrayViewIdentical(actual->children[i], expected->children[i]);
   }
-
-  void DescribeTo(std::ostream* os) const { *os << "is equivalent to the array view"; }
-  void DescribeNegationTo(std::ostream* os) const {
-    *os << "is not equivalent to the array view";
-  }
-};
+}
 
 TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcNanoarrowArrayRoundtrip) {
   struct ArrowError error;
@@ -884,7 +835,7 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcNanoarrowArrayRoundtrip) {
                                              -1, &roundtripped, nullptr),
               NANOARROW_OK);
 
-    EXPECT_THAT(roundtripped, ArrowArrayViewEqualTo{array_view.get()});
+    AssertArrayViewIdentical(roundtripped, array_view.get());
   }
 }
 
--- a/src/nanoarrow/ipc/decoder_test.cc
+++ b/src/nanoarrow/ipc/decoder_test.cc
@@ -819,7 +819,7 @@ struct ArrowArrayViewEqualTo {
     }
 
     field_path.push_back(0);
-    for (int64_t i = 0; i < actual->n_children; ++i) {
+    for (int64_t i = 0; i < actual->n_children; i++) {
       field_path.back() = i;
       if (!MatchAndExplain(field_path, actual->children[i], expected->children[i], os)) {
         return false;
@@ -841,6 +841,10 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcNanoarrowArrayRoundtrip) {
       arrow::ExportSchema(arrow::Schema({arrow::field("", GetParam())}), schema.get())
           .ok());
 
+  nanoarrow::UniqueArrayView array_view;
+  ASSERT_EQ(ArrowArrayViewInitFromSchema(array_view.get(), schema.get(), &error),
+            NANOARROW_OK);
+
   // now make one empty struct array with this schema and another with all zeroes
   nanoarrow::UniqueArray empty_array, zero_array;
   for (auto* array : {empty_array.get(), zero_array.get()}) {
@@ -850,10 +854,6 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcNanoarrowArrayRoundtrip) {
       ASSERT_EQ(ArrowArrayAppendEmpty(array, 5), NANOARROW_OK);
     }
     ASSERT_EQ(ArrowArrayFinishBuildingDefault(array, nullptr), NANOARROW_OK);
-
-    nanoarrow::UniqueArrayView array_view;
-    ASSERT_EQ(ArrowArrayViewInitFromSchema(array_view.get(), schema.get(), &error),
-              NANOARROW_OK);
     ASSERT_EQ(ArrowArrayViewSetArray(array_view.get(), array, &error), NANOARROW_OK)
         << error.message;
 
@@ -861,8 +861,8 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcNanoarrowArrayRoundtrip) {
     EXPECT_EQ(ArrowIpcEncoderInit(encoder.get()), NANOARROW_OK);
 
     nanoarrow::UniqueBuffer buffer, body_buffer;
-    ArrowIpcEncoderBuildContiguousBodyBuffer(encoder.get(), body_buffer.get());
-    EXPECT_EQ(ArrowIpcEncoderEncodeRecordBatch(encoder.get(), array_view.get(), &error),
+    EXPECT_EQ(ArrowIpcEncoderEncodeSimpleRecordBatch(encoder.get(), array_view.get(),
+                                                     body_buffer.get(), &error),
               NANOARROW_OK)
         << error.message;
     EXPECT_EQ(
--- a/src/nanoarrow/ipc/decoder_test.cc
+++ b/src/nanoarrow/ipc/decoder_test.cc
@@ -21,6 +21,7 @@
 #include <arrow/c/bridge.h>
 #include <arrow/ipc/api.h>
 #include <arrow/util/key_value_metadata.h>
+#include <gmock/gmock-matchers.h>
 #include <gtest/gtest.h>
 
 // For bswap32()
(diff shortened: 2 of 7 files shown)


src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -379,6 +381,49 @@ ArrowErrorCode ArrowIpcArrayStreamReaderInit(
...
+///
+/// Returns ENOMEM if allocation fails, NANOARROW_OK otherwise.
+ArrowErrorCode ArrowIpcEncodeRecordBatch(
+    struct ArrowBuffer* out,
+    struct ArrowArray* to_be_encoded,
+    ArrowIpcBufferEncodeCallback encode_buffer,
+    void *user_data, struct ArrowError* error);
+
+/// \brief Encapsulate a Message stored in a buffer by prepending the continuation indicator and size.
+ArrowErrorCode ArrowIpcEncodeEncapsulatedMessage(struct ArrowBuffer* buffer);
```
paleolimbot, line 425: This is great for now, although I wonder if `ArrowErrorCode ArrowIpcEncapsulateMessage(struct ArrowBuffer* buffer, struct ArrowIpcOutputStream* output_stream);` is closer to what we'll actually need for the high-level version of this that writes out an `ArrowArrayStream`.

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,446 @@
...
+    FLATCC_RETURN_UNLESS_0(Field_custom_metadata_start(builder));
+    NANOARROW_RETURN_NOT_OK(
+        ArrowIpcEncodeMetadata(builder, schema, &ns(Field_custom_metadata_push_start),
+                               &ns(Field_custom_metadata_push_end), error));
+    FLATCC_RETURN_UNLESS_0(Field_custom_metadata_end(builder));
+  }
+  return NANOARROW_OK;
+}
+
+ArrowErrorCode ArrowIpcEncodeSchemaImpl(flatcc_builder_t* builder,
```
paleolimbot, line 311:
```suggestion
static ArrowErrorCode ArrowIpcEncodeSchemaImpl(flatcc_builder_t* builder,
```

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,446 @@
...
+                                             void* user_data, struct ArrowError* error) {
+  FLATCC_RETURN_UNLESS_0(Message_start_as_root(builder));
+  FLATCC_RETURN_UNLESS_0(Message_version_add(builder, ns(MetadataVersion_V5)));
+
+  int64_t body_length = 0;
+
+  FLATCC_RETURN_UNLESS_0(Message_header_RecordBatch_start(builder));
+
+  // TODO add compression support; should be a parameter to EncodeRecordBatch since
+  //      encoding of buffers is handled by the callback
```
paleolimbot, line 395: I think this is a good argument for an `ArrowIpcEcoder` (since compression and dictionary support will both result in extra arguments getting added here, and potentially cached state)

src/nanoarrow/ipc/encoder_test.cc
```diff
@@ -0,0 +1,137 @@
...
+
+TEST(NanoarrowIpcTest, NanoarrowIpcEncodeSimpleSchemas) {
+  struct ArrowError error;
+  nanoarrow::UniqueBuffer buffer;
+
+  // null buffer, schema don't segfault
+  EXPECT_EQ(ArrowIpcEncodeSchema(nullptr, to_unique(Schema({})).get(), &error), EINVAL);
+  EXPECT_EQ(ArrowIpcEncodeSchema(buffer.get(), nullptr, &error), EINVAL);
+
+  FieldVector boring_fields{
```
paleolimbot, line 66: A potentially low-effort way to get a lot of coverage would be to just add your encoder into the existing decoder tests to check roundtrip (would be easier if our testing library had a more straightforward schema comparer)

paleolimbot (commented):
I know you're still working on this, just a few preliminary things to think about!

In general, I do think we are going to need an `ArrowIpcEncoder` to avoid painful refactors or excessive numbers of arguments when we add additional features. It may also help collapse `status = ArrowXXXX()` into `NANOARROW_RETURN_NOT_OK()` (by making the current flatcc builder something that's cleaned up on `ArrowIpcEncoderReset()`).

src/nanoarrow/nanoarrow_ipc.h
bkietz: I am hoping to leave full stream support for a follow up

src/nanoarrow/ipc/encoder_test.cc
bkietz: I'll copy those cases here

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,518 @@
...
+      (struct ArrowIpcEncoderPrivate*)encoder->private_data;
+  ArrowBufferReset(out);
+  size_t size = 0;
+  out->data = (uint8_t*)flatcc_builder_finalize_buffer(&private->builder, &size);
+  out->size_bytes = out->capacity_bytes = (int64_t)size;
+  return out->data ? NANOARROW_OK : ENOMEM;
+}
+
+ArrowErrorCode ArrowIpcEncodeEncapsulatedMessage(struct ArrowBuffer* buffer) {
+  if (!buffer) return EINVAL;
```
paleolimbot, line 98: nit: This is possibly a better way to do this than what we currently do, but every other nanoarrow function uses `NANOARROW_DCHECK()` for this (if we check arguments at all). In other words, we currently crash for programming errors and reserve EINVAL for inappropriate values (mostly).

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,518 @@
...
+
+      FLATCC_RETURN_UNLESS_0(
+          Union_mode_add(builder, schema_view->type == NANOARROW_TYPE_DENSE_UNION));
+      if (schema_view->union_type_ids) {
+        int8_t type_ids[128];
+        int n = _ArrowParseUnionTypeIds(schema_view->union_type_ids, type_ids);
+        if (n != 0) {
+          FLATCC_RETURN_UNLESS_0(Union_typeIds_start(builder));
+          int32_t* type_ids_32 = (int32_t*)ns(Union_typeIds_extend(builder, n));
+          if (!type_ids_32) return ENOMEM;
```
paleolimbot, line 296: nit: the Google style guide seems like it allows these types of statements (i.e., an `if` without curly braces) for "historical reasons", but all of our existing code uses (I think!) uses curly braces with the body on separate lines.

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,518 @@
...
+  features[0] = ns(Feature_COMPRESSED_BODY);
+  FLATCC_RETURN_UNLESS_0(Schema_features_end(builder));
+
+  FLATCC_RETURN_UNLESS_0(Message_header_Schema_end(builder));
+
+  FLATCC_RETURN_UNLESS_0(Message_bodyLength_add(builder, 0));
+  return ns(Message_end_as_root(builder)) ? NANOARROW_OK : ENOMEM;
+}
+
+ArrowErrorCode ArrowIpcEncoderEncodeRecordBatchImpl(
```
paleolimbot, line 433:
```suggestion
static ArrowErrorCode ArrowIpcEncoderEncodeRecordBatchImpl(
```

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,518 @@
...
+    struct ArrowIpcEncoder* encoder, const struct ArrowArrayView* array_view,
+    struct ArrowBuffer* buffers, struct ArrowBuffer* nodes, struct ArrowError* error) {
+  if (array_view->offset != 0) {
+    ArrowErrorSet(error, "Cannot encode arrays with nonzero offset");
+    return ENOTSUP;
+  }
+
+  for (int c = 0; c < array_view->n_children; ++c) {
```
paleolimbot, line 441:
```suggestion
  for (int64_t c = 0; c < array_view->n_children; ++c) {
```

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,518 @@
...
+  }
+  return NANOARROW_OK;
+}
+
+ArrowErrorCode ArrowIpcEncoderEncodeRecordBatch(struct ArrowIpcEncoder* encoder,
+                                                const struct ArrowArrayView* array_view,
+                                                struct ArrowError* error) {
+  if (!encoder || !encoder->private_data || !array_view || !error) return EINVAL;
+
+  if (array_view->null_count != 0) {
```
paleolimbot, line 465: We should probably make a function for this (I think there's one in Python that should maybe move to C), but we might have to compute the null count here if `null_count` is -1. For the purposes of this PR, checking for a NULL validity buffer will work for almost all record batch implementations (maybe not C#).

src/nanoarrow/ipc/encoder_test.cc
```diff
@@ -0,0 +1,313 @@
...
+  char first_byte;
+  memcpy(&first_byte, &check, sizeof(char));
+  if (first_byte) {
+    return NANOARROW_IPC_ENDIANNESS_LITTLE;
+  } else {
+    return NANOARROW_IPC_ENDIANNESS_BIG;
+  }
+}
+
+struct TestDecoder : ArrowIpcDecoder {
```
paleolimbot, line 42: I think there is already a `nanoarrow:UniqueIpcDecoder`?

https://github.com/apache/arrow-nanoarrow/blob/2aa2e697fd5d0ef3cd88961bb030f9e760db581b/src/nanoarrow/nanoarrow_ipc.hpp#L78

src/nanoarrow/ipc/encoder_test.cc
```diff
@@ -0,0 +1,313 @@
...
+  Buffer metadata_buf{metadata->data, metadata->size_bytes};
+  io::BufferReader body_reader{std::make_shared<Buffer>(body->data, body->size_bytes)};
+  auto maybe_batch = ipc::ReadRecordBatch(
+      metadata_buf, arrow::schema(schema.fields(), schema.metadata()), nullptr, {},
+      &body_reader);
+  EXPECT_TRUE(maybe_batch.ok()) << maybe_batch.status();
+  return std::move(maybe_batch).ValueOr(nullptr);
+}
+
+auto random_batch(const Schema& schema, int64_t num_rows) {
```
paleolimbot, line 84: I am skeptical to add additional Arrow C++ surface since the dependency in our tests is something I'd like to remove in the near future (although I'm more than happy to leave this and replace it later when I through and do that refactor). The integration tests will take care of sending random batches through IPC...here I think you can get full coverage with an array with a zero-size buffer, an array with more than one non-empty buffer, and an array with more than one non-empty child.

paleolimbot (commented):
A few more thoughts! Feel free to leave the `RecordBatch` encoding to a follow-up PR (it would be easier to review, although I'm also happy to review together if that's easier for you!).

src/nanoarrow/ipc/encoder.c
bkietz: I'll add braces

src/nanoarrow/ipc/encoder_test.cc
bkietz: I'll reuse that then

src/nanoarrow/common/inline_array.h
```diff
@@ -740,6 +740,31 @@ static inline int8_t ArrowArrayViewIsNull(const struct ArrowArrayView* array_vie
   }
 }
 
+static inline int64_t ArrowArrayViewGetNullCount(
```
paleolimbot, line 743: This would also need an entry in `nanoarrow.h` with documentation. I have a slight preference towards having this be `ArrowArrayViewComputeNullCount()` (with the user choosing if they want to trust the existing null count or not).

A dedicated PR for this would probably make sense (it could also use this to replace some Python-specific functionality: https://github.com/apache/arrow-nanoarrow/blob/fcf3a809f8b8f8facfb8f29284c006429cc91d49/python/src/nanoarrow/_lib.pyx#L355-L373 )

src/nanoarrow/common/inline_array.h
```diff
@@ -740,6 +740,31 @@ static inline int8_t ArrowArrayViewIsNull(const struct ArrowArrayView* array_vie
...
+    case NANOARROW_TYPE_DENSE_UNION:
+    case NANOARROW_TYPE_SPARSE_UNION:
+      // Unions are "never null" in Arrow land
+      return 0;
+    default:
+      break;
+  }
+
+  const uint8_t* validity_buffer = array_view->buffer_views[0].data.as_uint8;
+  if (validity_buffer) {
```
paleolimbot, line 761: For better or worse, all of our existing null checks in nanoarrow are explicit (i.e., `if (validity_buffer == NULL) {`)

src/nanoarrow/common/inline_array.h
```diff
@@ -740,6 +740,31 @@ static inline int8_t ArrowArrayViewIsNull(const struct ArrowArrayView* array_vie
...
+    return array_view->null_count;
+  }
+
+  switch (array_view->storage_type) {
+    case NANOARROW_TYPE_NA:
+      return array_view->length;
+    case NANOARROW_TYPE_DENSE_UNION:
+    case NANOARROW_TYPE_SPARSE_UNION:
+      // Unions are "never null" in Arrow land
+      return 0;
```
paleolimbot, line 755: Are these cases tested?

paleolimbot (commented):
Thanks!

The structure of this is great. I know I said earlier that one big PR was OK, but the scope has enlarged to the point where I would prefer to review as multiple PRs. In particular, it makes it easier to ensure full test coverage for each component (also vaguely better for the changelog). Perhaps:

- Add array view null counter
- Add encoder.c with the encoder infrastructure and test for create/destroy
- Add schema writing
- Add batch writing + unit tests 
- Add encoder to the ipc_files test (which is our current infrastructure for sending random batches of various types through our IPC library)

I know that there are some testing patterns and style differences with Arrow C++ that take some getting used to here. I'm absolutely open to changing those; however, if they do change they should change in the rest of the code base too (in dedicated PRs). I think the smaller PRs will help with that (to slowly get used to the existing conventions without having to rewrite things).

bkietz:
Rebased on https://github.com/apache/arrow-nanoarrow/pull/568

src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -763,6 +764,132 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcArrowArrayRoundtrip) {
   ArrowIpcDecoderReset(&decoder);
 }
 
+struct ArrowArrayViewEqualTo {
```
paleolimbot, line 767: This is awesome! The negative matches and messages here are also not tested and I am not sure anybody looking to see if we had a utility to help with array equality would look in `decoder_test.cc` to find it. Something like:

```c
void AssertArrayViewIdentical(actual, expected) {
  NANOARROW_DCHECK(actual->dictionary != nullptr);
  NANOARROW_DCHECK(expected->dictionary != nullptr);

  ASSERT_EQ(actual->storage_type, expected->storage_type);
  ASSERT_EQ(actual->offset, expected->offset);
  ASSERT_EQ(actual->length, expected->length);
  for (int i = 0; i < 3; i++) {
    auto a_buf = actual->buffer_views[i];
    auto e_buf = expected->buffer_views[i];
    ASSERT_EQ(a_buf.size_bytes,  e_buf->size_bytes);
    if (a_buf.size_bytes != 0) {
      ASSERT_EQ(memcmp(a_buf.data.data, e_buf.data.data, a_buf.size_bytes), 0);
    }
  }

  ASSERT_EQ(actual->n_children, expected->n_children);
  for (int i = 0; i < actual->n_children; i++) {
    AssertArrayViewIdentical(actual->children[i], expected->children[i]);
  }
}
```

...will give terrible error messages but can be followed up with a tested version of this helper (or a version that partly lives in the C library since this comes up in C situations as well).

src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -461,6 +465,21 @@ ArrowErrorCode ArrowIpcEncoderFinalizeBuffer(struct ArrowIpcEncoder* encoder,
...
+void ArrowIpcEncoderBuildContiguousBodyBuffer(struct ArrowIpcEncoder* encoder,
+                                              struct ArrowBuffer* body_buffer);
+
+/// \brief Encode a struct typed ArrayView to a flatbuffer RecordBatch, embedded in a
+/// Message.
+///
+/// Returns ENOMEM if allocation fails, NANOARROW_OK otherwise.
+ArrowErrorCode ArrowIpcEncoderEncodeRecordBatch(struct ArrowIpcEncoder* encoder,
+                                                const struct ArrowArrayView* array_view,
+                                                struct ArrowError* error);
```
paleolimbot, line 482: I wonder if this would be simpler as:

```c
ArrowErrorCode ArrowIpcEncoderEncodeSimpleRecordBatch(struct ArrowIpcEncoder* encoder,
                                                const struct ArrowArrayView* array_view,
                                                struct ArrowBuffer* out,
                                                struct ArrowError* error);
```

...with a potential future

```c
ArrowErrorCode ArrowIpcEncoderEncodeRecordBatch(struct ArrowIpcEncoder* encoder,
                                                const struct ArrowArrayView* array_view,
                                                struct ArrowIpcBufferEncoder* encoder,
                                                struct ArrowError* error);
```

...to handle advanced usage of the dissociated stream stuff where the offsets aren't local.

src/nanoarrow/ipc/encoder.c
```diff
@@ -424,3 +423,135 @@ ArrowErrorCode ArrowIpcEncoderEncodeSchema(struct ArrowIpcEncoder* encoder,
...
   return ns(Message_end_as_root(builder)) ? NANOARROW_OK : ENOMEM;
 }
+
+static ArrowErrorCode ArrowIpcEncoderBuildContiguousBodyBufferCallback(
+    struct ArrowBufferView buffer_view, struct ArrowIpcEncoder* encoder, int64_t* offset,
+    int64_t* length, struct ArrowError* error) {
+  struct ArrowBuffer* body_buffer = (struct ArrowBuffer*)encoder->encode_buffer_state;
+
+  int compressed_buffer_header =
+      encoder->codec != NANOARROW_IPC_COMPRESSION_TYPE_NONE ? sizeof(int64_t) : 0;
```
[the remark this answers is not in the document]
bkietz: I think it suffices for now. Even if we later extend with more advanced heuristics for determining if a buffer should be compressed, the IPC format only allows a single codec to be provided per schema/stream (which means we'll still have only one codec per encoder).

src/nanoarrow/ipc/encoder.c
```diff
@@ -424,3 +423,135 @@ ArrowErrorCode ArrowIpcEncoderEncodeSchema(struct ArrowIpcEncoder* encoder,
...
+    NANOARROW_RETURN_NOT_OK(
+        ArrowIpcEncoderEncodeRecordBatchImpl(encoder, child, buffers, nodes, error));
+  }
+  return NANOARROW_OK;
+}
+
+ArrowErrorCode ArrowIpcEncoderEncodeRecordBatch(struct ArrowIpcEncoder* encoder,
+                                                const struct ArrowArrayView* array_view,
+                                                struct ArrowError* error) {
+  NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL && schema != NULL);
```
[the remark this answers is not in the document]
bkietz: nope, that's a copy paste error. I found this in [another PR](https://github.com/apache/arrow-nanoarrow/pull/571/files#diff-3ec78d215b788b99da9e08463bcb818fc9ded78702b994962fdc50c9739fe4f3R317). It seems that `NANOARROW_DEBUG` isn't getting defined

src/nanoarrow/ipc/encoder.c
```diff
@@ -80,8 +76,11 @@ ArrowErrorCode ArrowIpcEncoderFinalizeBuffer(struct ArrowIpcEncoder* encoder,
   struct ArrowIpcEncoderPrivate* private =
       (struct ArrowIpcEncoderPrivate*)encoder->private_data;
 
-  int64_t size = (int64_t)flatcc_builder_get_buffer_size(&private->builder);
-  int32_t header[] = {-1, ArrowInt32ToLe((int32_t)size)};
+  int32_t size = (int32_t)flatcc_builder_get_buffer_size(&private->builder);
```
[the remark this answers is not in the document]
bkietz: I'm not sure what error you mean; overflowing `int32_t`?

src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -763,6 +764,132 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcArrowArrayRoundtrip) {
...
+    ASSERT_EQ(ArrowArrayInitFromSchema(array, schema.get(), nullptr), NANOARROW_OK);
+    ASSERT_EQ(ArrowArrayStartAppending(array), NANOARROW_OK);
+    if (array == zero_array.get()) {
+      ASSERT_EQ(ArrowArrayAppendEmpty(array, 5), NANOARROW_OK);
+    }
+    ASSERT_EQ(ArrowArrayFinishBuildingDefault(array, nullptr), NANOARROW_OK);
+
+    nanoarrow::UniqueArrayView array_view;
+    ASSERT_EQ(ArrowArrayViewInitFromSchema(array_view.get(), schema.get(), &error),
+              NANOARROW_OK);
```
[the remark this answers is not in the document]
bkietz: didn't know that; it seemed like initfromschema always needed to precede setarray. Thanks!

src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -763,6 +764,132 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcArrowArrayRoundtrip) {
...
+    };
+    if (null_count(actual) != null_count(expected)) {
+      prefixed() << "expected null_count=" << null_count(expected) << "\n";
+      prefixed() << "  actual null_count=" << null_count(actual) << "\n";
+      return false;
+    }
+
+    for (int64_t i = 0; actual->layout.buffer_type[i] != NANOARROW_BUFFER_TYPE_NONE &&
+                        i < NANOARROW_MAX_FIXED_BUFFERS;
+         ++i) {
```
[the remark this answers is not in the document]
bkietz: okay

src/nanoarrow/ipc/encoder.c
```diff
@@ -424,3 +423,135 @@ ArrowErrorCode ArrowIpcEncoderEncodeSchema(struct ArrowIpcEncoder* encoder,
   FLATCC_RETURN_UNLESS_0(Message_bodyLength_add(builder, 0));
   return ns(Message_end_as_root(builder)) ? NANOARROW_OK : ENOMEM;
 }
+
+static ArrowErrorCode ArrowIpcEncoderBuildContiguousBodyBufferCallback(
+    struct ArrowBufferView buffer_view, struct ArrowIpcEncoder* encoder, int64_t* offset,
+    int64_t* length, struct ArrowError* error) {
+  struct ArrowBuffer* body_buffer = (struct ArrowBuffer*)encoder->encode_buffer_state;
```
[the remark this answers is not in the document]
bkietz: okay

src/nanoarrow/ipc/encoder.c
```diff
@@ -424,3 +423,135 @@ ArrowErrorCode ArrowIpcEncoderEncodeSchema(struct ArrowIpcEncoder* encoder,
...
+
+  // store offset and length of the buffer
+  *offset = buffer_begin;
+  *length = buffer_view.size_bytes;
+
+  if (compressed_buffer_header) {
+    // Signal that the buffer is not compressed; eventually we will set this to the
+    // decompressed length of an actually compressed buffer.
+    NANOARROW_RETURN_NOT_OK(ArrowBufferAppendInt64(body_buffer, -1));
+  }
```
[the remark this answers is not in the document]
bkietz: Would it really be preferable to raise an error for "codec not supported?"

src/nanoarrow/nanoarrow_ipc.h
bkietz: okay

src/nanoarrow/ipc/decoder_test.cc
bkietz: can we leave extraction of an array equality helper for a follow up?

.github/workflows/r-check.yaml
```diff
@@ -59,6 +59,7 @@ jobs:
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
+          install-quarto: false
```
bkietz, line 62: GH actions seems to have been broken by https://github.com/r-lib/actions/blame/d1a6cbabea8dd2f137598ba2a70ae37ac82d7941/setup-r-dependencies/action.yaml#L215

which tries setting up quarto by default. I'm not sure how else to disable this

bkietz: ... or alternatively to allow quarto to setup https://github.com/apache/arrow-nanoarrow/actions/runs/10220526007/job/28281045091?pr=555#step:1:39

src/nanoarrow/ipc/decoder_test.cc
paleolimbot: Yes! But since you still need something to test the roundtrip, this suggestion was to use something more compact than the equality helper that is currently here.

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: Ah, that would be:

https://github.com/apache/arrow-nanoarrow/blob/f74d57cd742c58a0563d51ac30ab96f1dc9576b5/CMakeLists.txt#L148

...that doesn't exist for `nanoarrow_ipc`. (No need to fix that here since it's not your fault 😬 )

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: This is narrowing a `size_t` down to an `int32_t`. A debug check is probably ok since it would take a few million columns to hit this (but there is also `_NANOARROW_CHECK_UPPER_LIMIT()`).

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: I had in mind something more like how it is specified in Arrow C++ (as a member of an Options struct that includes some other options about IPC writing). I don't see any advantages to premptively adding references to a feature that doesn't exist yet (and might never exist, although I hope that it does).

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: I had in mind just removing all references to compression for now including the ability to specify a compression (they seem easy enough to add back in if or when that feature is added)

src/nanoarrow/ipc/encoder.c
```diff
@@ -424,3 +423,135 @@ ArrowErrorCode ArrowIpcEncoderEncodeSchema(struct ArrowIpcEncoder* encoder,
...
+  ArrowBufferResize(&private->nodes, 0, 0);
+  NANOARROW_RETURN_NOT_OK(ArrowIpcEncoderEncodeRecordBatchImpl(
+      encoder, array_view, &private->buffers, &private->nodes, error));
+
+  FLATCC_RETURN_UNLESS_0(RecordBatch_nodes_create(  //
+      builder, (struct ns(FieldNode)*)private->nodes.data,
+      private->nodes.size_bytes / sizeof(struct ns(FieldNode))));
+  FLATCC_RETURN_UNLESS_0(RecordBatch_buffers_create(  //
+      builder, (struct ns(Buffer)*)private->buffers.data,
+      private->buffers.size_bytes / sizeof(struct ns(Buffer))));
```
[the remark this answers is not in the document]
paleolimbot: I was wondering if we could avoid creating the new root/fields and copying our cached buffer into it (updating in place instead), but flatcc is probably be very good at anticipating repeated message building and avoiding allocations (and the copy is probably not expensive).

src/nanoarrow/ipc/encoder.c
```diff
@@ -27,31 +27,36 @@
 
 #define ns(x) FLATBUFFERS_WRAP_NAMESPACE(org_apache_arrow_flatbuf, x)
 
-#define FLATCC_RETURN_UNLESS_0(x) \
-  if (ns(x) != 0) return ENOMEM;
+#define FLATCC_RETURN_UNLESS_0_NO_NS(x)                               \
+  if ((x) != 0) {                                                     \
+    ArrowErrorSet(error, "%s:%d: %s failed", __FILE__, __LINE__, #x); \
```
paleolimbot, line 32: Even though `ArrowError*` is usually called `error`, there is some prior art that uses a macro argument to handle this elsewhere (e.g., `NANOARROW_RETURN_NOT_OK_WITH_ERROR()`). (I wasn't sure the error was getting updated when reading the below code)

.github/workflows/r-check.yaml
paleolimbot: I don't know either! No need to solve that here (perhaps @assignUser knows if the quarto action is or will be whitelisted since I think it is needed for some Arrow R package jobs).

assignUser: hm not sure if we use quarto in any of our jobs, if we do I can get it allow-listed :+1:

src/nanoarrow/ipc/decoder_test.cc
paleolimbot: Thanks! In the meantime I bet we can pin the sha of the action before the change or something (I'll try Monday)

.github/workflows/r-check.yaml
paleolimbot: https://github.com/apache/arrow-nanoarrow/pull/572 is a workaround...I don't know if we have a direct use of quarto, but with the current head of the v2 branch there is a step that lists the quarto action (even though it isn't run, since we don't have any .qmd files):

https://github.com/r-lib/actions/blame/d1a6cbabea8dd2f137598ba2a70ae37ac82d7941/setup-r-dependencies/action.yaml#L240C26-L240C49

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
bkietz: okay

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
bkietz: okay

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
bkietz: I'll remove compression support for now. I'm not sure how an options struct would be advantageous here, but we can hash that out later

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
bkietz: I'm not sure what state flatcc maintains inside its builders. Some of the example code suggests we could have it maintain independent vectors which are not specifically part of a field which is currently being built. I've tried to use that a few times and gotten odd errors when I strayed from strict state machine style, so I'd prefer to experiment independent of other changes in a follow up.

src/nanoarrow/ipc/encoder.c
bkietz: I'll make `error` an explicit argument

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: Agreed (it might not be useful, or might not be useful here!)

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: Should be fixed by https://github.com/apache/arrow-nanoarrow/pull/573

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: Agreed!

paleolimbot (approved):
Thank you!

PR #556, opened 2024-07-12
fix: Minor fix for ArrayViewAllocateChildren

zeroshade:
`ArrowArrayViewAllocateChildren` allocates an array even if asked to allocate for `n_children == 0` and the initialization unconditionally calls `ArrowArrayViewAllocateChildren` with `n_children`. To my knowledge, the `children` member *must* be `NULL` if `n_children` is 0, so I also fixed that in this PR.

### DIFF
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -309,8 +309,8 @@ ArrowErrorCode ArrowArraySetBuffer(struct ArrowArray* array, int64_t i,
   return NANOARROW_OK;
 }
 
-ArrowErrorCode ArrowArrayViewInitFromArray(struct ArrowArrayView* array_view,
-                                           const struct ArrowArray* array) {
+static ArrowErrorCode ArrowArrayViewInitFromArray(struct ArrowArrayView* array_view,
+                                                  struct ArrowArray* array) {
   struct ArrowArrayPrivateData* private_data =
       (struct ArrowArrayPrivateData*)array->private_data;
 
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -309,8 +309,8 @@ ArrowErrorCode ArrowArraySetBuffer(struct ArrowArray* array, int64_t i,
   return NANOARROW_OK;
 }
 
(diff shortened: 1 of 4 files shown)


paleolimbot (commented):
> When working with nanoarrow for a PR with libcudf, I ran into an issue where ArrowArrayViewSetArray expects the ArrayView to already have been initialized with the appropriate type, children, dictionary, etc.

Yes, this is a "feature" of the array view to make it slightly faster to ingest many arrays with the same structure (at the expense of making it slightly more verbose to ingest a single array).

The intention is that the array view should be initialized with `ArrowArrayViewInitFromSchema()`. I am not sure that the "init from array" version should be exposed, because it assumes a very specific `private_data`.

https://github.com/apache/arrow-nanoarrow/blob/2aa2e697fd5d0ef3cd88961bb030f9e760db581b/src/nanoarrow/common/array.c#L314-L315

There are some other functions that assume the array private data (e.g., the builder API), but I am not sure anybody would ever actually want a view of an array that they just built (except for us, and even then that approach causes some problems for things like unions with non-sequential ids).

Is there a reason that `ArrowArrayViewInitFromSchema()` isn't the right tool for the job in your cudf work?

> In addition, `ArrowArrayViewAllocateChildren` allocates an array even if asked to allocate

Thank you!

zeroshade:
> Yes, this is a "feature" of the array view to make it slightly faster to ingest many arrays with the same structure (at the expense of making it slightly more verbose to ingest a single array).

> The intention is that the array view should be initialized with ArrowArrayViewInitFromSchema(). I am not sure that the "init from array" version should be exposed, because it assumes a very specific private_data.

Ah. That makes a lot of sense, fair enough i'll change this PR.

> Is there a reason that ArrowArrayViewInitFromSchema() isn't the right tool for the job in your cudf work?

After poking this for a while, I realized I actually *did* have the schema available and didn't have to build one entirely new to make the test work, so `ArrowArrayViewInitFromSchema` can do the job, even though it's slightly more verbose for some of the other tests. 

I did, however, come across one other minor issue which i'll update in this PR.

zeroshade:
@paleolimbot I've updated the PR description and the files themselves to better describe what is being changed here.

paleolimbot (approved):
Thank you!

PR #559, opened 2024-07-17
refactor(python): Split ArrowArray and ArrowArrayStream modules

paleolimbot:
This is the final splitting of modules from `_lib.pyx`, which no longer exists! It separates the array stream and array components to live in separate .pyx files and adds basic docstrings to internal functions. It also removes a few references to `_lib` and `c_lib` that had made it through to this PR.


danepitkin (approved):
Overall LGTM! Nice work!

PR #560, opened 2024-07-17
chore(ci): Use `docker compose` instead of `docker-compose` to build images

paleolimbot:


eitsupi:
Do you want to install docker compose on the runners?
Or, since it looks like the runners already have buildx installed, I believe you can use the `docker buildx bake` command instead of `docker compose`.

PR #561, opened 2024-07-17
refactor: Separate implementation from interface for nanoarrow_testing component

paleolimbot:
This PR implements `nanoarrow_testing` as a target instead of a (massive) header-only C++ implementation. This was suggested multiple times by multiple reviewers and makes it possible to add more testing utilities without further bloating a massive implementation/source mishmash. It also makes it more clear how to actually use the testing utilities because the header file is easier to skim (and doxygen no longer chokes on the header, which had previously resulted in these utilities not making it into the documentation builld!). Another advantage of this is removing `nlohmann/json.hpp` from the public header (i.e., a build-time only dependency).

The general pattern of `testing/testint.cc` is:

```cpp
namespace writer_internal {
namespace {
// Implementations of functions that don't need the state of the TestingJsonWriter
}
}

// TestingJsonWriter implementations

namespace writer_internal {
namespace {
// Implementations of functions that don't need the state of the TestingJsonReader
}
}

// TestingJsonReader implementations
```


bkietz (approved):
LGTM; straightforward extraction of a testing module.

PR #562, opened 2024-07-19
add ArrowArrayViewComputeNullCount

bkietz:
Adds `ArrowArrayViewComputeNullCount()` and tests. Extracted from https://github.com/apache/arrow-nanoarrow/pull/555

### DIFF
--- a/r/src/materialize_unspecified.h
+++ b/r/src/materialize_unspecified.h
@@ -35,19 +35,18 @@ static inline int nanoarrow_materialize_unspecified(struct ArrayViewSlice* src,
 
   if (ArrowArrayViewComputeNullCount(src->array_view) == 0) {
     // We can blindly set all the values to NA_LOGICAL without checking
-    for (int64_t i = 0; i < length; i++) {
+    for (int64_t i = 0; i < src->length; i++) {
       result[dst->offset + i] = NA_LOGICAL;
     }
     return NANOARROW_OK;
   }
 
   int64_t total_offset = src->array_view->array->offset + src->offset;
-  int64_t length = src->length;
   const uint8_t* bits = src->array_view->buffer_views[0].data.as_uint8;
 
   // Count non-null values and warn
   int64_t n_bad_values = 0;
-  for (int64_t i = 0; i < length; i++) {
+  for (int64_t i = 0; i < src->length; i++) {
     n_bad_values += ArrowBitGet(bits, total_offset + i);
     result[dst->offset + i] = NA_LOGICAL;
   }
--- a/r/src/materialize_unspecified.h
+++ b/r/src/materialize_unspecified.h
@@ -33,7 +33,7 @@ static inline int nanoarrow_materialize_unspecified(struct ArrayViewSlice* src,
 
   int* result = LOGICAL(dst->vec_sexp);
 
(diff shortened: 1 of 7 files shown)


thirdparty/zlib/CMakeLists.txt
```diff
@@ -23,3 +23,4 @@ fetchcontent_makeavailable(nanoarrow_zlib)
 
 add_library(ZLIB::ZLIB ALIAS zlibstatic)
 target_include_directories(zlibstatic INTERFACE ${zlib_BINARY_DIR} ${zlib_SOURCE_DIR})
+target_include_directories(zlib INTERFACE ${zlib_BINARY_DIR} ${zlib_SOURCE_DIR})
```
paleolimbot, line 26: I don't mind this, but is it needed to build?

r/src/materialize_unspecified.h
```diff
@@ -33,27 +33,26 @@ static inline int nanoarrow_materialize_unspecified(struct ArrayViewSlice* src,
 
   int* result = LOGICAL(dst->vec_sexp);
 
-  int64_t total_offset = src->array_view->array->offset + src->offset;
-  int64_t length = src->length;
-  const uint8_t* bits = src->array_view->buffer_views[0].data.as_uint8;
-
-  if (length == 0 || src->array_view->storage_type == NANOARROW_TYPE_NA ||
-      ArrowBitCountSet(bits, total_offset, length) == 0) {
+  if (ArrowArrayViewComputeNullCount(src->array_view) == src->length) {
```
paleolimbot, line 36: I'm not sure this can drop in here because we could be only materializing a portion of an array (i.e., there is an additional offset/length on top of the `ArrowArrayView`).

That said, I'm pretty sure the code that I have here will crash if `bits` is NULL here. The unspecified type doesn't come up very often and I don't think I tested this particularly well when I wrote it.

paleolimbot (commented):
Thanks! The C and Python changes look great!

r/src/materialize_unspecified.h
bkietz: I'll remove this

thirdparty/zlib/CMakeLists.txt
bkietz: I needed this when building against conda. Without it, building the bundled zlib raises

```
/home/ben/nanoarrow/build/_deps/nanoarrow_zlib-src/test/example.c:8:10: fatal error: zlib.h: No such file or directory
    8 | #include "zlib.h"
      |          ^~~~~~~~
```

Ideally there would be an option we could pass to fetchcontent in order to disable building tests, but it seems unconditional. Since the test links to zlib (and not to zlibstatic), the include directories attached to zlibstatic don't apply. I don't know why this isn't failing more places; presumably there's just another zlib.h that they can always pick up.

From the fetchcontent'd CMakeLists.txt for zlib:

```cmake

#============================================================================
# Example binaries
#============================================================================

add_executable(example test/example.c)
target_link_libraries(example zlib)
add_test(example example)

```

paleolimbot: Got it!

paleolimbot (approved):
Thank you!

PR #564, opened 2024-07-23
feat: Add IPC writer scaffolding

bkietz:
Add `ArrowIpcEncoder`, init/reset, and tests. Extracted from https://github.com/apache/arrow-nanoarrow/pull/555#pullrequestreview-2186295630

### DIFF
--- a/src/nanoarrow/ipc/encoder.c
+++ b/src/nanoarrow/ipc/encoder.c
@@ -59,11 +59,12 @@
 
 struct ArrowIpcEncoderPrivate {
   flatcc_builder_t builder;
-  struct ArrowBuffer buffers, nodes;
+  struct ArrowBuffer buffers;
+  struct ArrowBuffer nodes;
 };
 
 ArrowErrorCode ArrowIpcEncoderInit(struct ArrowIpcEncoder* encoder) {
-  NANOARROW_DCHECK(encoder);
+  NANOARROW_DCHECK(encoder != NULL);
   memset(encoder, 0, sizeof(struct ArrowIpcEncoder));
   encoder->encode_buffer = NULL;
   encoder->encode_buffer_state = NULL;
@@ -71,16 +72,17 @@ ArrowErrorCode ArrowIpcEncoderInit(struct ArrowIpcEncoder* encoder) {
   encoder->private_data = ArrowMalloc(sizeof(struct ArrowIpcEncoderPrivate));
   struct ArrowIpcEncoderPrivate* private =
       (struct ArrowIpcEncoderPrivate*)encoder->private_data;
-  ArrowBufferInit(&private->buffers);
-  ArrowBufferInit(&private->nodes);
   if (flatcc_builder_init(&private->builder) == -1) {
+    ArrowFree(private);
     return ESPIPE;
   }
+  ArrowBufferInit(&private->buffers);
+  ArrowBufferInit(&private->nodes);
   return NANOARROW_OK;
 }
 
 void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
-  NANOARROW_DCHECK(encoder && encoder->private_data);
+  NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL);
   struct ArrowIpcEncoderPrivate* private =
       (struct ArrowIpcEncoderPrivate*)encoder->private_data;
   flatcc_builder_clear(&private->builder);
@@ -92,17 +94,24 @@ void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
 
 ArrowErrorCode ArrowIpcEncoderFinalizeBuffer(struct ArrowIpcEncoder* encoder,
                                              struct ArrowBuffer* out) {
-  NANOARROW_DCHECK(encoder && encoder->private_data && out);
+  NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL && out != NULL);
   struct ArrowIpcEncoderPrivate* private =
       (struct ArrowIpcEncoderPrivate*)encoder->private_data;
-  ArrowBufferReset(out);
-  size_t size = flatcc_builder_get_buffer_size(&private->builder);
+
+  int64_t size = (int64_t)flatcc_builder_get_buffer_size(&private->builder);
   if (size == 0) {
     // Finalizing an empty flatcc_builder_t triggers an assertion
-    return NANOARROW_OK;
+    return ArrowBufferResize(out, 0, 0);
   }
 
-  out->size_bytes = out->capacity_bytes = (int64_t)size;
-  out->data = (uint8_t*)flatcc_builder_finalize_buffer(&private->builder, &size);
-  return out->data ? NANOARROW_OK : ENOMEM;
+  void* data = flatcc_builder_get_direct_buffer(&private->builder, NULL);
+  if (data == NULL) {
+    return ENOMEM;
+  }
+  NANOARROW_RETURN_NOT_OK(ArrowBufferResize(out, size, 0));
+  memcpy(out->data, data, size);
+
+  // don't deallocate yet, just wipe the builder's current Message
+  flatcc_builder_reset(&private->builder);
+  return NANOARROW_OK;
 }
--- a/src/nanoarrow/ipc/encoder.c
+++ b/src/nanoarrow/ipc/encoder.c
@@ -1,0 +1,108 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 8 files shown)


src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,108 @@
...
+  flatcc_builder_clear(&private->builder);
+  ArrowBufferReset(&private->nodes);
+  ArrowBufferReset(&private->buffers);
+  ArrowFree(private);
+  memset(encoder, 0, sizeof(struct ArrowIpcEncoder));
+}
+
+ArrowErrorCode ArrowIpcEncoderFinalizeBuffer(struct ArrowIpcEncoder* encoder,
+                                             struct ArrowBuffer* out) {
+  NANOARROW_DCHECK(encoder && encoder->private_data && out);
```
paleolimbot, line 95:
```suggestion
  NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL && out != NULL);
```

src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -379,6 +395,44 @@ ArrowErrorCode ArrowIpcArrayStreamReaderInit(
...
+                                  int64_t* length, struct ArrowError* error);
+
+  /// \brief Pointer to arbitrary data used by encode_buffer()
+  void* encode_buffer_state;
+
+  /// \brief Private resources managed by this library
+  void* private_data;
+};
+
+/// \brief Initialize an encoder
```
paleolimbot, line 427:
```suggestion
/// \brief Initialize an encoder
///
/// If NANOARROW_OK is returned, the caller must call `ArrowIpcEncoderReset()`
/// to release resources allocated by this function.
```

paleolimbot (commented):
Thank you!

The CI failure is I think because of the missing line here:

https://github.com/apache/arrow-nanoarrow/blob/d6368d0672470304f0a01cc365133c8d19d1ea4c/ci/scripts/bundle.py#L205-L206

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,108 @@
...
+  encoder->encode_buffer = NULL;
+  encoder->encode_buffer_state = NULL;
+  encoder->codec = NANOARROW_IPC_COMPRESSION_TYPE_NONE;
+  encoder->private_data = ArrowMalloc(sizeof(struct ArrowIpcEncoderPrivate));
+  struct ArrowIpcEncoderPrivate* private =
+      (struct ArrowIpcEncoderPrivate*)encoder->private_data;
+  ArrowBufferInit(&private->buffers);
+  ArrowBufferInit(&private->nodes);
+  if (flatcc_builder_init(&private->builder) == -1) {
+    return ESPIPE;
```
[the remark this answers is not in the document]
bkietz: Aaargh C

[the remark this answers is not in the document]
bkietz: (still adding them, but:) is buffer reset necessary? Buffer init produces empty buffers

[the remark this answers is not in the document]
paleolimbot: Ah that's a great point. Maybe put a comment there (or move the buffer inits after constructing the builder)?

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,108 @@
...
+  ArrowBufferReset(out);
+  size_t size = flatcc_builder_get_buffer_size(&private->builder);
+  if (size == 0) {
+    // Finalizing an empty flatcc_builder_t triggers an assertion
+    return NANOARROW_OK;
+  }
+
+  out->size_bytes = out->capacity_bytes = (int64_t)size;
+  out->data = (uint8_t*)flatcc_builder_finalize_buffer(&private->builder, &size);
+  return out->data ? NANOARROW_OK : ENOMEM;
```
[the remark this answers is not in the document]
bkietz: This depends. In order of increasing complexity:

- If it's acceptable to always copy from the buffer where flatcc has been operating
  in-place to a fresh output buffer then that's certainly simplest. In that case we
  resize an initialized `out` then memcpy.
- We could keep using the flatcc default allocator, but overwrite the output
  buffer's allocator to use `flatcc_builder_{free,alloc}`. (In this case, `out` is
  initialized when finalizing the encoder so the argument shouldn't be initialized)
  with the flatcc allocator and builder's
  result. This would save us a copy (unless we need to copy the result into a buffer
  from some other pool).
- We could eventually implement custom allocators for flatcc, which allows us to
  control what memory flatcc is using to build *and* saves us a finalizing copy
  no matter where we want the output. In this case again we require an externally
  initialized `out` whose allocator we reuse for flatcc building.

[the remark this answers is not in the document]
bkietz: I'll just copy for now

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,117 @@
...
+ArrowErrorCode ArrowIpcEncoderFinalizeBuffer(struct ArrowIpcEncoder* encoder,
+                                             struct ArrowBuffer* out) {
+  NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL && out != NULL);
+  struct ArrowIpcEncoderPrivate* private =
+      (struct ArrowIpcEncoderPrivate*)encoder->private_data;
+
+  int64_t size = (int64_t)flatcc_builder_get_buffer_size(&private->builder);
+  if (size == 0) {
+    // Finalizing an empty flatcc_builder_t triggers an assertion
+    return ArrowBufferResize(out, 0, 0);
```
paleolimbot, line 104:
```suggestion
    return NANOARROW_OK;
```

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  void* data = flatcc_builder_get_direct_buffer(&private->builder, NULL);
+  if (data == NULL) {
+    return ENOMEM;
+  }
+  NANOARROW_RETURN_NOT_OK(ArrowBufferResize(out, size, 0));
+  memcpy(out->data, data, size);
```
paleolimbot, line 112: I think this is the best of both worlds:

```suggestion
  NANOARROW_RETURN_NOT_OK(ArrowBufferReserve(out, size));
  if (flatcc_builder_copy_buffer(&private->builder, out->data + out->size_bytes, out->capacity_bytes - out_size_bytes) != NULL) {
    return EIO;
  }
```

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,117 @@
...
+
+#if !defined(__STDC_NO_ATOMICS__)
+#include <stdatomic.h>
+#undef NANOARROW_IPC_USE_STDATOMIC
+#define NANOARROW_IPC_USE_STDATOMIC 1
+#endif
+#endif
+#endif
+
+#endif
```
paleolimbot, line 44:
```suggestion
```

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,117 @@
...
+
+#include "nanoarrow/ipc/flatcc_generated.h"
+#include "nanoarrow/nanoarrow.h"
+#include "nanoarrow/nanoarrow_ipc.h"
+
+// R 3.6 / Windows builds on a very old toolchain that does not define ENODATA
+#if defined(_WIN32) && !defined(_MSC_VER) && !defined(ENODATA)
+#define ENODATA 120
+#endif
```
paleolimbot, line 53:
```suggestion
```

paleolimbot (commented):
Sorry for missing a few of these in the initial round!

I think we want to append to the output buffer rather than resize it (more flexible for the caller, who can choose to resize it to zero before they call or not depending what they're about to do/what they just did)!

src/nanoarrow/ipc/encoder.c
bkietz: flatcc_builder_copy_buffer returns NULL for me. I've rewritten this using ArrowBufferAppend, which works

src/nanoarrow/ipc/encoder.c
```diff
@@ -0,0 +1,84 @@
...
+  struct ArrowIpcEncoderPrivate* private =
+      (struct ArrowIpcEncoderPrivate*)encoder->private_data;
+
+  int64_t size = (int64_t)flatcc_builder_get_buffer_size(&private->builder);
+  if (size == 0) {
+    // Finalizing an empty flatcc_builder_t triggers an assertion
+    return NANOARROW_OK;
+  }
+
+  void* data = flatcc_builder_get_direct_buffer(&private->builder, NULL);
```
paleolimbot, line 74: My reading of https://github.com/dvidelabs/flatcc/blob/fe24a8012d689d4f7ca9e1ccdfeb012c74ca96bb/include/flatcc/flatcc_builder.h#L1802-L1805 is that `flatcc_builder_get_direct_buffer()` it might fail for large output. It's a bummer the copy function doesn't work (I was trying to replicate https://github.com/dvidelabs/flatcc/blob/fe24a8012d689d4f7ca9e1ccdfeb012c74ca96bb/src/runtime/builder.c#L1972-L1988 but using our own allocation).

Feel free to test/solve later!

paleolimbot (approved):
Thank you!

src/nanoarrow/ipc/encoder.c
bkietz: The returns-null case is checked and we return ENOMEM if we can't allocate the direct buffer. We can try again with copy_buffer later

paleolimbot: I don't think it's an allocation failure, I think the builder uses multiple pages of non-contiguous memory for large output. +1 to debugging later!

PR #568, opened 2024-07-26
feat: Add IPC schema encoding

bkietz:
- added ArrowIpcEncoderEncodeSchema
- added a parameter to ArrowIpcEncoderFinalizeBuffer which controls whether encapsulated/padded message buffers will be produced instead of raw
- tests reuse the decoder tests, replacing arrow C++'s encoder with ArrowIpcEncoder

Extracted from https://github.com/apache/arrow-nanoarrow/pull/555#pullrequestreview-2186295630

### DIFF
--- a/src/nanoarrow/ipc/decoder_test.cc
+++ b/src/nanoarrow/ipc/decoder_test.cc
@@ -469,6 +469,99 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcArrowTypeRoundtrip) {
   ArrowIpcDecoderReset(&decoder);
 }
 
+std::string ArrowSchemaMetadataToString(const char* metadata) {
+  struct ArrowMetadataReader reader;
+  struct ArrowStringView key;
+  struct ArrowStringView value;
+  NANOARROW_DCHECK(ArrowMetadataReaderInit(&reader, metadata) == NANOARROW_OK);
+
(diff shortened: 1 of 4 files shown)


src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -749,6 +749,53 @@ TEST_P(ArrowSchemaParameterizedTestFixture, NanoarrowIpcArrowSchemaRoundtrip) {
...
 
+std::string ArrowSchemaToString(const struct ArrowSchema* schema) {
+  std::string out;
+  size_t n = 1024;
+  while (out.size() < n) {
+    out.resize(n);
+    n = ArrowSchemaToString(schema, out.data(), out.size(), /*recursive=*/true);
+  }
+  return out;
+}
```
paleolimbot, line 760: This is nanoarrow's fault for not having a proper schema comparison; however, this is technically not comparing metadata or flags. We could tack those on to this output string or punt to a better schema comparison (your call!).

paleolimbot (approved):
Thank you!

src/nanoarrow/ipc/decoder_test.cc
bkietz: For now, I'll just add the metadata to the string.

paleolimbot (approved):

PR #569, opened 2024-07-26
feat: Add schema construction DSL

bkietz:
Construct `nanoarrow::UniqueSchema`s with a tiny C++ DSL:

```c++
  schema{children{
      {"i", "i32",
       metadata{
           "some_key=some_value",
       }},
      {"i", "i32",
       dictionary{{"u"}},
       metadata{
           "some_key=some_value",
       },
       ARROW_FLAG_NULLABLE},
  }};
```

The python-equivalent signature is approximately:
```python
def schema(format="+s", name="", metadata=[], flags=0, children=[], dictionary=None):
```

After the first two arguments, all the rest can be provided in any order.

### DIFF
--- a/src/nanoarrow/testing/testing_test.cc
+++ b/src/nanoarrow/testing/testing_test.cc
@@ -1908,3 +1908,21 @@ TEST(NanoarrowTestingTest, NanoarrowTestingTestArrayStreamComparison) {
 
 )");
 }
+
+TEST(SchemaDsl, Basic) {
+  using namespace nanoarrow::testing::dsl;
+
+  schema{children{
+      {"i", "i32",
+       metadata{
+           "some_key=some_value",
+       }},
+      {"i", "i32",
+       dictionary{{"u"}},
+       metadata{
+           "some_key=some_value",
+       },
+       ARROW_FLAG_NULLABLE},
+  }};
+}
+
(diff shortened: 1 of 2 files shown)


src/nanoarrow/testing/testing_test.cc
```diff
@@ -1908,3 +1908,19 @@ TEST(NanoarrowTestingTest, NanoarrowTestingTestArrayStreamComparison) {
...
+      {"i", "int32 field name",
+       metadata{
+           "some_key=some_value",
+       }},
+      {"i", dictionary{"u"}, "dictionary field name",
+       metadata{
+           "some_key=some_value",
+       },
+       ARROW_FLAG_NULLABLE},
+  }};
```
paleolimbot, line 1925: In tests I think I would still prefer to see the C calls (but perhaps this is not where you were headed with this)

paleolimbot (commented):
Just some initial thoughts! If we're really headed for readable schema construction, Arrow C++-like constructors (e.g., `int32("field name")`) are more likely to be what we want (and might be substantially less complicated to implement). I probably would have gone towards templating on `ArrowType` and using our C library constructors instead of introducing another way to create schemas.

It's worth looking at how https://github.com/man-group/sparrow does this (I think they have something pretty similar to construct types/schemas).

I was hoping to merge https://github.com/apache/arrow-nanoarrow/pull/561 before adding anything to the testing library (it could use a review!).

PR #570, opened 2024-07-30
feat: add ArrowIpcOutputStream

bkietz:
This is a prerequisite for adding nanoarrow to the IPC integration tests (since we'll need a full stream writer into which we'll push batches, which will need an output stream into which it'll push bytes).

### DIFF
--- a/src/nanoarrow/ipc/encoder.c
+++ b/src/nanoarrow/ipc/encoder.c
@@ -62,7 +62,7 @@ ArrowErrorCode ArrowIpcEncoderInit(struct ArrowIpcEncoder* encoder) {
 }
 
 void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
-  NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL);
+  NANOARROW_DCHECK(encoder != NULL);
   struct ArrowIpcEncoderPrivate* private =
       (struct ArrowIpcEncoderPrivate*)encoder->private_data;
   if (private != NULL) {
--- a/src/nanoarrow/ipc/encoder.c
+++ b/src/nanoarrow/ipc/encoder.c
@@ -65,11 +65,13 @@ void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
   NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL);
   struct ArrowIpcEncoderPrivate* private =
       (struct ArrowIpcEncoderPrivate*)encoder->private_data;
-  flatcc_builder_clear(&private->builder);
-  ArrowBufferReset(&private->nodes);
-  ArrowBufferReset(&private->buffers);
-  ArrowFree(private);
-  memset(encoder, 0, sizeof(struct ArrowIpcEncoder));
+  if (private != NULL) {
+    flatcc_builder_clear(&private->builder);
+    ArrowBufferReset(&private->nodes);
+    ArrowBufferReset(&private->buffers);
+    ArrowFree(private);
+    memset(encoder, 0, sizeof(struct ArrowIpcEncoder));
+  }
 }
 
 ArrowErrorCode ArrowIpcEncoderFinalizeBuffer(struct ArrowIpcEncoder* encoder,
--- a/src/nanoarrow/ipc/writer.c
+++ b/src/nanoarrow/ipc/writer.c
@@ -38,10 +38,10 @@ static ArrowErrorCode ArrowIpcOutputStreamBufferWrite(struct ArrowIpcOutputStrea
                                                       int64_t buf_size_bytes,
                                                       int64_t* size_written_out,
                                                       struct ArrowError* error) {
-  NANOARROW_UNUSED(error);
   struct ArrowIpcOutputStreamBufferPrivate* private_data =
       (struct ArrowIpcOutputStreamBufferPrivate*)stream->private_data;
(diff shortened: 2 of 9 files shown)


bkietz:
This PR is based off https://github.com/apache/arrow-nanoarrow/pull/568 so please don't merge until that one is merged

src/nanoarrow/ipc/encoder.c
```diff
@@ -52,11 +52,13 @@ void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
   NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL);
   struct ArrowIpcEncoderPrivate* private =
       (struct ArrowIpcEncoderPrivate*)encoder->private_data;
-  flatcc_builder_clear(&private->builder);
-  ArrowBufferReset(&private->nodes);
-  ArrowBufferReset(&private->buffers);
-  ArrowFree(private);
-  memset(encoder, 0, sizeof(struct ArrowIpcEncoder));
+  if (private != NULL) {
```
paleolimbot, line 68: This has already been checked as a debug check?

src/nanoarrow/ipc/writer.c
```diff
@@ -0,0 +1,152 @@
...
+#include <string.h>
+
+#include "nanoarrow/nanoarrow.h"
+#include "nanoarrow/nanoarrow_ipc.h"
+
+void ArrowIpcOutputStreamMove(struct ArrowIpcOutputStream* src,
+                              struct ArrowIpcOutputStream* dst) {
+  memcpy(dst, src, sizeof(struct ArrowIpcOutputStream));
+  src->release = NULL;
+}
```
paleolimbot, line 30: Should this be `static inline`?

src/nanoarrow/ipc/writer.c
```diff
@@ -0,0 +1,152 @@
...
+struct ArrowIpcOutputStreamBufferPrivate {
+  struct ArrowBuffer* output;
+};
+
+static ArrowErrorCode ArrowIpcOutputStreamBufferWrite(struct ArrowIpcOutputStream* stream,
+                                                      const void* buf,
+                                                      int64_t buf_size_bytes,
+                                                      int64_t* size_written_out,
+                                                      struct ArrowError* error) {
+  NANOARROW_UNUSED(error);
```
paleolimbot, line 41: It is a little verbose, but you can use `NANOARROW_RETURN_NOT_OK_WITH_ERROR` (I think that's what it's called) here. In general, we try to guarantee that `error` is always populated if anything other than `NANOARROW_OK` is returned.

src/nanoarrow/ipc/writer.c
```diff
@@ -0,0 +1,152 @@
...
+                                                    const void* buf,
+                                                    int64_t buf_size_bytes,
+                                                    int64_t* size_written_out,
+                                                    struct ArrowError* error) {
+  struct ArrowIpcOutputStreamFilePrivate* private_data =
+      (struct ArrowIpcOutputStreamFilePrivate*)stream->private_data;
+
+  if (private_data->stream_finished) {
+    *size_written_out = 0;
+    return NANOARROW_OK;
```
paleolimbot, line 101: Should this be an error? Would this be like writing to a closed file?

src/nanoarrow/ipc/writer.c
```diff
@@ -0,0 +1,152 @@
...
+
+    // Try to close the file now
+    if (private_data->close_on_release) {
+      if (fclose(private_data->file_ptr) == 0) {
+        private_data->file_ptr = NULL;
+      }
+    }
+
+    // Maybe return error
+    if (has_error) {
```
paleolimbot, line 122: Should this return an error if there was an error closing the file?

paleolimbot (approved):
Thank you!

src/nanoarrow/ipc/writer.c
bkietz: I could make it `static inline` if you prefer; it is quite simple. I exported it only because `ArrowIpcInputStreamMove` is also exported; should I inline them both?

src/nanoarrow/ipc/encoder.c
bkietz: I'll remove the debug check. IIUC we should always be allowed to call Reset (this is how decoder's Reset functions)

src/nanoarrow/ipc/writer.c
bkietz: IIUC the connotation of `close_on_release` is that close errors are/should be ignored (this is the input stream's behavior as well). From the doccomment, "If this behaviour is needed, pass false to close_on_release and handle closing the file independently from stream."

src/nanoarrow/ipc/writer.c
bkietz: That makes more sense to me, but again I was going for consistency with InputStream where reading from a closed stream always yields zero bytes and no error.

bkietz:
I think if any of these needs to be addressed we can/should do so in a follow up.

src/nanoarrow/ipc/encoder.c
paleolimbot: +1 for consistency with the decoder (I think this is not consistent across nanoarrow and should be fixed at some point!)

src/nanoarrow/ipc/writer.c
paleolimbot: Consistency seems like the right move!

PR #572, opened 2024-08-05
fix(ci): Pin r-lib actions as a workaround for latest action updates

paleolimbot:
First noticed at https://github.com/apache/arrow-nanoarrow/pull/555#discussion_r1702201173 , the R check action is failing because an update to r-lib actions resulted in some quarto actions being invoked, and these have not yet been whitelisted for use in Apache repositories. It also may be that we don't need the quarto actions (we probably don't) but some brief experimentation to attempt circumventing the use of the quarto action did not result in a successful workflow. Hence, a pin to unblock PR checks until either the v2 branch is updated or it is clear how to avoid the failure.

### DIFF
--- a/.github/workflows/r-check.yaml
+++ b/.github/workflows/r-check.yaml
@@ -52,11 +52,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-
-      # r-lib/actions pinned because the latest action on the v2 branch requires
-      # quarto-dev actions which have not yet beeen whitelisted by Apache Infra
-      - uses: r-lib/actions/setup-pandoc@f4937e0dc26f9b99c969cd3e4ca943b576e7f991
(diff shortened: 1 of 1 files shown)


assignUser (commented):
I speculate (without checking :D) that the issue is that even if the action skips the quarto related action the gh runner will still attempt to fetch all actions in `- uses:`  tags. 

I'll see if there is either a way to go around that or get the action allowlisted.

jiayuasu:
@paleolimbot We had the same issue at Sedona R: https://github.com/apache/sedona/pull/1537

If this is a very important action, we can create an ASF infra ticket on JIRA and ask ASF together to whitelist it. We did it once before.

assignUser:
@jiayuasu there is an open infa ticket to allow the quarto action: https://issues.apache.org/jira/projects/INFRA/issues/INFRA-26031

assignUser:
This issue is fixed now!

PR #573, opened 2024-08-05
refactor: Consolidate per-target actions in CMakeLists.txt

paleolimbot:
In previous PRs we consolidated the extensions into the main CMakeLists.txt; however, there were some things happening for certain targets (like installing them or setting NANOARROW_DEBUG) but not others.

Noticed in https://github.com/apache/arrow-nanoarrow/pull/555#discussion_r1702103767 where there was a DCHECK referencing a variable that didn't exist that made it through CI 😬

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -356,31 +356,13 @@ foreach(target
             DESTINATION lib
             EXPORT nanoarrow-exports)
 
-    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-      if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
-        target_compile_options(${target}
-                               PRIVATE -Wall
-                                       -Werror
-                                       -Wextra
-                                       -Wpedantic
-                                       -Wno-type-limits
-                                       -Wmaybe-uninitialized
-                                       -Wunused-result
-                                       -Wconversion
-                                       -Wno-sign-conversion)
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -346,9 +339,50 @@ if(NANOARROW_BUILD_TESTS OR NANOARROW_BUILD_INTEGRATION_TESTS)
...
+        nanoarrow_testing
+        nanoarrow_c_data_integration)
+  if(TARGET ${target})
+    target_compile_definitions(${target} PUBLIC "$<$<CONFIG:Debug>:NANOARROW_DEBUG>")
+
+    install(TARGETS ${target}
+            DESTINATION lib
+            EXPORT nanoarrow-exports)
+
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
```
bkietz, line 359: It's possibly not a concern for nanoarrow, but CMake doesn't define this variable in all generators. Specifically it's defined for single configuration generators like Ninja or makefiles, but on multi configuration generators like Visual Studio or XCode we have instead https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html

I think the most correct way to do this is by wrapping these options in a generator expression:
```
target_compile_options(${target} PRIVATE "$<$<CONFIG:Debug>:-Wall -Werror -Wextra ...>")
```

bkietz (approved):
LGTM, just a few comments for potential follow up

PR #574, opened 2024-08-05
fix: Fix Meson build for separated nanoarrow_testing target

paleolimbot:
After https://github.com/apache/arrow-nanoarrow/pull/561 , the nanoarrow_testing library is no longer header-only and requires linking!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -29,9 +29,11 @@ project(
     ]
 )
 
-if meson.get_compiler('cpp').get_id() == 'gcc' or meson.get_compiler('cpp').get_id() == 'clang'
-  add_project_arguments(['-fvisibility=hidden'], language: 'cpp')
-endif
+# This bit depends on proper symbol exporting which hasn't yet been implemented
+# generally for all of nanoarrow's libraries
+# if meson.get_compiler('cpp').get_id() == 'gcc' or meson.get_compiler('cpp').get_id() == 'clang'
+#   add_project_arguments(['-fvisibility=hidden'], language: 'cpp')
+# endif
 
 nanoarrow_dep_args = []
 if host_machine.system() == 'windows' and get_option('default_library') == 'shared'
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -148,50 +164,28 @@ if get_option('tests')
...
+  nanoarrow_tests = ['utils', 'buffer', 'array', 'schema', 'array-stream', 'nanoarrow-hpp']
+
+  foreach name : nanoarrow_tests
     exc = executable(
         name + '-test',
         sources: 'src/nanoarrow/common/' + name.replace('-', '_') + '_test.cc',
-        link_with: nanoarrow_lib,
         include_directories: incdir,
-        dependencies: config['deps'],
+        dependencies: [nanoarrow_testing_dep, nanoarrow_dep, arrow_dep, gtest_dep, gmock_dep],
```
WillAyd, line 175: Not sure if it matters too much, but `nanoarrow_dep` is duplicative here, since `nanoarrow_testing_dep` already includes that in its `declare_dependency` declaration

meson.options
```diff
@@ -25,6 +25,7 @@ option('integration_tests', type: 'boolean',
 option('namespace', type: 'string',
        description: 'A prefix for exported symbols')
 option('device', type: 'boolean', description: 'Build device libraries', value: false)
+option('testing', type: 'boolean', description: 'Build testing libraries', value: false)
```
WillAyd, line 28: Is there any advantage to adding this as a new build option versus just having it tied to the existing `tests` option?

WillAyd (commented):
Looks great - just a few comments / questions

meson.build
paleolimbot: Does it matter that `nanoarrow_testing_dep` doesn't list `link_with: nanoarrow_lib`? Or maybe that would only matter with hidden visiblity? There were some places that listed `nanoarrow_ipc_dep` and `nanoarrow_dep`, which I tried to replicate here, but I really have no idea 😬

meson.options
paleolimbot: I think so...one could use the testing library from another project, too (in which case you probably wouldn't want to build nanoarrow's tests too).

meson.build
WillAyd: Don't think so - I believe the `nanoarrow_dep` it has as a dependency will take care of that. Using CMake as an analogy, I think:

```python
nanoarrow_dep = declare_dependency(link_with: nanoarrow_lib, ...)
```

is akin to using the PUBLIC keyword when setting the link libraries of a target

```cmake
target_link_libraries(nanoarrow PUBLIC nanoarrow_lib)
```

i.e. it becomes transitive for any target that references the dependency

WillAyd: (in the CMake example, `nanoarrow` would probably be an OBJECT library)

paleolimbot: I'll update!

WillAyd (approved):
nice work. you are becoming a meson expert

PR #571, opened 2024-08-01
feat: Add IPC stream writing

bkietz:
- adds ArrowIpcArrayStreamWriter along with Init() and Reset() methods
  - embeds an Encoder and an OutputStream
- adds WriteSchema() and WriteArrayView() methods which encode then write a schema or array
- adds WriteArrayStream() method which serializes an entire ArrowArrayStream, finishing with an explicit EOS
  - non-blocking IO is not currently supported

PR #582, opened 2024-08-12
fix: IPC streams did not include RecordBatch headers

bkietz:
- As noted https://github.com/apache/arrow-nanoarrow/pull/571#discussion_r1712730976 RecordBatch Messages were not being written during IPC streaming. This has been corrected
- Testing uses one more round trip case where we have arrow C++ read from a stream written by nanoarrow.
- During testing it also became apparent that encapsulated messages were not always stored with the correct size; metadata_size should include padding (but not the continuation or size itself) https://arrow.apache.org/docs/format/Columnar.html#encapsulated-message-format

### DIFF
--- a/src/nanoarrow/ipc/encoder.c
+++ b/src/nanoarrow/ipc/encoder.c
@@ -78,6 +78,18 @@ void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
   memset(encoder, 0, sizeof(struct ArrowIpcEncoder));
 }
 
+static ArrowErrorCode ArrowIpcEncoderWriteContinuationAndSize(struct ArrowBuffer* out,
+                                                              size_t size) {
+  _NANOARROW_CHECK_UPPER_LIMIT(size, INT32_MAX);
+  NANOARROW_RETURN_NOT_OK(ArrowBufferAppendInt32(out, -1));
+
+  if (ArrowIpcSystemEndianness() == NANOARROW_IPC_ENDIANNESS_BIG) {
+    return ArrowBufferAppendInt32(out, (int32_t)bswap32((uint32_t)size));
+  } else {
+    return ArrowBufferAppendInt32(out, (int32_t)size);
+  }
+}
+
 ArrowErrorCode ArrowIpcEncoderFinalizeBuffer(struct ArrowIpcEncoder* encoder,
                                              char encapsulate, struct ArrowBuffer* out) {
   NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL && out != NULL);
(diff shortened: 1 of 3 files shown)


src/nanoarrow/ipc/files_test.cc
```diff
@@ -250,20 +250,33 @@ class TestFile {
...
-    // Read the roundtripped buffer using nanoarrow
     nanoarrow::UniqueSchema roundtripped_schema;
     std::vector<nanoarrow::UniqueArray> roundtripped_arrays;
-    ASSERT_EQ(ReadArrowArrayStreamIPC(dir_prefix, roundtripped_schema.get(),
-                                      &roundtripped_arrays, &error),
-              NANOARROW_OK);
+    {
+      SCOPED_TRACE("Read the roundtripped buffer using nanoarrow");
+      ASSERT_EQ(ReadArrowArrayStreamIPC(dir_prefix, roundtripped_schema.get(),
+                                        &roundtripped_arrays, &error),
```
paleolimbot, line 265: Should this line refer to the buffer `roundtripped`? (i.e., maybe the previous test was just reading from the file and not the roundtripped data?)

src/nanoarrow/ipc/encoder.c
```diff
@@ -78,30 +78,33 @@ void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
   memset(encoder, 0, sizeof(struct ArrowIpcEncoder));
 }
 
+static ArrowErrorCode ArrowIpcEncoderWriteContinuationAndSize(struct ArrowBuffer* out,
+                                                              size_t size) {
+  _NANOARROW_CHECK_UPPER_LIMIT(size, INT32_MAX);
+  int32_t header[] = {-1, ArrowIpcSystemEndianness() == NANOARROW_IPC_ENDIANNESS_BIG
+                              ? (int32_t)bswap32((uint32_t)size)
+                              : (int32_t)size};
```
paleolimbot, line 86: nit: If you can avoid the ternary operator, I find

```c
if (ArrowIpcSystemEndianness() == NANOARROW_IPC_ENDIANNESS_BIG) {
  // ...
} else {
  // ...
}
```

much easier to parse

src/nanoarrow/ipc/encoder.c
```diff
@@ -78,30 +78,33 @@ void ArrowIpcEncoderReset(struct ArrowIpcEncoder* encoder) {
...
+  return ArrowBufferAppend(out, &header, sizeof(header));
```
paleolimbot, line 87: Should this be `header` or `&header`?

paleolimbot (commented):
Thanks!

I am not sure that this is actually running the nanoarrow read tests against your encoder (although after this PR your encoder is getting checked via Arrow C++). It may be worth checking `docker compose run --rm coverage` (a quick check on main suggests that most of `writer.c` is never run during tests, although this may be the coverage setup).

src/nanoarrow/ipc/encoder.c
bkietz: `header` works because it's an array and arrays are convertible to pointer. I prefer `&header` because that would be correct even if `header` were not an array

src/nanoarrow/ipc/encoder.c
bkietz: okay

src/nanoarrow/ipc/files_test.cc
bkietz: I think you're right. FWIW after reading from roundtripped instead of the file the tests still pass.

paleolimbot (approved):

PR #585, opened 2024-08-13
feat: Add IPC integration test executable

bkietz:
- Adds a new executable for use with the archery integration tests
  - When reading files, it ignores the Footer entirely
  - A few private workarounds are added, most notably the stream writer tracks Blocks in order to populate the Footer
- Tested against patched archery https://github.com/apache/arrow/compare/main...bkietz:arrow:nanoarrow-integration-tests
  - issue to add nanoarrow to the tests https://github.com/apache/arrow/issues/43680

### DIFF
--- a/src/nanoarrow/integration/ipc_integration.cc
+++ b/src/nanoarrow/integration/ipc_integration.cc
@@ -62,46 +62,6 @@ ArrowErrorCode JsonToArrow(struct ArrowError*);
 ArrowErrorCode StreamToFile(struct ArrowError*);
 ArrowErrorCode FileToStream(struct ArrowError*);
 
-int main(int argc, char** argv) try {
-  std::string command = GetEnv("COMMAND");
-
-  ArrowErrorCode error_code;
-  struct ArrowError error;
-
-  if (command == "VALIDATE") {
-    std::cout << "Validating that " << GetEnv("ARROW_PATH") << " reads identical to "
-              << GetEnv("JSON_PATH") << std::endl;
-
-    error_code = Validate(&error);
-  } else if (command == "JSON_TO_ARROW") {
-    std::cout << "Producing " << GetEnv("ARROW_PATH") << " from " << GetEnv("JSON_PATH")
-              << std::endl;
-
-    error_code = JsonToArrow(&error);
-  } else if (command == "STREAM_TO_FILE") {
-    error_code = StreamToFile(&error);
-  } else if (command == "FILE_TO_STREAM") {
-    error_code = FileToStream(&error);
-  } else {
-    if (argc == 1 || argv[1] == std::string{"-h"} || argv[1] == std::string{"--help"}) {
-      // skip printing usage if for example --gtest_list_tests is used;
-      // that command obviously doesn't need the extra noise
-      std::cerr << kUsage;
-    }
-    testing::InitGoogleTest(&argc, argv);
-    return RUN_ALL_TESTS();
-  }
-
-  if (error_code != NANOARROW_OK) {
-    std::cerr << "Command " << command << " failed (" << error_code << "="
-              << strerror(error_code) << "): " << error.message << std::endl;
-  }
-  return error_code;
-} catch (std::exception const& e) {
-  std::cerr << "Uncaught exception: " << e.what() << std::endl;
-  return 1;
-}
-
 struct File {
   ~File() {
     if (file_ != nullptr) {
@@ -370,4 +330,44 @@ TEST(Integration, ErrorMessages) {
     EXPECT_THAT(std::string(error.message),
                 testing::HasSubstr("Expected file of more than 8 bytes, got 3"));
   }
+}
+
+int main(int argc, char** argv) try {
+  std::string command = GetEnv("COMMAND");
+
+  ArrowErrorCode error_code;
+  struct ArrowError error;
+
+  if (command == "VALIDATE") {
+    std::cout << "Validating that " << GetEnv("ARROW_PATH") << " reads identical to "
+              << GetEnv("JSON_PATH") << std::endl;
+
+    error_code = Validate(&error);
+  } else if (command == "JSON_TO_ARROW") {
+    std::cout << "Producing " << GetEnv("ARROW_PATH") << " from " << GetEnv("JSON_PATH")
+              << std::endl;
+
+    error_code = JsonToArrow(&error);
+  } else if (command == "STREAM_TO_FILE") {
+    error_code = StreamToFile(&error);
+  } else if (command == "FILE_TO_STREAM") {
+    error_code = FileToStream(&error);
+  } else {
+    if (argc == 1 || argv[1] == std::string{"-h"} || argv[1] == std::string{"--help"}) {
+      // skip printing usage if for example --gtest_list_tests is used;
+      // that command obviously doesn't need the extra noise
+      std::cerr << kUsage;
+    }
+    testing::InitGoogleTest(&argc, argv);
+    return RUN_ALL_TESTS();
+  }
+
+  if (error_code != NANOARROW_OK) {
+    std::cerr << "Command " << command << " failed (" << error_code << "="
+              << strerror(error_code) << "): " << error.message << std::endl;
+  }
+  return error_code;
+} catch (std::exception const& e) {
+  std::cerr << "Uncaught exception: " << e.what() << std::endl;
+  return 1;
 }
--- a/src/nanoarrow/integration/ipc_integration.cc
+++ b/src/nanoarrow/integration/ipc_integration.cc
@@ -23,6 +23,8 @@
 #include <nanoarrow/nanoarrow_ipc.hpp>
 #include <nanoarrow/nanoarrow_testing.hpp>
 
+#define NANOARROW_IPC_FILE_PADDED_MAGIC "ARROW1\0"
+
 std::string GetEnv(char const* name) {
   char const* val = std::getenv(name);
   return val ? val : "";
--- a/src/nanoarrow/integration/ipc_integration.cc
+++ b/src/nanoarrow/integration/ipc_integration.cc
@@ -123,8 +123,7 @@ struct File {
 
     size_t bytes_read = 0;
     while (bytes_read < contents.size()) {
-      bytes_read +=
-          fread(contents.data() + bytes_read, 1, contents.size() - bytes_read, file_);
+      bytes_read += fread(&contents[bytes_read], 1, contents.size() - bytes_read, file_);
     }
     return contents;
   }
@@ -370,4 +369,3 @@ TEST(Integration, ErrorMessages) {
                 testing::HasSubstr("Expected file of more than 8 bytes, got 3"));
   }
 }
-
--- a/src/nanoarrow/integration/ipc_integration.cc
+++ b/src/nanoarrow/integration/ipc_integration.cc
@@ -18,47 +18,89 @@
 #include <cstdlib>
 #include <fstream>
 
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
 #include <nanoarrow/nanoarrow_ipc.hpp>
 #include <nanoarrow/nanoarrow_testing.hpp>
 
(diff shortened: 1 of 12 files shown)


src/nanoarrow/ipc/writer.c
```diff
@@ -275,6 +289,11 @@ ArrowErrorCode ArrowIpcWriterWriteArrayView(struct ArrowIpcWriter* writer,
                                     &private->buffer),
       error);
 
+  struct ns(Block) block = {private->offset, (int32_t) private->buffer.size_bytes,
+                            private->body_buffer.size_bytes};
+  private->offset += private->buffer.size_bytes + private->body_buffer.size_bytes;
+  NANOARROW_RETURN_NOT_OK(ArrowBufferAppend(&private->blocks, &block, sizeof(block)));
```
paleolimbot, line 295: Most people writing a stream will probably not want this...should there be a way to turn it off?

src/nanoarrow/ipc/json_integration.cc
```diff
@@ -0,0 +1,405 @@
...
+    ArrowErrorSet(error, "%s:%d: %s was null", __FILE__, __LINE__, #x); \
+    return ENOMEM;                                                      \
+  }
+
+  struct ArrowIpcEncoderPrivate {
+    flatcc_builder_t builder;
+    struct ArrowBuffer buffers;
+    struct ArrowBuffer nodes;
+    int encoding_footer;
+  };
```
paleolimbot, line 217: Given the tight dependence here, I imagine this should be in the encoder (with a basic test).

paleolimbot (commented):
I think this file would fit better as `integration/ipc_integration.cc`, next to the C data integration tests. You could separate them into a support lib/exe like Arrow C++, or you could build the `main()` function only if something is defined to ensure you can link to it from your tests.

I do think this would benefit from tests (e.g., to ensure error messages actually show up). I believe there are a few like that for the C Data integration functions (just enough to make sure everything is plugged in).

src/nanoarrow/ipc/json_integration.cc
```diff
@@ -0,0 +1,405 @@
...
+
+  operator FILE*() const { return file_; }
+
+  FILE* file_;
+};
+
+struct Table {
+  nanoarrow::UniqueSchema schema;
+  std::vector<nanoarrow::UniqueArray> batches;
+};
```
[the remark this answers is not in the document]
bkietz: okay

src/nanoarrow/ipc/writer.c
bkietz: will do

src/nanoarrow/ipc/json_integration.cc
bkietz: I'm not sure what you mean; are we going to make footer encoding public at this time after all?

paleolimbot: I think that would be better than copying the private data struct into another file (it can be "public" in the sense that it appears in the header but could be documented that it is currently only for internal use)

src/nanoarrow/ipc/encoder.c
```diff
@@ -412,10 +414,12 @@ ArrowErrorCode ArrowIpcEncoderEncodeSchema(struct ArrowIpcEncoder* encoder,
 
   flatcc_builder_t* builder = &private->builder;
 
-  FLATCC_RETURN_UNLESS_0(Message_start_as_root(builder), error);
-  FLATCC_RETURN_UNLESS_0(Message_version_add(builder, ns(MetadataVersion_V5)), error);
+  if (!private->encoding_footer) {
```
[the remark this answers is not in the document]
bkietz: How would you like to see them tested? We don't currently have anything which reads raw Footers, not even a public arrow C++ API which we could borrow. I could use flatcc to assert that flatcc wrote a valid Footer but that seems specious

src/nanoarrow/ipc/json_integration.cc
bkietz: okay

src/nanoarrow/ipc/encoder.c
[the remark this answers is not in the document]
paleolimbot: I think a "is it plugged in" check is fine (like, even just that it outputs a buffer with some minimum size).

src/nanoarrow/integration/ipc_integration.cc
```diff
@@ -0,0 +1,373 @@
...
+    std::cerr << "Command " << command << " failed (" << error_code << "="
+              << strerror(error_code) << "): " << error.message << std::endl;
+  }
+  return error_code;
+} catch (std::exception const& e) {
+  std::cerr << "Uncaught exception: " << e.what() << std::endl;
+  return 1;
+}
+
+struct File {
```
paleolimbot, line 103: No need to do it in this PR, but eventually we can consolidate/test the things that appear in both the C Data integration and IPC integration tests into nanoarrowarrow_testing

src/nanoarrow/integration/ipc_integration.cc
```diff
@@ -0,0 +1,373 @@
...
+    std::string contents(ftell(file_), '\0');
+    rewind(file_);
+
+    size_t bytes_read = 0;
+    while (bytes_read < contents.size()) {
+      bytes_read +=
+          fread(contents.data() + bytes_read, 1, contents.size() - bytes_read, file_);
+    }
+    return contents;
+  }
```
paleolimbot, line 130:
```suggestion
  std::string read() const {
    fseek(file_, 0, SEEK_END);
    std::vector<uint8_t> contents(ftell(file_), '\0');
    rewind(file_);

    size_t bytes_read = 0;
    while (bytes_read < contents.size()) {
      bytes_read +=
          fread(contents.data() + bytes_read, 1, contents.size() - bytes_read, file_);
    }
    return std::string(contents.begin(), contents.end());
  }
```

(This bit fails to compile for me...mabye `std::string::data` not being writable before C++17?)

```
[build] /Users/deweydunnington/Desktop/rscratch/arrow-nanoarrow/src/nanoarrow/integration/ipc_integration.cc:127:11: error: no matching function for call to 'fread'
[build]           fread(contents.data() + bytes_read, 1, contents.size() - bytes_read, file_);
[build]           ^~~~~
[build] /Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr/include/stdio.h:150:9: note: candidate function not viable: no known conversion from 'const value_type *' (aka 'const char *') to 'void *' for 1st argument
[build] size_t   fread(void * __restrict __ptr, size_t __size, size_t __nitems, FILE * __restrict __stream);
[build]          ^
```

src/nanoarrow/integration/ipc_integration.cc
```diff
@@ -0,0 +1,373 @@
...
+  } else if (command == "FILE_TO_STREAM") {
+    error_code = FileToStream(&error);
+  } else {
+    if (argc == 1 || argv[1] == std::string{"-h"} || argv[1] == std::string{"--help"}) {
+      // skip printing usage if for example --gtest_list_tests is used;
+      // that command obviously doesn't need the extra noise
+      std::cerr << kUsage;
+    }
+    testing::InitGoogleTest(&argc, argv);
+    return RUN_ALL_TESTS();
```
paleolimbot, line 90: Is there any way to separate this in such a way that we're not using the same executable for both running tests and as the artifact needed for the integration test setup? Maybe a define to separate the two cases if a separate file is too hard? (No need to solve if this is time consuming!)

src/nanoarrow/ipc/writer.c
```diff
@@ -329,3 +362,60 @@ ArrowErrorCode ArrowIpcWriterWriteArrayStream(struct ArrowIpcWriter* writer,
   ArrowArrayViewReset(&array_view);
   return NANOARROW_OK;
 }
+
+ArrowErrorCode ArrowIpcWriterStartFile(struct ArrowIpcWriter* writer,
+                                       struct ArrowError* error) {
```
paleolimbot, line 367: This would also benefit from a "is it plugged in" test (until we have time to circle back and finalize the API)

src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -417,6 +417,36 @@ ArrowErrorCode ArrowIpcArrayStreamReaderInit(
     struct ArrowArrayStream* out, struct ArrowIpcInputStream* input_stream,
     struct ArrowIpcArrayStreamReaderOptions* options);
 
+/// \brief The magic which appears at the beginning and end of an IPC file, 0-padded.
+#define NANOARROW_IPC_FILE_PADDED_MAGIC "ARROW1\0"
```
paleolimbot, line 421: Can this be kept private? I also find it a little odd that it's a string (would it get interpreted as a nul-terminated 6 byte string instead of an 8 byte magic header by accident?). (Fine as long as it works and is internal!)

src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -417,6 +417,36 @@ ArrowErrorCode ArrowIpcArrayStreamReaderInit(
...
+
+/// \brief Represents a byte range in an IPC file.
```
paleolimbot, line 423: Can these be moved to the end of the header, maybe outside the doxygen IPC group, and preceded by something along the lines of "internal API"?

src/nanoarrow/nanoarrow_ipc.h
bkietz: Well, it's tested that [`sizeof(NANOARROW_IPC_FILE_PADDED_MAGIC) == 8`](https://github.com/apache/arrow-nanoarrow/pull/585/files#diff-56dcfcba171647e18e8d2b6ce1d7cccd7ff94e44fe53fd87bcba7281a0c11fa7R33). I can make it internal for now

src/nanoarrow/ipc/writer.c
bkietz: Okay

src/nanoarrow/nanoarrow_ipc.h
paleolimbot: Yes, but `strlen(NANOARROW_IPC_FILE_PADDED_MAGIC)` would probably get you 6 (and because it's a string, somebody might do it)

src/nanoarrow/ipc/writer.c
```diff
@@ -329,3 +364,62 @@ ArrowErrorCode ArrowIpcWriterWriteArrayStream(struct ArrowIpcWriter* writer,
...
+  NANOARROW_RETURN_NOT_OK(
+      ArrowIpcEncoderEncodeFooter(&private->encoder, &private->footer, error));
+  NANOARROW_RETURN_NOT_OK_WITH_ERROR(
+      ArrowIpcEncoderFinalizeBuffer(&private->encoder, /*encapsulate=*/0,
+                                    &private->buffer),
+      error);
+
+  _NANOARROW_CHECK_RANGE(private->buffer.size_bytes, 0, INT32_MAX);
+  int32_t size = (int32_t) private->buffer.size_bytes;
+  struct ArrowStringView magic = ArrowCharView(NANOARROW_IPC_FILE_PADDED_MAGIC);
```
paleolimbot, line 410: I am fairly sure that this will only append 6 bytes to the end of the file (and I think you want 8?)

bkietz: No, at the end of the file we do not pad the magic

paleolimbot: I think a comment or `DCHECK(strlen(...) == 6)` or `ArrowBufferWrite()` with explicit size would be helpful (to avoid somebody who is probably future me thinking this is a mistake!)

paleolimbot (approved):
Thanks!

We can tackle testing the failure modes and/or consolidating integration test logic in a separate PR. I have full faith in your implementation but for when this is modified by a future contributor we do need a basic check to make sure that if there is a validation failure that it is actually reported (since that type of failure would be silently unnoticed in the normal integration test run).

PR #586, opened 2024-08-14
feat(python): Implement bindings to IPC writer

paleolimbot:
This PR implements bindings to the IPC writer in the nanoarrow C library. This adds:

- An `ipc.StreamWriter()` class roughly mirroring pyarrow's `ipc.Stream()`
- `Schema.serialize()` and `Array.serialize()` to match pyarrow's `serialize()` methods.

```python
import io
import nanoarrow as na 
from nanoarrow.ipc import StreamWriter, InputStream

out = io.BytesIO()
writer = StreamWriter.from_writable(out)
writer.write_stream(InputStream.example())

na.Array(InputStream.from_readable(out.getvalue()))
#> nanoarrow.Array<non-nullable struct<some_col: int32>>[3]
#> {'some_col': 1}
#> {'some_col': 2}
#> {'some_col': 3}
```

### DIFF
--- a/python/src/nanoarrow/ipc.py
+++ b/python/src/nanoarrow/ipc.py
@@ -272,13 +272,14 @@ class Writer:
         self._writer.write_end_of_stream()
         self.release()
 
-    def write(self, obj, schema=None):
+    def write(self, obj, schema=None, write_schema=True):
         if not self._is_valid():
             raise ValueError("Can't write to released Writer")
 
         with c_array_stream(obj, schema=schema) as stream:
             if self._iterator is None:
                 self._iterator = ArrayViewBaseIterator(stream._get_cached_schema())
+            if write_schema:
                 self._writer.write_schema(self._iterator._schema)
             for array in stream:
                 self._iterator._set_array(array)
--- a/python/src/nanoarrow/ipc.py
+++ b/python/src/nanoarrow/ipc.py
@@ -243,7 +243,6 @@ class Stream:
 
 
 class Writer:
-
     def __init__(self):
         self._writer = None
         self._desc = None
--- a/python/src/nanoarrow/ipc.py
+++ b/python/src/nanoarrow/ipc.py
@@ -26,6 +26,7 @@ from nanoarrow._ipc_lib import (
 )
 from nanoarrow._utils import obj_is_buffer
 from nanoarrow.array_stream import c_array_stream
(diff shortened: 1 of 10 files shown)


python/src/nanoarrow/ipc.py
```diff
@@ -236,9 +243,198 @@ def __repr__(self) -> str:
             return f"<{class_label} <invalid>>"
 
 
+class Writer:
```
jorisvandenbossche, line 246: Add a class docstring for Writer?

jorisvandenbossche: Also, to understand a bit the different classes here. The already existing `Stream` object is essentially the "reader" equivalent?

And for this writer, does that only support the stream format or also the file format?

python/src/nanoarrow/ipc.py
```diff
@@ -236,9 +243,198 @@ def __repr__(self) -> str:
...
+
+    def close(self):
+        """Close stream and write end-of-stream marker"""
+        if not self._is_valid():
+            return
+
+        self._writer.write_end_of_stream()
+        self.release()
+
+    def write_stream(self, obj, schema=None, *, write_schema=None):
```
jorisvandenbossche, line 277: There could in theory also be a `write_array` (or maybe rather `write_batch`) method that takes a single array to write to the stream?

jorisvandenbossche (commented):
Looks good! 
Added some minor comments/questions, didn't go through everything, but already posting what I have for today

python/src/nanoarrow/ipc.py
paleolimbot: `Stream` is a bit of a compromise...we can't just "subclass a record batch reader" like is done in Arrow C++/Pyarrow, but we need a place to put the code that deals with file IO...so the stream is sort of an internal stepping stone to eliminate anything dealing with files in `nanoarrow.array_stream.ArrayStream`. It should have a better name now that its purpose is more clear.

`Writer` is definitely a user-facing concept and should probably be a `StreamWriter` (since it doesn't do files)

python/src/nanoarrow/ipc.py
paleolimbot: There could be! An array can be implicitly converted to a stream and will work here, but I'll add one for symmetry (and some day it might be faster!)

PR #591, opened 2024-08-20
fix: Ensure that schema metadata is always present even if empty

bkietz:
An absent Schema.custom_metadata was causing flatcc to produce invalid flatbuffers on s390x. For now, just populate that field with an empty vector

### DIFF
--- a/src/nanoarrow/integration/ipc_integration.cc
+++ b/src/nanoarrow/integration/ipc_integration.cc
@@ -257,6 +257,7 @@ ArrowErrorCode Validate(struct ArrowError* error) {
   NANOARROW_RETURN_NOT_OK(arrow_table.FromIpcFile(arrow_path, error));
 
   nanoarrow::testing::TestingJSONComparison comparison;
+  comparison.set_compare_metadata_order(false);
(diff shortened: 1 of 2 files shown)


src/nanoarrow/integration/ipc_integration.cc
```diff
@@ -257,6 +257,7 @@ ArrowErrorCode Validate(struct ArrowError* error) {
   NANOARROW_RETURN_NOT_OK(arrow_table.FromIpcFile(arrow_path, error));
 
   nanoarrow::testing::TestingJSONComparison comparison;
+  comparison.set_compare_metadata_order(false);
```
paleolimbot, line 260: I thought this change was already merged?

paleolimbot (approved):
I am not familiar with the details here, but thank you!

src/nanoarrow/integration/ipc_integration.cc
bkietz: Indeed... https://github.com/apache/arrow-nanoarrow/blob/main/src/nanoarrow/integration/ipc_integration.cc#L260

I don't know why this is showing up here.

PR #592, opened 2024-08-20
feat: Revendor flatcc

bkietz:
After https://github.com/dvidelabs/flatcc/pull/289 flatcc will not enforce alignment on empty vectors. This will let us unskip the null_trivial case in the integration tests


paleolimbot (approved):
Thanks!

I think it just need a `cmake-format -i CMakeLists.txt` and to possibly add `flatcc_generated.h` to the files we ignore for `clang-format` in `.pre-commit.yaml`.

bkietz:
> to possibly add `flatcc_generated.h` to the files we ignore

It is already excluded, and everything in `thirdparty/` too. However the CI job uses `--all-files` which overrides exclusions. This seems to have been the case since `pre-commit` was added to the project, so I assume we want to remove that flag now

Issue #583, opened 2024-08-13
Add string view support


WillAyd:
I've noticed polars in particular has started to export Arrow C data with the "vu" format string for string views, but I don't see any support for those in nanoarrow. Is this something that can be added easily? I think the same API we have in place for strings can be used for string views as well?

@bkietz probably knows more

WillAyd:
Sounds good. Will take a look if I get some free time

paleolimbot:
Great! No worries if you aren't able to get to it, but post a note here if you do so I don't step on your toes!

PR #596, opened 2024-08-23
feat: String/Binary View Support

WillAyd:
closes https://github.com/apache/arrow-nanoarrow/issues/583


src/nanoarrow/common/inline_array.h
```diff
@@ -938,6 +954,22 @@ static inline struct ArrowStringView ArrowArrayViewGetStringUnsafe(
...
+      const union ArrowBufferViewData value_view = array_view->buffer_views[1].data;
+      union ArrowBinaryViewType bvt;
+      const size_t idx = sizeof(union ArrowBinaryViewType) * i;
+      memcpy(&bvt, value_view.as_uint8 + idx, sizeof(union ArrowBinaryViewType));
+      const int32_t inline_size = bvt.inlined.size;
+      view.size_bytes = inline_size;
+      if (inline_size <= ARROW_VIEW_INLINE_SIZE) {
+        view.data = value_view.as_char + idx + sizeof(int32_t);
+      } else {
+        const int32_t buf_index = bvt.ref.buffer_index + 2;
```
WillAyd, line 968: This isn't working at the moment. The inline data seems to work, but for the non-inline values the children buffers are not initialized.

I assume something has to be changed to accommodate this in or around `ArrowArrayViewSetArrayInternal`

WillAyd:
Hmm looks like the Python process whereby we generate the pxd file does not play nice with the union containing nested struct declarations

WillAyd:
Sounds good. Enjoy the time away!

src/nanoarrow/common/inline_array.h
```diff
@@ -914,6 +914,29 @@ static inline double ArrowArrayViewGetDoubleUnsafe(
   }
 }
 
+enum { ARROW_BINARY_VIEW_INLINE_SIZE = 12 };
```
WillAyd, line 917: I originally had these as macros but the Cython generator did not like that. However, it doesn't appear to like the enum hack either.

Any ideas on what the best practice for the library here is @paleolimbot ?

paleolimbot: Perhaps more like "existing hack" than "best practice", but:

https://github.com/apache/arrow-nanoarrow/blob/ee26444fbfafb47d67b49d8920570d71bae828d0/python/bootstrap.py#L40-L41

WillAyd:
@paleolimbot started to look at this in more earnest to get a full round trip, instead of just a read. The current read implemenation also only assumes one variadic buffer.

I see the Arrow specification says that binary view types have an extra member which is a buffer containing the lengths of the variadic buffers:

https://arrow.apache.org/docs/format/CDataInterface.html#binary-view-arrays

Do you think its worth adding that to the `ArrowArray` struct, or should we create a new struct, potentially with a layout like:

```c
struct ArrayBinaryViewArray {
  struct ArrowArray array_data;
  // Variadic buffer lengths for Binary View arrays
  int64_t* variadic_buffer_lengths;
};
```

So that it can be "safely" cast to/from ArrowArray pointers?

WillAyd:
With that approach the one thing I still wonder is where we should store the length of `variadic_buffer_lengths`. I didn't see anything for that in the specification, so maybe using a NULL sentinel to denote the end of the array is the requirement?

WillAyd:
Ah OK interesting. Do you know if this is confirmed to be working upstream in Arrow? The reason I ask is that when running the test suite for what's already in this PR, when I hit the C++ Export method:

https://github.com/apache/arrow/blob/c2123b8b90ab952f854912459bb33ebaf0d99611/cpp/src/arrow/c/bridge.cc#L661

The `n_buffers` being assigned is just 3. I was (likely mistakenly) expecting that to be 4 to account for the validity buffer, inline/prefix buffer, out of line buffer, and a buffer that holds the sizes to all out of line buffers

The Arrow docs also read a little vague to me:

> int64_t ArrowArray.n_buffers
>
>    Mandatory. The number of physical buffers backing this array. The number of buffers is a function of the data type, as described in the [Columnar format specification](https://arrow.apache.org/docs/format/Columnar.html#format-columnar), except for the the binary or utf-8 view type, which has one additional buffer compared to the Columnar format specification (see [Binary view arrays](https://arrow.apache.org/docs/format/CDataInterface.html#c-data-interface-binary-view-arrays)).
>
>    Buffers of children arrays are not included.

I'm not sure if that is saying that the `n_buffers` argument should be 4 for the view types, or that it should be 3 and the consumer is expected to know that when working with view types, you actually have `n_buffers + 1` buffers attached to the `ArrowArray`

WillAyd:
Bah ignore what I just said - bad debugging. Digging deeper!

src/nanoarrow/common/array.c
```diff
@@ -720,7 +731,40 @@ static int ArrowArrayViewSetArrayInternal(struct ArrowArrayView* array_view,
...
   }
 
-  // Check the number of buffers
+  if (array_view->storage_type == NANOARROW_TYPE_STRING_VIEW ||
+      array_view->storage_type == NANOARROW_TYPE_BINARY_VIEW) {
+    const int64_t n_buffers = array->n_buffers;
+    const int64_t n_fixed_buffers = 2;
+
+    int64_t n_variadic_buffers = n_buffers - n_fixed_buffers - 1;
+    // Theoretically if the variadic buffers are not used the above subtraction
```
WillAyd, line 740: Not sure if this is true - need to add a test when there is an empty StringView array

src/nanoarrow/common/array.c
```diff
@@ -720,7 +731,40 @@ static int ArrowArrayViewSetArrayInternal(struct ArrowArrayView* array_view,
...
+    array_view->variadic_buffer_views =
+        ArrowMalloc(sizeof(struct ArrowBufferView) * n_variadic_buffers);
+
+    for (int64_t i = 0; i < n_variadic_buffers; ++i) {
+      ++buffers_required;
+      array_view->variadic_buffer_views[i].data.data =
+          array->buffers[i + n_fixed_buffers];
+
+      // TODO: why do we need this?
+      // If non-null, set buffer size to unknown.
```
WillAyd, line 755: This was modeled from the prior loop for fixed buffers, although I don't fully understand why the sizes are set like this

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,99 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
   struct ArrowArrayPrivateData* private_data =
       (struct ArrowArrayPrivateData*)array->private_data;
 
-  struct ArrowBuffer* offset_buffer = ArrowArrayBuffer(array, 1);
-  struct ArrowBuffer* data_buffer = ArrowArrayBuffer(
-      array, 1 + (private_data->storage_type != NANOARROW_TYPE_FIXED_SIZE_BINARY));
-  int32_t offset;
-  int64_t large_offset;
-  int64_t fixed_size_bytes = private_data->layout.element_size_bits[1] / 8;
+  if (private_data->storage_type == NANOARROW_TYPE_STRING_VIEW ||
```
WillAyd, line 500: Changes to this function are in an intermediate state - need to redo this part

src/nanoarrow/common/inline_types.h
```diff
@@ -808,6 +815,15 @@ struct ArrowArrayView {
   /// type_id == union_type_id_map[128 + child_index]. This value may be
   /// NULL in the case where child_id == type_id.
   int8_t* union_type_id_map;
+
+  /// \brief Number of variadic buffers
```
WillAyd, line 819: @paleolimbot this PR is not quite ready for a full PR review, but figured I'd ping some more usability questions off you.

As you suggested, I took these members from the great work that @jorisvandenbossche started. Assuming we have `n_buffers` of 4, two of those buffers are accessible from the `buffer_views` member (the validity mask and inline/prefix buffer), one buffer is accessible from `variadic_buffer_views`, and the last is accessible from `variadic_buffer_sizes`

Is that what you had in mind?

paleolimbot: I think so! (I haven't played with the C++ or pyarrow implementations yet but I believe that was the result of the discussion).

python/src/nanoarrow/_types.pyx
```diff
@@ -181,10 +181,10 @@ cdef int to_format(int type_id, int element_size_bits, size_t out_size, char* ou
 
     cdef const char* format_const = ""
     cdef int element_size_bits_calc = 0
-    if type_id == _types.STRING:
+    if type_id in (_types.STRING, _types.LARGE_STRING, _types.STRING_VIEW):
```
paleolimbot, line 184: I am not sure this applies here...these are "buffer" types and a "buffer" type of STRING_VIEW would probably be the one with the fixed-size preview and length (the "buffer" type of the subsequent buffers would probably be STRING). (We have a little bit hijacked the `ArrowType` enum to avoid creating an `ArrowBufferDataType` enum that would be a subset of that enum)

src/nanoarrow/common/array.c
```diff
@@ -642,6 +673,10 @@ void ArrowArrayViewReset(struct ArrowArrayView* array_view) {
     ArrowFree(array_view->union_type_id_map);
   }
 
+  if (array_view->variadic_buffer_views != NULL) {
+    ArrowFree(array_view->variadic_buffer_views);
```
paleolimbot, line 677: It seems expensive to maintain this member...maybe a function like `ArrowArrayViewVariadicBufferView(array_view, i)` could materialize a particular one on demand rather than creating them all at once?

src/nanoarrow/common/array.c
paleolimbot: The buffer sizes are set to the -1 unknown sentinel because for non-CPU data we can't know (some of) them before accessing the buffer content themselves, and for non-CPU data that takes some orchestration to not be horribly inefficient.

I think here you probably don't want to loop over the buffers..."set array" is supposed to be a very cheap operation and this would potentially be a loop over all n items in the array.

src/nanoarrow/common/array_test.cc
```diff
@@ -3079,6 +3129,20 @@ void TestGetFromBinary(BuilderClass& builder) {
   EXPECT_EQ(buffer_view.size_bytes, strlen("four"));
   EXPECT_EQ(memcmp(buffer_view.data.as_char, "four", buffer_view.size_bytes), 0);
 
+  if constexpr (!std::is_same_v<BuilderClass, FixedSizeBinaryBuilder>) {
```
paleolimbot, line 3132: Just a note that we're still on C++11 for the tests 🙂

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +469,116 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
+};
+
+struct ArrowBinaryViewTypeRefData {
+  int32_t size;
+  uint8_t data[NANOARROW_BINARY_VIEW_PREVIEW_SIZE];
+  int32_t buffer_index;
+  int32_t offset;
+};
+
+union ArrowBinaryViewType {  // TODO: C++ impl uses alignas which comes in C11
```
paleolimbot, line 491: I believe you can use a `union` to force alignment of a member in C99 (e.g., `union { int64_t dummy; char data; }` would align as an int64 (but double check that!)

src/nanoarrow/testing/testing.cc
```diff
@@ -1932,6 +1940,7 @@ ArrowErrorCode SetArrayColumnBuffers(const json& value, ArrowArrayView* array_vi
       }
       break;
     }
+    case NANOARROW_BUFFER_TYPE_DATA_VIEW:  // TODO: doesn't seem right?
```
paleolimbot, line 1943: You can probably punt on this for this PR...it's basically generating this: https://arrow.apache.org/docs/format/Integration.html#json-test-data-format , scroll down to `FieldData`.

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +469,116 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
+    } else {
+      int32_t n_vbufs = private_data->n_variadic_buffers;
+      if (n_vbufs == 0 ||
+          private_data->variadic_buffers[n_vbufs - 1].size_bytes + value.size_bytes >
+              NANOARROW_BINARY_VIEW_BLOCK_SIZE) {
+        // TODO: ArrowMalloc?
+        ++n_vbufs;
+        // TODO: these should be realloc for > 0 case
+        private_data->variadic_buffers =
+            (struct ArrowBuffer*)malloc(sizeof(struct ArrowBuffer) * (n_vbufs));
```
WillAyd, line 519: I assume the `inline_` headers intentionally do not import from anywhere else, but `ArrowMalloc` is defined in `nanoarrow.h` with an implementation in `utils.c`

Would it be preferable to inline that definition/declaration into `inline_buffer.h`, or do we just want to stick with malloc here?

paleolimbot: It should compile with `ArrowMalloc()` because of how `nanoarrow.h` is set up (it includes the inline versions after all the appropriate declarations); however, this causes some IDE trouble occasionally and is not very clean. You could just use `ArrowMalloc()` for this PR, or fix that in another PR (briefly: move the namespacing to `inline_types.h` and move the `nanoarrow-malloc` group of functions to `inline_buffer.h` should do it).

src/nanoarrow/common/array.c
paleolimbot: ...or, revisiting this, `ArrowBufferView ArrowArrayViewBufferView(array_view, i)` (which would either pick up the buffer view from the fixed buffers array or from the variadic buffers if they exist).

WillAyd: Assuming we go down the path of `ArrowArrayViewBufferView(array_view, i)` do you think its worth moving existing code away from using the `buffer_views` member? I think part of the awkwardness right now is how heavily that is used, but since it is a stack allocated array of `BufferView` objects it can't be re-used here

paleolimbot: I think so (although maybe in a later PR since it's a fairly wide-reaching change?)

src/nanoarrow/common/array.c
```diff
@@ -449,14 +462,33 @@ static ArrowErrorCode ArrowArrayFinalizeBuffers(struct ArrowArray* array) {
     NANOARROW_RETURN_NOT_OK(ArrowArrayFinalizeBuffers(array->dictionary));
   }
 
+  if (private_data->storage_type == NANOARROW_TYPE_STRING_VIEW ||
+      private_data->storage_type == NANOARROW_TYPE_BINARY_VIEW) {
+    const int32_t nfixed_buf = 2;
+    const int32_t n_vbuf = private_data->n_variadic_buffers;
+    private_data->buffer_data = (const void**)realloc(
```
paleolimbot, line 469:
```suggestion
    private_data->buffer_data = (const void**)ArrowRealloc(
```

src/nanoarrow/common/array.c
```diff
@@ -449,14 +462,33 @@ static ArrowErrorCode ArrowArrayFinalizeBuffers(struct ArrowArray* array) {
     NANOARROW_RETURN_NOT_OK(ArrowArrayFinalizeBuffers(array->dictionary));
   }
 
+  if (private_data->storage_type == NANOARROW_TYPE_STRING_VIEW ||
+      private_data->storage_type == NANOARROW_TYPE_BINARY_VIEW) {
```
paleolimbot, line 466: I think this should be in `ArrowArrayFlushInternalPointers()` (if I remember correctly, the "finalize buffers" step is skipped for non CPU data)

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,113 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
   return NANOARROW_OK;
 }
 
+#define NANOARROW_BINARY_VIEW_INLINE_SIZE 12
+#define NANOARROW_BINARY_VIEW_PREVIEW_SIZE 4
+#define NANOARROW_BINARY_VIEW_BLOCK_SIZE 32 << 10  // 32KB
```
paleolimbot, line 473: For the purposes of this PR, I wonder if you can omit the chunking logic (i.e., when building a string/binary view array, simply use the third fixed buffer as your one variadic buffer and always append the data to that one).

For "build by buffer" we would need to support `ArrowArraySetBuffer()` on the variadic buffers (which would let a user implement this logic themselves), and perhaps a `union ArrowBinaryViewType ArrowBinaryViewInlined()`/`union ArrowBinaryViewType ArrowBinaryViewReference()` helper (if I didn't miss one you made already!)

src/nanoarrow/common/inline_array.h
```diff
@@ -810,6 +874,11 @@ static inline int64_t ArrowArrayViewListChildOffset(
   }
 }
 
+static inline struct ArrowBufferView ArrowArrayViewBufferView(
+    const struct ArrowArrayView* array_view, int64_t i) {
+  return array_view->buffer_views[i];
```
paleolimbot, line 879: Does this need to be `if(...) return array-view->buffer_views[i]; else return array_view->variadic_buffer_views[i - 2];`?

paleolimbot (commented):
I know you're still working on this but just a few notes!

src/nanoarrow/common/array.c
```diff
@@ -148,12 +156,17 @@ ArrowErrorCode ArrowArrayInitFromType(struct ArrowArray* array,
...
   ArrowBufferInit(&private_data->buffers[0]);
   ArrowBufferInit(&private_data->buffers[1]);
-  private_data->buffer_data[0] = NULL;
-  private_data->buffer_data[1] = NULL;
-  private_data->buffer_data[2] = NULL;
+  private_data->buffer_data =
+      (const void**)ArrowMalloc(sizeof(void*) * NANOARROW_MAX_FIXED_BUFFERS);
+  for (int i = 0; i < NANOARROW_MAX_FIXED_BUFFERS; ++i) {
+    private_data->buffer_data[i] = NULL;
+  }
```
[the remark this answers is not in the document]
WillAyd: Makes sense - will go back down that path

src/nanoarrow/common/inline_array.h
WillAyd: Most of the plumbing was already in place so I ended up implementing the chunking logic for appends, at least to match what the Arrow C++ implementation does (I think)

src/nanoarrow/common/array_test.cc
```diff
@@ -3101,6 +3241,89 @@ TEST(ArrayViewTest, ArrayViewTestGetString) {
...
+
+  // TODO: is this a bug in the Arrow C++ export? Within gdb I see
+  // (gdb) p ((int32_t*)array->buffers[4])[0]
+  // $34 = 32766
+  // (gdb) p ((int32_t*)array->buffers[4])[1]
+  // $35 = 0
+  // (gdb) p ((int32_t*)array->buffers[4])[2]
+  // $36 = 32
+  // where does the 0 for the second array element come from?
+  EXPECT_EQ(array_view.variadic_buffer_sizes[2], str2.size());
```
WillAyd, line 3285: Per comment above - this is a bug with the Arrow C++ implementation I think? Unless the variadic buffer is intentionally 64 bits, although the individual buffers only allow up to 32 bits of offset storage?

WillAyd: I did confirm this exists in the implementation this way; opened https://github.com/apache/arrow/issues/44165 upstream to ask about it

paleolimbot: I think it is intentional! https://arrow.apache.org/docs/format/CDataInterface.html#binary-view-arrays

src/nanoarrow/common/inline_array.h
paleolimbot: I can add the "build by buffer" stuff in a future PR...the chunking is great to have for appends and ensures you have some test structures to roundtrip!

src/nanoarrow/common/array_test.cc
WillAyd: Ah great - thanks!

src/nanoarrow/common/array.c
paleolimbot: I know you're still working here, but just a bump that I don't think we want `array_view->variadic_buffer_view` to exist, since it will involve a potentially very large allocation on `ArrowArrayViewSetArray()`. Those are buffer views that we can materialize on demand by either adding `ArrowBufferView ArrowArrayViewVariadicBuffer(array_view, i)` or `ArrowArrayViewBuffer(array_view, i)`.

paleolimbot (commented):
A few more notes! I am gearing up for a release and it would be very cool to include this if we can...what do you see as blockers for getting this PR across the finish line and can I help with any of them?

python/tests/test_schema.py
```diff
@@ -107,6 +107,9 @@ def test_schema_simple():
     assert na.interval_months().type == na.Type.INTERVAL_MONTHS
     assert na.interval_day_time().type == na.Type.INTERVAL_DAY_TIME
     assert na.interval_month_day_nano().type == na.Type.INTERVAL_MONTH_DAY_NANO
+    # TODO: run_end_encoded
+    assert na.binary_view().type == na.Type.BINARY_VIEW
```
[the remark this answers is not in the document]
WillAyd: Having a hard time deciphering the code suggestion - do you just want me to remove these?

WillAyd:
> A few more notes! I am gearing up for a release and it would be very cool to include this if we can...what do you see as blockers for getting this PR across the finish line and can I help with any of them?

Great! I should have time to see this through.

There are no major blockers from my end. Let me know any feedback you have and I will take a look

src/nanoarrow/common/utils.c
```diff
@@ -179,6 +179,16 @@ void ArrowLayoutInit(struct ArrowLayout* layout, enum ArrowType storage_type) {
...
 
+    case NANOARROW_TYPE_BINARY_VIEW:
+      layout->buffer_type[1] = NANOARROW_BUFFER_TYPE_DATA_VIEW;
+      layout->buffer_data_type[1] = NANOARROW_TYPE_BINARY_VIEW;
+      layout->element_size_bits[1] = 128;
+      break;
+    case NANOARROW_TYPE_STRING_VIEW:
+      layout->buffer_type[1] = NANOARROW_BUFFER_TYPE_DATA_VIEW;
+      layout->buffer_data_type[1] = NANOARROW_TYPE_STRING_VIEW;
+      layout->element_size_bits[1] = 128;
```
paleolimbot, line 190: This can be tested with something like this:

https://github.com/apache/arrow-nanoarrow/blob/5d53738636cac9d6bf14aebf791a123814a00697/src/nanoarrow/common/schema_test.cc#L826

src/nanoarrow/common/array.c
```diff
@@ -456,10 +470,26 @@ static void ArrowArrayFlushInternalPointers(struct ArrowArray* array) {
...
+  const int32_t nfixed_buf = is_binary_view ? 2 : NANOARROW_MAX_FIXED_BUFFERS;
+
+  for (int32_t i = 0; i < nfixed_buf; i++) {
     private_data->buffer_data[i] = ArrowArrayBuffer(array, i)->data;
   }
 
+  if (is_binary_view) {
+    const int32_t nvirt_buf = private_data->n_variadic_buffers;
+    private_data->buffer_data = (const void**)ArrowRealloc(
+        private_data->buffer_data, sizeof(void*) * (nfixed_buf + nvirt_buf + 1));
```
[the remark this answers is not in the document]
WillAyd: The allocation for `private_data->buffer_data` happens up front in `ArrowArrayInitFromType` and only reserves enough for `NANOARROW_FIXED_SIZE_BUFFERS`; in the case of having variadic buffers, you wouldn't really know what extra space needs to be allocated until you know how many variadic buffers you are dealing with.

Maybe there is another place you think is more suitable for this reallocation? But I don't _think_ this can be done up front in the Init function

src/nanoarrow/common/array_test.cc
```diff
@@ -895,6 +895,146 @@ TEST(ArrayTest, ArrayTestAppendToLargeStringArray) {
...
+  // TODO: need to add overload for ViewArrayAsBytes
+  /*
+  EXPECT_THAT(nanoarrow::ViewArrayAsBytes<64>(&array),
+              ElementsAre("1234"_asv, NA, NA, "56789"_asv, ""_asv));
+  */
+  ArrowArrayRelease(&array);
+}
+
+TEST(ArrayTest, ArrayTestAppendToBinaryViewArray) {
+  struct ArrowArray array;
```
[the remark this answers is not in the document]
WillAyd: Are you thinking that function should be added in the test suite or nanoarrow itself?

src/nanoarrow/common/array.c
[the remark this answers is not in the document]
paleolimbot: Ignore me...I was mixing up the the `ArrowBuffer*` array and the `const void**` array :flushed:

src/nanoarrow/testing/testing.cc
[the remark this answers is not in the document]
WillAyd: OK no problem - I think will still keep the case statement for the DATA_VIEW enumeration that has been added - otherwise will get a warning during compilation

```
../src/nanoarrow/testing/testing.cc:1833:10: warning: enumeration value ‘NANOARROW_BUFFER_TYPE_DATA_VIEW’ not handled in switch [-Wswitch]
 1833 |   switch (array_view->layout.buffer_type[buffer_i]) {
      |          ^
```

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,111 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
-      offset = ((int32_t*)offset_buffer->data)[array->length];
-      if ((((int64_t)offset) + value.size_bytes) > INT32_MAX) {
-        return EOVERFLOW;
+    if (value.size_bytes <= NANOARROW_BINARY_VIEW_INLINE_SIZE) {
+      memcpy(bvt.inlined.data, value.data.as_char, value.size_bytes);
+    } else {
+      int32_t n_vbufs = private_data->n_variadic_buffers;
+      if (n_vbufs == 0 ||
+          private_data->variadic_buffers[n_vbufs - 1].size_bytes + value.size_bytes >
+              NANOARROW_BINARY_VIEW_BLOCK_SIZE) {
```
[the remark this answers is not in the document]
WillAyd: Is the `buffer_index` in this case supposed to point to the index of the first buffer that has been added? If so, any preference for the name of that parameter?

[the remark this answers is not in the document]
paleolimbot: Mostly just to tell you which one just got added (so that you can put a reference to it in the item you're about to append to the data buffer). Perhaps there's a better parameter name (or a way to retrieve the number of currently used variadic buffers using a function?)

src/nanoarrow/testing/testing.cc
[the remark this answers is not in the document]
paleolimbot: Got it! You should be able to return ENOTSUP there (which might be a better choice than a silent skip for when we get here).

src/nanoarrow/common/array_test.cc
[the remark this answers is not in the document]
paleolimbot: Just as a test pattern (like you added below for consolidating the builder tests):

```c
void TestAppendToDataViewArray(enum ArrowType view_type) {
  // ...
}

TEST(ArrayTest, ArrayTestAppendToBinaryViewArray) {
  ASSERT_NO_FATAL_FAILURE(TestAppendToDataViewArray(NANOARROW_TYPE_BINARY_VIEW));
}

TEST(ArrayTest, ArrayTestAppendToStringViewArray) {
  ASSERT_NO_FATAL_FAILURE(TestAppendToDataViewArray(NANOARROW_TYPE_STRING_VIEW));
}
```

python/tests/test_schema.py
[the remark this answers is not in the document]
bkietz: I think it's just that you can remove the TODO

src/nanoarrow/common/inline_array.h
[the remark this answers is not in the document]
WillAyd: Yea...struggling with what to name this, especially if we allow this function to add multiple buffers at once. Maybe the second function to get the number of buffers allocated is a better pattern?

So

```c
int64_t nbuf_capacity = ArrowArrayVariadicBufferSize(struct ArrowArray *array);
int64_t additional_buffers = 3;
NANOARROW_RETURN_NOT_OK(ArrowArrayAddVariadicBuffers(array, additional_buffers));
// nbuf_capacity + additional_buffers == ArrowArrayVariadicBufferSize(struct ArrowArray *array);
```

versus

```c
int64_t nbuf_capacity = private_data->n_variadic_buffers;
int64_t next_buffer_avail;
int64_t additional_buffers = 3;
NANOARROW_RETURN_NOT_OK(ArrowArrayAddVariadicBuffers(array, additional_buffers, &next_buffer_avail));
// next_buffer_avail + additional_buffers == private_data->n_variadic_buffers;
```

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,136 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
+    bvt.inlined.size = (int32_t)value.size_bytes;
 
-  switch (private_data->storage_type) {
-    case NANOARROW_TYPE_STRING:
-    case NANOARROW_TYPE_BINARY:
-      offset = ((int32_t*)offset_buffer->data)[array->length];
-      if ((((int64_t)offset) + value.size_bytes) > INT32_MAX) {
-        return EOVERFLOW;
+    if (value.size_bytes <= NANOARROW_BINARY_VIEW_INLINE_SIZE) {
+      memcpy(bvt.inlined.data, value.data.as_char, value.size_bytes);
```
bkietz, line 538: The standard also requires that all bytes after `value.size_bytes` are 0
```suggestion
      memcpy(bvt.inlined.data, value.data.as_char, value.size_bytes);
      memset(btt.inlined.data + bvt.inlined.size, 0, NANOARROW_BINARY_VIEW_INLINE_SIZE - bvt.inlined.size);
```

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,136 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
+    } else {
+      const int32_t n_vbufs = private_data->n_variadic_buffers;
+      int64_t buf_index = n_vbufs - 1;
+      if (n_vbufs == 0 ||
+          private_data->variadic_buffers[n_vbufs - 1].size_bytes + value.size_bytes >
+              NANOARROW_BINARY_VIEW_BLOCK_SIZE) {
+        NANOARROW_RETURN_NOT_OK(ArrowArrayAddVariadicBuffers(array, 1, &buf_index));
```
bkietz, line 545: This doesn't handle the case where `value.size_bytes > NANOARROW_BINARY_VIEW_BLOCK_SIZE`, in that case you'll need an extra large variadic buffer to accommodate it

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,136 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
+// of the ArrowBinaryViewType. For Cython support in this library, we define
+// those structs outside of the ArrowBinaryViewType
+struct ArrowBinaryViewTypeInlinedData {
+  int32_t size;
+  uint8_t data[NANOARROW_BINARY_VIEW_INLINE_SIZE];
+};
+
+struct ArrowBinaryViewTypeRefData {
+  int32_t size;
+  uint8_t data[NANOARROW_BINARY_VIEW_PREVIEW_SIZE];
```
bkietz, line 486:
```suggestion
  uint8_t prefix[NANOARROW_BINARY_VIEW_PREVIEW_SIZE];
```

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,136 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
+    return ENOMEM;
+  }
+  private_data->variadic_buffer_sizes = (int64_t*)ArrowRealloc(
+      private_data->variadic_buffer_sizes, sizeof(int64_t) * n_bufs_needed);
+  if (private_data->variadic_buffer_sizes == NULL) {
+    return ENOMEM;
+  }
+
+  for (int32_t i = 0; i < nbuffers; i++) {
+    ArrowBufferInit(&private_data->variadic_buffers[n_current_bufs + i]);
```
bkietz, line 518:
```suggestion
    ArrowBufferInit(&private_data->variadic_buffers[n_current_bufs + i]);
    private_data->variadic_buffer_sizes[i] = 0;
```

src/nanoarrow/common/inline_array.h
```diff
@@ -467,52 +468,136 @@ static inline ArrowErrorCode ArrowArrayAppendDouble(struct ArrowArray* array,
...
-      offset += (int32_t)value.size_bytes;
-      NANOARROW_RETURN_NOT_OK(ArrowBufferAppend(offset_buffer, &offset, sizeof(int32_t)));
+      struct ArrowBuffer* variadic_buf = &private_data->variadic_buffers[buf_index];
+      memcpy(bvt.ref.data, value.data.as_char, NANOARROW_BINARY_VIEW_PREVIEW_SIZE);
+      bvt.ref.buffer_index = (int32_t)buf_index;
+      bvt.ref.offset = (int32_t)variadic_buf->size_bytes;
       NANOARROW_RETURN_NOT_OK(
-          ArrowBufferAppend(data_buffer, value.data.data, value.size_bytes));
-      break;
+          ArrowBufferAppend(variadic_buf, value.data.as_char, value.size_bytes));
```
bkietz, line 553: This is fine for now, but the more usual pattern for building view arrays is to preallocate the data buffers (which reduces the frequency of reallocation)

src/nanoarrow/common/inline_array.h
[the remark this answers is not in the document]
paleolimbot: > Maybe the second function to get the number of buffers allocated is a better pattern?

Yes! Maybe `ArrowArrayVariadicBufferCount()` though? (Invoking `Size` suggests to me that there's only one of them and that the function will tell you how many bytes there are).

src/nanoarrow/common/inline_array.h
```diff
@@ -972,6 +1101,26 @@ static inline struct ArrowBufferView ArrowArrayViewGetBytesUnsafe(
...
+      if (inline_size <= NANOARROW_BINARY_VIEW_INLINE_SIZE) {
+        view.data.as_uint8 = value_view.as_uint8 + idx +
+                             sizeof(((union ArrowBinaryViewType*)0)->inlined.size);
+      } else {
+        const int32_t buf_index = bvt.ref.buffer_index;
+        const int32_t nfixed_buf = NANOARROW_BINARY_VIEW_FIXED_BUFFERS;
+        const struct ArrowBufferView variadic_vw =
+            ArrowArrayViewBufferView(array_view, buf_index + nfixed_buf);
+        view.data.as_uint8 = variadic_vw.data.as_uint8 + bvt.ref.offset;
+      }
```
bkietz, line 1121: I think it'd probably be best to add `const struct ArrowBinaryView* as_binary_view;` to `union ArrowBufferViewData`, and extract a function

```c
static struct ArrowBufferView ArrowArrayViewGetBytesFromViewArrayUnsafe(
    const struct ArrowArrayView* array_view, int64_t i) {
  const struct ArrowBinaryView* bv = &array_view->buffer_views[1].data.as_binary_view[i];
  struct ArrowBufferView out{.size_bytes = bv->inlined.size};
  if (bv->inlined.size <= NANOARROW_BINARY_VIEW_INLINE_SIZE) {
    out.data.as_uint8 = bv->inlined.data;
    return out;
  }

  const int32_t buf_index = bv->ref.buf_index + NANOARROW_BINARY_VIEW_FIXED_BUFFERS;
  out.data.data = array_view->array->buffers[buf_index];
  out.data.as_uint8 += bv->ref.offset;
  return out;
}
```
(replacing ArrowArrayViewBufferView).

 Then we can rewrite this as:
```suggestion
      view = ArrowArrayViewGetBytesFromViewArrayUnsafe(string_array, i);
```

and ArrowArrayViewGetStringUnsafe's case as

```c++
    case NANOARROW_TYPE_STRING_VIEW:
    case NANOARROW_TYPE_BINARY_VIEW: {
      struct ArrowBufferView buf_view = ArrowArrayViewGetBytesFromViewArrayUnsafe(string_array, i);
      view.data = buf_view.data.as_char;
      view.size_bytes = buf_view.size_bytes;
      break;
    }
```

bkietz (changes requested):

src/nanoarrow/common/inline_array.h
WillAyd: Ah OK interesting. So right now whenever a variadic buffer's capacity may be exceeded I am starting afresh in a new buffer - should I be packing the existing buffer as full as possible before-hand and potentially splitting the value array across two variadic buffers? Or is that only a consideration when `value.size_bytes > NANOARROW_BINARY_VIEW_BLOCK_SIZE`?

bkietz: Ah, I wrote this before observing that you aren't presizing data buffers https://github.com/apache/arrow-nanoarrow/pull/596#discussion_r1769271911

Since you aren't presizing this isn't a bug

src/nanoarrow/common/inline_array.h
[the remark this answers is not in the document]
WillAyd: OK I think I've got this right. FWIW, maybe we would also add a `ArrowArrayGetBuffer(struct ArrowArray* array, int32_t i)` function to abstract many of the calls that have to reach into `private_data` as well?

src/nanoarrow/common/array_test.cc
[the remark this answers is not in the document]
WillAyd: Does it matter that the string test uses `ArrowArrayAppendString` whereas the binary one uses `ArrowArrayAppendBytes` for the append func? Should I just use `ArrowArrayAppendBytes` in the shared test implementation? Or parametrize it to accept a std::function?

src/nanoarrow/common/inline_array.h
WillAyd: Awesome idea - thanks!

src/nanoarrow/common/inline_array.h
WillAyd: Assuming we want to leave that to a follow up PR at this point, but is the idea that we would just pre-allocate the entire `NANOARROW_BINARY_VIEW_BLOCK_SIZE` bytes? Or is there a more refined algorithm?

src/nanoarrow/common/inline_array.h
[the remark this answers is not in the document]
paleolimbot: I think we have this one!

https://github.com/apache/arrow-nanoarrow/blob/561de01aa4997f6b7900f576da3f302ab149c63a/src/nanoarrow/common/inline_array.h#L40-L49

We do use it a number of internals (but could possibly use it in others!)

src/nanoarrow/common/inline_array.h
paleolimbot: I may be misunderstanding this, but would:

```c
if (value.size_bytes < NANOARROW_BINARY_VIEW_BLOCK_SIZE) {
  NANOARROW_RETURN_NOT_OK(ArrowBufferReserve(variadic_buf, NANOARROW_BINARY_VIEW_BLOCK_SIZE));
}
```

...just after adding a new variadic buffer solve this? (Feel free to ignore if this is off base...I agree that there's no bug here!)

WillAyd: I think if we wanted to reserve the size of the buffer we should do it after adding it with the call to `ArrowArrayAddVariadicBuffers`

bkietz (changes requested):
Just a few more nits, looking good

src/nanoarrow/common/inline_array.h
paleolimbot: The blocking logic is rather specific to our appending, and we should make sure that others can implement their own logic if they want, which might have nothing to do with our block size. (Also, `value.size_bytes` is greater than the block size, that would result the initial reserve being immediately discarded).

src/nanoarrow/common/array_test.cc
```diff
@@ -895,6 +895,144 @@ TEST(ArrayTest, ArrayTestAppendToLargeStringArray) {
...
+  EXPECT_EQ(inline_buffer[6].ref.buffer_index, 1);
+  EXPECT_EQ(inline_buffer[6].ref.offset, 0);
+
+  EXPECT_EQ(memcmp(vbuf1, str1.c_str(), str1.size()), 0);
+  EXPECT_EQ(sizes_buffer[0], str1.size() + filler.size());
+
+  EXPECT_EQ(memcmp(vbuf2, str2.c_str(), str2.size()), 0);
+  EXPECT_EQ(sizes_buffer[1], str2.size());
+
+  // TODO: need to add overload for ViewArrayAsBytes
```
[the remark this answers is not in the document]
WillAyd: https://github.com/apache/arrow-nanoarrow/issues/633

src/nanoarrow/common/array_test.cc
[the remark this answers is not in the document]
paleolimbot: Sorry for missing this comment! Probably we should test both functions to make sure they both work (I believe an `if` statement would work to get that coverage but we've also used `std::function` (for example, in the integration test json tests) and I think either works!

bkietz (approved):

paleolimbot (approved):
This is huge! Thank you!

PR #603, opened 2024-09-05
chore(ci): Fix failing weekly jobs

paleolimbot:
This PR:

- Updates the integration test workflow to not cache a weekly build, which was failing with "no space left on device". It also runs more tests than are strictly necessary at the commit level (and takes slightly longer without the cached weekly build), so this PR makes it part of a weekly run instead.
- Updates the alpine dockerfile to avoid a `desc` dependency, since this fails to install on s390x
- Update two tests to account for a function that returns the value of `errno` (that might not be the same everywhere)

It also updates the weekly jobs for consistency:

- Updates the weekly jobs to always run on the `maint-**` branches
- Update all of them to cancel previous runs of the same workflow (because many of them have long-running jobs)

It doesn't fix IPC writing on big endian (a different kind of failure that I'll punt to another PR).

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -21,7 +21,7 @@ on:
   push:
     branches:
       - main
-      - 'maint-**'
+      - 'maint-*'
   pull_request:
     paths:
       - .github/workflows/integration.yaml
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -32,6 +32,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.repository }}-${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
 jobs:
   integration:
     runs-on: ubuntu-latest
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -21,10 +21,13 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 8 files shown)


docker-compose.yml
```diff
@@ -48,12 +46,10 @@ services:
     command: "/bin/bash /build-docs.sh /nanoarrow"
 
   integration:
-    # Based on an arrow-dev repo image that is amd64 only
-    image: ${REPO}:integration-amd64
+    # Don't cache here (building the image takes about the same amount of time
+    # as downloading on CI)
```
assignUser, line 50: That's probably because the base image is hosted on dockerhub (and pretty big iirc), in my experience things on ghcr.io have much better download characteristics. Though maybe the build is just that fast too ^^

.github/workflows/integration.yaml
```diff
@@ -21,14 +21,21 @@ on:
   push:
     branches:
       - main
+      - 'maint-**'
```
assignUser, line 24:
```suggestion
      - 'maint-*'
```
Probably not an issue but I'd say `**` only makes sense for paths

assignUser (approved):
LGTM

docker-compose.yml
paleolimbot: In any case, I moved it to a weekly job for now since it's a better fit for that suite (and either building or downloading is still >10 minutes) 🙂

PR #605, opened 2024-09-06
chore: More reliable definition of C standard in CMakeLists

paleolimbot:
Some local update on my local machine resulted in nanoarrow not compiling because clang gave a warning for the C11 extensions in the flatcc generated code. I believe this was because the C standard had already been set to C99 at the beginning of the file, whereas the check to update it to C11 only happened after this had already been set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,9 @@ option(NANOARROW_ARROW_STATIC
 
 if(NOT DEFINED CMAKE_C_STANDARD)
   if(NANOARROW_IPC)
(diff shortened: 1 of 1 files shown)


assignUser (approved):
Hm, the value of `CMAKE_C_STANDARD` is used when a new target is  created to initialize the relevant target property. So your existing code should have worked, as afaik the variable is not immutable once set. Could it just have been a case of needing to clean the build tree?

Your solution looks ok otherwise, so if this works in your local case :shrug:

assignUser:
Oh yeah :facepalm: true!

PR #606, opened 2024-09-06
chore(python): Fix type stubs for updated module structure

paleolimbot:
This PR updates the type stubs for the Cython modules to reflect the updated structure. I don't think this mechanism for generating stubs is the best long-term strategy (requires remembering to do so!)...we could also do something like omit the generated stubs from the source tree and generate them only before packaging; however, this would make them not useful for local development (which is one of the primary short-term benefits of including them).


jorisvandenbossche:
Unfortunately a perfect one doesn't exist for now, I think ;)

The `stubtest` didn't catch that the stub files were out of sync with the updated code? (I don't expect stubtest to verify if the stubs are correct, but at least that it matches the set of objects that are present in the imported module)

PR #608, opened 2024-09-15
feat(r): Add bindings to IPC writer

paleolimbot:


r/R/ipc.R
```diff
@@ -107,6 +109,42 @@ read_nanoarrow.connection <- function(x, ..., lazy = FALSE) {
   check_stream_if_requested(reader, lazy)
 }
 
+#' @rdname read_nanoarrow
```
amoeba, line 112: This puts `write_nanoarrow` on `read_nanoarrow`'s page which either needs to be updated or maybe `write_nanoarrow` just needs its own help page. I'm happy to contribute the latter.

r/src/ipc.c
```diff
@@ -112,6 +157,34 @@ static SEXP call_readbin(void* hdata) {
   return R_NilValue;
 }
 
+static SEXP call_writebin(void* hdata) {
+  struct ConnectionInputStreamHandler* data = (struct ConnectionInputStreamHandler*)hdata;
+
+  // Write 16MB chunks
```
amoeba, line 163: where does the 16MB value come from?

r/src/ipc.c
```diff
@@ -159,3 +261,40 @@ SEXP nanoarrow_c_ipc_array_reader_connection(SEXP con) {
...
+  output_stream->release = &release_con_output_stream;
+  output_stream->private_data = (SEXP)con;
+  nanoarrow_preserve_sexp(con);
+
+  SEXP writer_xptr = PROTECT(writer_owning_xptr());
+  struct ArrowIpcWriter* writer = (struct ArrowIpcWriter*)R_ExternalPtrAddr(writer_xptr);
+
+  int code = ArrowIpcWriterInit(writer, output_stream);
+  if (code != NANOARROW_OK) {
+    Rf_error("ArrowIpcWriterInit() failed");
```
amoeba, line 280: Could a more useful error message be produced here? It looks like code is an errno-style error so it seems like it'd be possible.

amoeba (approved):
Looks good to me. Left some notes, feel free to ignore or open separate issues. I'll note that I couldn't do a thorough job reviewing the C parts of this.

r/R/ipc.R
```diff
@@ -35,6 +35,8 @@
 #' @param x A `raw()` vector, connection, or file path from which to read
 #'   binary data. Common extensions indicating compression (.gz, .bz2, .zip)
 #'   are automatically uncompressed.
+#' @param data An object to write as an Arrow IPC stream, converted using
```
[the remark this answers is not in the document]
paleolimbot: I added a note about `data.frame`, although anything that works with `as_nanoarrow_array_stream()` should work out of the box (e.g., many arrow package objects). We should probably come up with some prose about those types of objects ("array streamable"?)...this is also what ADBC uses for several types of input.

r/R/ipc.R
```diff
@@ -24,8 +24,8 @@
 #' from files or URLs and is essentially a high performance equivalent of
 #' a CSV file that does a better job maintaining types.
 #'
-#' The nanoarrow package does not currently have the ability to write serialized
-#' IPC data: use [arrow::write_ipc_stream()] to write data from R, or use
+#' The nanoarrow package implements an IPC writer; however, you can also
```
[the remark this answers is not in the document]
paleolimbot: I'll make a PR into Arrow after we release!

r/src/ipc.c
paleolimbot: I added a note about this...it's somewhere between a big value (to reduce the number of R calls) and a small value (to minimize the copying/ensure there's an R call every once in a while to catch an interrupt).

r/src/ipc.c
paleolimbot: I added the errno value! Pretty much all that can happen here is a failure to allocate (rare, since it's a few bytes). I never quite got to adding a macro like `ERROR_NOT_OK()` to help with this.

r/R/ipc.R
paleolimbot: I added a basic level of inclusion on the read page but feel free to contribute something better!

r/R/ipc.R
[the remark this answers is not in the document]
amoeba: Sounds good.

r/R/ipc.R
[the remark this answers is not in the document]
amoeba: Interesting. Documenting the actual set of types on `?as_nanoarrow_array_stream` might be good then and using the phrase "Arrow-streamable" here might be a good start.

PR #609, opened 2024-09-16
fix: Ensure footer test passes on big endian

paleolimbot:
This PR fixes the verification run failure for big endian, which is currently failing on the `NanoarrowIpcNanoarrowFooterRoundtrip` tests. Checked with:

```bash
export NANOARROW_ARCH=s390x
docker compose run --rm verify
```

### DIFF
--- a/src/nanoarrow/ipc/decoder_test.cc
+++ b/src/nanoarrow/ipc/decoder_test.cc
@@ -1012,11 +1012,6 @@ TEST_P(ArrowSchemaParameterizedTestFixture, NanoarrowIpcNanoarrowFooterRoundtrip
   memcpy(&roundtripped_block, decoder->footer->record_batch_blocks.data,
          sizeof(roundtripped_block));
 
-#ifdef __BIG_ENDIAN__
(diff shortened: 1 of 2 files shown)


src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -984,7 +984,13 @@ TEST_P(ArrowSchemaParameterizedTestFixture, NanoarrowIpcNanoarrowFooterRoundtrip
   EXPECT_EQ(
       ArrowIpcEncoderFinalizeBuffer(encoder.get(), /*encapsulate=*/false, buffer.get()),
       NANOARROW_OK);
+
+#ifdef __BIG_ENDIAN__
+  uint32_t footer_size_le = bswap32(static_cast<int32_t>(buffer->size_bytes));
```
bkietz, line 989: I believe byteswap functions take only unsigned integers

```suggestion
  uint32_t footer_size_le = bswap32(static_cast<uint32_t>(buffer->size_bytes));
```

bkietz (approved):
LGTM, one nit

PR #614, opened 2024-09-17
fix: Ensure CMake linking against built/installed nanoarrow works for all components

### LINKED ISSUE #407: Make extensions available as CMake components

paleolimbot:
Closes #407.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,7 +148,6 @@ endif()
 
 # Add the nanoarrow library target
 add_library(nanoarrow ${NANOARROW_BUILD_SOURCES})
-add_library(nanoarrow::nanoarrow ALIAS nanoarrow)
 
 target_include_directories(nanoarrow
                            PUBLIC $<BUILD_INTERFACE:${NANOARROW_BUILD_INCLUDE_DIR}>
@@ -220,8 +219,8 @@ if(NANOARROW_IPC)
                                     $<INSTALL_INTERFACE:include>)
 
   install(TARGETS nanoarrow_ipc DESTINATION lib)
-  install(FILES src/nanoarrow/nanoarrow_ipc.h src/nanoarrow/ipc/flatcc_generated.h
-          DESTINATION include/nanoarrow)
+  install(FILES src/nanoarrow/nanoarrow_ipc.h src/nanoarrow/nanoarrow_ipc.hpp
+                src/nanoarrow/ipc/flatcc_generated.h DESTINATION include/nanoarrow)
 endif()
 
 if(NANOARROW_IPC AND (NANOARROW_BUILD_INTEGRATION_TESTS OR NANOARROW_BUILD_TESTS))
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -337,9 +338,10 @@ if(NANOARROW_TESTING
                              PUBLIC $<BUILD_INTERFACE:${NANOARROW_BUILD_INCLUDE_DIR}>
                                     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/src>
                                     $<INSTALL_INTERFACE:include>)
-  target_link_libraries(nanoarrow_testing PRIVATE nlohmann_json::nlohmann_json nanoarrow)
-  target_link_libraries(nanoarrow_testing PUBLIC nanoarrow)
+  target_link_libraries(nanoarrow_testing PRIVATE nlohmann_json::nlohmann_json)
+  target_link_libraries(nanoarrow_testing PUBLIC nanoarrow nanoarrow_coverage_config)
```
assignUser, line 342: nit: you can combine these into one invocation but if you prefer it this way that's also fine.

examples/cmake-scenarios/CMakeLists.txt
```diff
@@ -29,21 +29,31 @@ option(FIND_NANOARROW "Find an existing nanoarrow" ON)
...
-                       GIT_SHALLOW TRUE)
-  fetchcontent_makeavailable(nanoarrow_example_cmake_minimal)
+  # We need all these components for our test
+  set(NANOARROW_DEVICE ON)
+  set(NANOARROW_IPC ON)
+  set(NANOARROW_TESTING ON)
+
+  # Using SOURCE_DIR here such that CI accurately reflects the state of the
+  # source; however, using GIT_TAG or URL of a pinned version is a more realistic
+  # user pattern.
```
assignUser, line 49: :+1:

examples/cmake-scenarios/build.sh
```diff
@@ -20,12 +20,18 @@
 set -exuo pipefail
 
 # Build nanoarrow statically.
-cmake -S ../.. -B scratch/nanoarrow_build_static/ -DCMAKE_INSTALL_PREFIX=scratch/nanoarrow_install_static/
+cmake -S ../.. -B scratch/nanoarrow_build_static/ \
+    -DCMAKE_INSTALL_PREFIX=scratch/nanoarrow_install_static/ \
+    -DNANOARROW_IPC=ON -DNANOARROW_DEVICE=ON -DNANOARROW_TESTING=ON
```
assignUser, line 25: Do the cmake presets need updating as well?

assignUser (approved):
:+1:

CMakeLists.txt
paleolimbot: Way cleaner...thanks!

examples/cmake-scenarios/build.sh
paleolimbot: We don't currently use presets for anything (although we do define a few and possibly should make better use of them!).

PR #626, opened 2024-09-19
fix: Improve validation of offset buffers for sliced arrays

paleolimbot:
This PR updates validation of offset buffers in string/binary, large string/binary, list, and large list types such that portions of the buffers not strictly required by the slice referenced by `offset` and `length` are not accessed/validated. This came up in the IPC integration tests because C# exports full buffers rather than the portions of buffers strictly required.

This highlights how our test suite would benefit from helpers to reduce repetition (although I'd prefer to handle that at a later time if possible).

### DIFF
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -998,6 +998,11 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
         }
 
         last_offset = array_view->buffer_views[1].data.as_int32[offset_plus_length];
+        if (last_offset < 0) {
+          ArrowErrorSet(error, "Expected last offset >= 0 but found %" PRId64,
+                        last_offset);
+          return EINVAL;
+        }
 
         // If the data buffer size is unknown, assign it; otherwise, check it
         if (array_view->buffer_views[2].size_bytes == -1) {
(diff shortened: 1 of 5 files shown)


src/nanoarrow/common/array.c
```diff
@@ -1249,13 +1249,21 @@ static int ArrowArrayViewValidateFull(struct ArrowArrayView* array_view,
...
+              array_view->buffer_views[i].data.as_int32 + array_view->offset;
+          offset_minimal.size_bytes = (array_view->length + 1) * sizeof(int32_t);
+          NANOARROW_RETURN_NOT_OK(ArrowAssertIncreasingInt32(offset_minimal, error));
+        } else if (array_view->length > 0) {
+          struct ArrowBufferView offset_minimal;
+          offset_minimal.data.as_int64 =
+              array_view->buffer_views[i].data.as_int64 + array_view->offset;
+          offset_minimal.size_bytes = (array_view->length + 1) * sizeof(int64_t);
+          NANOARROW_RETURN_NOT_OK(ArrowAssertIncreasingInt64(offset_minimal, error));
         }
```
bkietz, line 1267: Nit:
```suggestion
        if (array_view->length == 0) {
          continue;
        }
        if (array_view->layout.element_size_bits[i] == 32) {
          struct ArrowBufferView offset_minimal;
          offset_minimal.data.as_int32 =
              array_view->buffer_views[i].data.as_int32 + array_view->offset;
          offset_minimal.size_bytes = (array_view->length + 1) * sizeof(int32_t);
          NANOARROW_RETURN_NOT_OK(ArrowAssertIncreasingInt32(offset_minimal, error));
        } else {
          struct ArrowBufferView offset_minimal;
          offset_minimal.data.as_int64 =
              array_view->buffer_views[i].data.as_int64 + array_view->offset;
          offset_minimal.size_bytes = (array_view->length + 1) * sizeof(int64_t);
          NANOARROW_RETURN_NOT_OK(ArrowAssertIncreasingInt64(offset_minimal, error));
        }
```

bkietz (changes requested):

Issue #466, opened 2024-05-14
python: Schema repr should not get truncated


jorisvandenbossche:
Looking at this example, I think it might be reasonable to truncate the schema repr when it is embedded in the Array repr, but then when inspecting the schema itself I would expected to see the full repr (or at least the truncation can be more relaxed, and show more by default):

```
In [29]: url = "https://github.com/apache/arrow-experiments/raw/main/data/arrow-commits/arrow-commits.arrows"
    ...: array = na.ArrayStream.from_url(url).read_all()

In [30]: array
Out[30]: 
nanoarrow.Array<non-nullable struct<commit: string, time: timestamp('...>[15487]
{'commit': '49cdb0fe4e98fda19031c864a18e6156c6edbf3c', 'time': datetime.datet...
{'commit': '1d966e98e41ce817d1f8c5159c0b9caa4de75816', 'time': datetime.datet...
{'commit': '96f26a89bd73997f7532643cdb27d04b70971530', 'time': datetime.datet...
{'commit': 'ee1a8c39a55f3543a82fed900dadca791f6e9f88', 'time': datetime.datet...
{'commit': '3d467ac7bfae03cf2db09807054c5672e1959aec', 'time': datetime.datet...
{'commit': 'ef6ea6beed071ed070daf03508f4c14b4072d6f2', 'time': datetime.datet...
{'commit': '53e0c745ad491af98a5bf18b67541b12d7790beb', 'time': datetime.datet...
{'commit': '3ba6d286caad328b8572a3b9228045da8c8d2043', 'time': datetime.datet...
{'commit': '4ce9a5edd2710fb8bf0c642fd0e3863b01c2ea20', 'time': datetime.datet...
{'commit': '2445975162905bd8d9a42ffc9cd0daa0e19d3251', 'time': datetime.datet...
...and 15477 more items

In [31]: array.schema
Out[31]: <Schema> non-nullable struct<commit: string, time: timestamp('us', 'UTC'), fi...
```

paleolimbot:
Good point! Maybe include the output of `schema.fields` by default (perhaps truncated to a reasonable number of lines?)

```python
import nanoarrow as na

url = "https://github.com/apache/arrow-experiments/raw/main/data/arrow-commits/arrow-commits.arrows"
schema = na.ArrayStream.from_url(url).schema

schema.fields
#> [<Schema> 'commit': string,
#>  <Schema> 'time': timestamp('us', 'UTC'),
[...]

PR #628, opened 2024-09-20
fix: python schema repr does not truncate output

danepitkin:
Fix https://github.com/apache/arrow-nanoarrow/issues/466

```
>>> import nanoarrow as na
>>> url = "https://github.com/apache/arrow-experiments/raw/main/data/arrow-commits/arrow-commits.arrows"
>>> schema = na.ArrayStream.from_url(url).schema
>>> schema
<Schema> non-nullable struct<commit: string, time: timestamp('us', 'UTC'), files: int32, merge: bool, message: string>
```

### DIFF
--- a/python/src/nanoarrow/_repr_utils.py
+++ b/python/src/nanoarrow/_repr_utils.py
@@ -27,12 +27,15 @@ def make_class_label(obj, module=None):
 
 
 def c_schema_to_string(obj, max_char_width=80):
-    max_char_width = max(max_char_width, 10)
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
Thanks!

I do think there's a point where printing out everything is more harmful than helpful (e.g., if there are thousands of columns); however, this is a substantial improvement over what is here now!

danepitkin:
Thanks for the review! I suppose we could have a really high limit like 1000 chars in that case. Although, if a user prints out a schema, I assume they are in some sort of interactive or debug mode? In that use case, it might not make sense to limit the schema column output at all. 🤷

PR #629, opened 2024-09-20
fix: Accommodate IPC messages without continuation bytes

bkietz:
Before arrow 0.15, encapsulated messages started with just the length instead of with a continuation token


paleolimbot (approved):
Thanks! It seems it hasn't caused problems for other implementations to allow this. It seems like better practice to enforce the continuation eventually (but the near-term benefit of getting all integration tests passing without complaint is far better!)

PR #630, opened 2024-09-20
fix: ignore empty (but present) union validity bitmaps from before 1.0

bkietz:
For metadata versions before V5, ignore the extra validity bitmap. If the bitmap is non-empty, we raise an error (which is the same bail used in arrow-C++ and -java).


paleolimbot (approved):
Thanks! Again, we should maybe eventually remove this complexity (but getting the integration tests passing is a better near-term goal).

Issue #634, opened 2024-09-25
ArrowArrayViewSetArray fails for BinaryView array with no variadic buffers


WillAyd:
Testing out the new implementation with data like:

```python
import pyarrow as pa
arr = pa.array(["foo", "baz"], type=pa.string_view())
```

an extension that calls `ArrowArrayViewSetArray` will throw an error of `Expected array with 2 buffer(s) but found 3`

When debugging, it appears that the Arrow C bridge may be exporting the variadic length buffer, even though it does not exist. This can be confirmed from gdb:

```
Thread 1 "python" hit Breakpoint 2, ArrowArrayViewSetArrayInternal (array_view=0x7fffffffaf60, array=0x7ffff0e862d0, error=0x7fffffffb020) at /home/willayd/clones/nanoarrow_mre/build/_deps/nanoarrow-project-src/src/nanoarrow/common/array.c:764
764	    const int64_t n_buffers = array->n_buffers;
(gdb) p ((union ArrowBinaryView*)array->buffers[1])[0]
$2 = {inlined = {size = 3, data = "foo\000\000\000\000\000\000\000\000"}, ref = {size = 3, prefix = "foo", buffer_index = 0, offset = 0}, 
  alignment_dummy = 31366206292230147}
(gdb) p ((union ArrowBinaryView*)array->buffers[1])[1]
$3 = {inlined = {size = 3, data = "bar\000\000\000\000\000\000\000\000"}, ref = {size = 3, prefix = "bar", buffer_index = 0, offset = 0}, 
  alignment_dummy = 32195220879704067}
(gdb) p *array
$4 = {length = 2, null_count = 0, offset = 0, n_buffers = 3, n_children = 0, buffers = 0x7ffff0e58b00, children = 0x8, dictionary = 0x0, 
  release = 0x7ffff245afa0 <polars_arrow::ffi::array::c_release_array>, private_data = 0x7ffff0e86280}
(gdb) p array->buffers[2]
$5 = (const void *) 0x8
```

is that extra buffer supposed to be part of the C data interface, or is it a bug in the Arrow bridge?

paleolimbot:
Ah, I think it's supposed to be there: https://arrow.apache.org/docs/format/CDataInterface.html#c.ArrowArray.n_buffers

WillAyd:
OK great - will take a look

PR #635, opened 2024-09-25
fix; Properly ingest Binary View types without variadic buffers

WillAyd:
closes https://github.com/apache/arrow-nanoarrow/issues/634

### DIFF
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -115,7 +115,7 @@ static ArrowErrorCode ArrowArraySetStorageType(struct ArrowArray* array,
       break;
     case NANOARROW_TYPE_BINARY_VIEW:
     case NANOARROW_TYPE_STRING_VIEW:
-      array->n_buffers = NANOARROW_BINARY_VIEW_FIXED_BUFFERS;
+      array->n_buffers = NANOARROW_BINARY_VIEW_FIXED_BUFFERS + 1;
       break;
     case NANOARROW_TYPE_STRING:
     case NANOARROW_TYPE_LARGE_STRING:
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -765,11 +765,9 @@ static int ArrowArrayViewSetArrayInternal(struct ArrowArrayView* array_view,
     const int32_t nfixed_buf = NANOARROW_BINARY_VIEW_FIXED_BUFFERS;
 
     const int32_t nvariadic_buf = (int32_t)(n_buffers - nfixed_buf - 1);
-    if (nvariadic_buf > 0) {
-      array_view->n_variadic_buffers = nvariadic_buf;
-      buffers_required += nvariadic_buf + 1;
-      array_view->variadic_buffer_sizes = (int64_t*)array->buffers[n_buffers - 1];
-    }
+    array_view->n_variadic_buffers = nvariadic_buf;
+    buffers_required += nvariadic_buf + 1;
+    array_view->variadic_buffer_sizes = (int64_t*)array->buffers[n_buffers - 1];
   }
 
   if (buffers_required != array->n_buffers) {
--- a/src/nanoarrow/common/array_test.cc
+++ b/src/nanoarrow/common/array_test.cc
@@ -3343,7 +3343,50 @@ TEST(ArrayViewTest, ArrayViewTestGetString) {
   TestGetFromBinary<FixedSizeBinaryBuilder>(fixed_size_builder);
 }
 
-template <typename BuilderClass>
+static const void* GetBinaryData(const struct ArrowBufferView* bv) {
+  return bv->data.data;
+}
+
+static const void* GetBinaryData(const struct ArrowStringView* sv) { return sv->data; }
+
+template <typename BuilderClass, typename ValueT,
+          ValueT (*GetFunc)(const struct ArrowArrayView*, int64_t)>
+void TestGetFromInlinedBinaryView(BuilderClass& builder) {
+  struct ArrowArray array;
+  struct ArrowSchema schema;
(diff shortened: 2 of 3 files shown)


src/nanoarrow/common/array_test.cc
```diff
@@ -3343,7 +3343,50 @@ TEST(ArrayViewTest, ArrayViewTestGetString) {
   TestGetFromBinary<FixedSizeBinaryBuilder>(fixed_size_builder);
 }
 
-template <typename BuilderClass>
+static const void* GetBinaryData(const struct ArrowBufferView* bv) {
```
WillAyd, line 3346: Open to different ideas on how to best template these tests; if constexpr would be really nice here!

src/nanoarrow/common/array_test.cc
```diff
@@ -3380,29 +3423,17 @@ void TestGetFromBinaryView(BuilderClass& builder) {
   EXPECT_EQ(ArrowArrayViewIsNull(&array_view, 2), 1);
   EXPECT_EQ(ArrowArrayViewIsNull(&array_view, 3), 0);
 
-  auto string_view = ArrowArrayViewGetStringUnsafe(&array_view, 3);
```
WillAyd, line 3383: A bit of a refactor mixed in, but I didn't want to keep repeating this pattern of testing both String/Byte get functions for all Ts

WillAyd: Also generally curious - why do we keep the test suite tied to C++11? Is that to let downstream libraries run the tests without imposing a more recent standard?

src/nanoarrow/common/array_test.cc
paleolimbot: Can you accept `std::function` or a function pointer as an argument? Like:

https://github.com/apache/arrow-nanoarrow/blob/53c9f8f69743d633b831b735de1402cca38713ad/src/nanoarrow/testing/testing_test.cc#L58-L63

```c
template <typename BuilderClass>
void TestGetFromInlinedBinaryView(BuilderClass& builder, enum ArrowType type_id, std::string (*GetBinaryDataAsString)(ArrowArrayView*, int64_t)) { ... }

```

paleolimbot: It's a personal preference, but I usually find that at the point I get to `if constexpr` I've missed a better way to abstract things. Some people love it though!

paleolimbot (commented):
Thanks for the fix!

Does this mean we are exporting an incorrect value for `n_buffers` when we build arrays?

WillAyd:
Good point - yes the builder starts with n_array = 2. I'll take a look at that

src/nanoarrow/common/array.c
```diff
@@ -115,7 +115,7 @@ static ArrowErrorCode ArrowArraySetStorageType(struct ArrowArray* array,
       break;
     case NANOARROW_TYPE_BINARY_VIEW:
     case NANOARROW_TYPE_STRING_VIEW:
-      array->n_buffers = NANOARROW_BINARY_VIEW_FIXED_BUFFERS;
+      array->n_buffers = NANOARROW_BINARY_VIEW_FIXED_BUFFERS + 1;
```
WillAyd, line 118: While this "works" I'm not sure if we still care to maintain this distinction of `NANOARROW_BINARY_FIXED_VIEW_BUFFERS` being 2 when all other types really have 3 fixed buffers. I suppose the last buffer isn't fixed, in the sense that it can be moved. So semantically maybe there is some value in having this still? But we could probably reconsolidate some code if we go back to just using `NANOARROW_MAX_FIXED_BUFFERS` like the other types

src/nanoarrow/common/array_test.cc
```diff
@@ -3343,8 +3387,49 @@ TEST(ArrayViewTest, ArrayViewTestGetString) {
   TestGetFromBinary<FixedSizeBinaryBuilder>(fixed_size_builder);
 }
 
-template <typename BuilderClass>
-void TestGetFromBinaryView(BuilderClass& builder) {
+template <typename BuilderClass, typename ValueT>
+void TestGetFromInlinedBinaryView(
```
WillAyd, line 3391: There is some inconsistency between the signature of this test and `TestAppendToInlinedDataViewArray` preceding it. Whereas here we are accepting `GetValueFunc` and `GetValueDataFunc` as arguments, the aforementioned test still takes a template argument.

Should I just move them both to using a function argument instead of a template argument?

src/nanoarrow/common/inline_array.h
```diff
@@ -525,6 +525,7 @@ static inline ArrowErrorCode ArrowArrayAddVariadicBuffers(struct ArrowArray* arr
     private_data->variadic_buffer_sizes[i] = 0;
   }
   private_data->n_variadic_buffers = n_bufs_needed;
+  array->n_buffers = NANOARROW_BINARY_VIEW_FIXED_BUFFERS + 1 + n_bufs_needed;
```
WillAyd, line 528: For clarity, we might want to rename `n_bufs_needed` to `nvariadic_bufs_needed`

paleolimbot: That seems reasonable to me if you're up for it!

src/nanoarrow/common/array_test.cc
paleolimbot: If you're up for it, having them both take a function parameter I think makes sense!

src/nanoarrow/common/array.c
paleolimbot: I think it's still good to keep (you could leave a note along those lines as a comment if you'd like next to the definition)

paleolimbot (approved):
Thank you!

PR #636, opened 2024-09-27
feat(r): Implement string view support in R bindings

paleolimbot:
This PR adds support for string view and binary view types to the R bindings. As a side effect of this, conversion of character vectors to Arrow types is now simpler (just goes through nanoarrow C's array builder) and supports more types (e.g., the arrow package is no longer required to create large_string, large_binary, or fixed_size_binary).

``` r
library(nanoarrow)

long_strings <- rep(strrep(letters, 100), 100)

(array <- as_nanoarrow_array(long_strings, schema = na_string_view()))
#> <nanoarrow_array string_view[2600]>
#>  $ length    : int 2600
#>  $ null_count: int 0
#>  $ offset    : int 0
#>  $ buffers   :List of 11
#>   ..$ :<nanoarrow_buffer validity<bool>[0][0 b]> ``
#>   ..$ :<nanoarrow_buffer unknown<string_view>[2600][41600 b]>`
#>   ..$ :<nanoarrow_buffer data<string>[32700 b]> `aaaaaaaaaaaaaaaaaaaaaaaaaaa...`
#>   ..$ :<nanoarrow_buffer data<string>[32700 b]> `ppppppppppppppppppppppppppp...`
#>   ..$ :<nanoarrow_buffer data<string>[32700 b]> `eeeeeeeeeeeeeeeeeeeeeeeeeee...`
#>   ..$ :<nanoarrow_buffer data<string>[32700 b]> `ttttttttttttttttttttttttttt...`
#>   ..$ :<nanoarrow_buffer data<string>[32700 b]> `iiiiiiiiiiiiiiiiiiiiiiiiiii...`
#>   ..$ :<nanoarrow_buffer data<string>[32700 b]> `xxxxxxxxxxxxxxxxxxxxxxxxxxx...`
#>   ..$ :<nanoarrow_buffer data<string>[32700 b]> `mmmmmmmmmmmmmmmmmmmmmmmmmmm...`
#>   ..$ :<nanoarrow_buffer data<string>[31100 b]> `bbbbbbbbbbbbbbbbbbbbbbbbbbb...`
#>   ..$ :<nanoarrow_buffer data<int64>[8][64 b]> `32700 32700 32700 32700 3270...`
#>  $ dictionary: NULL
#>  $ children  : list()

identical(convert_array(array), long_strings)
#> [1] TRUE
```

<sup>Created on 2024-09-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


r/R/type.R
```diff
@@ -460,7 +472,10 @@ NANOARROW_TYPE <- list(
   LARGE_STRING = 35L,
   LARGE_BINARY = 36L,
   LARGE_LIST = 37L,
-  INTERVAL_MONTH_DAY_NANO = 38L
+  INTERVAL_MONTH_DAY_NANO = 38L,
+  RUN_END_ENCODED = 39L,
```
amoeba, line 476: I didn't expect to see the additioin of `RUN_END_ENCODED` here. Just noted while I do my review.

r/tests/testthat/test-convert-array.R
```diff
@@ -893,6 +907,22 @@ test_that("convert to vector works for blob::blob()", {
...
+test_that("convert to vector works for binary_view -> blob::blob()", {
+  skip_if_not_installed("blob")
+
+  array <- as_nanoarrow_array(letters, schema = na_binary_view())
+  expect_identical(
+    convert_array(array, blob::blob()),
+    blob::as_blob(lapply(letters, charToRaw))
+  )
+
+  array_with_nulls <- as_nanoarrow_array(c(letters, NA), schema = na_binary_view())
```
amoeba, line 919: Should this also work? It doesn't for me:

```r
# Test all nulls
> array_with_nulls <- as_nanoarrow_array(c(rep(NA, 10)), schema = na_binary_view())
Error: NotImplemented: binary_view
```

amoeba: ah, so this is similar to other calls, like,

```r
> as_nanoarrow_array(-1, schema = na_binary_view())
Error: NotImplemented: binary_view
```

which is an error bubbled up from the arrow package:

```
> traceback()
10: vec_to_Array(x, type)
9: as_arrow_array.default(x, type = arrow::as_data_type(schema))
8: arrow::as_arrow_array(x, type = arrow::as_data_type(schema)) at array.R#43
7: as_nanoarrow_array(arrow::as_arrow_array(x, type = arrow::as_data_type(schema))) at as-array.R#40
6: as_nanoarrow_array.default(x, schema = schema, .from_c = TRUE) at array.R#43
5: as_nanoarrow_array(x, schema = schema, .from_c = TRUE) at as-array.R#428
4: as_nanoarrow_array_from_c(c(NA, NA, NA, NA, NA, NA, NA, NA, NA,
   NA), <pointer: 0x600002e68280>)
3: .Call(nanoarrow_c_as_array_default, x, schema) at as-array.R#60
2: as_nanoarrow_array.default(rep(NA, 10), schema = na_string_view()) at array.R#43
1: as_nanoarrow_array(rep(NA, 10), schema = na_string_view())
>
```

So maybe this isn't something to change?

r/tests/testthat/test-as-array.R
```diff
@@ -322,6 +320,30 @@ test_that("as_nanoarrow_array() works for character() -> na_large_string()", {
...
 })
 
+test_that("as_nanoarrow_array() works for character() -> na_string_view()", {
+  # Without nulls
+  array <- as_nanoarrow_array(letters, schema = na_string_view())
+  expect_identical(infer_nanoarrow_schema(array)$format, "vu")
+  expect_identical(as.raw(array$buffers[[1]]), raw())
+  expect_identical(array$offset, 0L)
+  expect_identical(array$null_count, 0L)
+  # All these strings are shorter than four characters and thus are all inlined
```
amoeba, line 330: would it be worthwhile to test non-lineable input here?

r/src/as_array.c
```diff
@@ -428,83 +403,57 @@ static void as_array_data_frame(SEXP x_sexp, struct ArrowArray* array, SEXP sche
...
-      Rf_error("ArrowBitmapReserve() failed");
-    }
-
-    for (int64_t i = 0; i < len; i++) {
-      uint8_t is_valid = VECTOR_ELT(x_sexp, i) != R_NilValue;
-      ArrowBitmapAppendUnsafe(&bitmap, is_valid, 1);
+    if (item == R_NilValue) {
+      result = ArrowArrayAppendNull(array, 1);
+      if (result != NANOARROW_OK) {
+        Rf_error("ArrowArrayAppendString() failed");
```
amoeba, line 438:
```suggestion
        Rf_error("ArrowArrayAppendNull() failed");
```

amoeba (approved):
LGTM.

r/tests/testthat/test-convert-array.R
paleolimbot: Currently we fall back on the Arrow package if we can't convert something. I think that decimal types (create/consume) and `list_of()` (create) are the only two places where we still fall back in this way. In the next release or two I think we should be able to remove those fallbacks (which cause confusion in a number of ways!).

amoeba: sounds good

r/tests/testthat/test-as-array.R
amoeba: sorry, meant non-inlineable. hope that makes sense.

paleolimbot: Good call!

r/R/type.R
paleolimbot: Good catch! I missed adding that when we added a basic level of support in the C library (and added it here to make the incrementing of the numbers make a little more sense!)

amoeba: 👌

PR #637, opened 2024-09-27
feat(python): Add StringView and BinaryView IO to Python bindings

paleolimbot:
This PR implements StringView support in the Python bindings. It is a thin wrapper around the C functions added, although we should perhaps abstract some of the buffer info calculation into the C library since I had to work around that in the R bindings as well.

```python
import nanoarrow as na

array = na.Array(["abc", "def", None, "longer than 12 bytes"], na.string_view())
array
#> nanoarrow.Array<string_view>[4]
#> 'abc'
#> 'def'
#> None
#> 'longer than 12 bytes'
array.buffers
#> (nanoarrow.c_buffer.CBufferView(bool[1 b] 11010000),
#>  nanoarrow.c_buffer.CBufferView(string_view[64 b] b'\x03\x00\x00\x00abc\x00\x00\x00\x00\x00\x00\x00\x00\x00'...),
#>  nanoarrow.c_buffer.CBufferView(string[20 b] b'longer than 12 bytes'),
#>  nanoarrow.c_buffer.CBufferView(int64[8 b] 20))
```


python/src/nanoarrow/_array.pyx
```diff
@@ -189,13 +196,48 @@ cdef class CArrayView:
...
         if i < 0 or i >= self.n_buffers:
             raise IndexError(f"{i} out of range [0, {self.n_buffers}]")
 
-        buffer_type = self._ptr.layout.buffer_type[i]
+        if (
+            _types.is_data_view(self._ptr.storage_type)
+            and i == (2 + self._ptr.n_variadic_buffers)
+        ):
+            return (
+                NANOARROW_BUFFER_TYPE_DATA,
```
WillAyd, line 213: Checking the type of the variadic buffer is actually missing from the C++ test; we might want to go back and add this there too

python/src/nanoarrow/_array.pyx
```diff
@@ -249,6 +294,24 @@ cdef class CArrayView:
 
         return dictionary
 
+    def _iter_bytes(self, int64_t offset, int64_t length):
```
WillAyd, line 297:
```suggestion
    def _iter_bytes(self, int64_t offset, int64_t length) -> bytes | None:
```

There _might_ be some performance benefit from Cython to declaring the return type here, especially since almost everything else looks to be typed.

python/src/nanoarrow/_array.pyx
```diff
@@ -249,6 +294,24 @@ cdef class CArrayView:
...
+        cdef ArrowBufferView item_view
+        for i in range(offset, length):
+            if ArrowArrayViewIsNull(self._ptr, i):
+                yield None
+            else:
+                item_view = ArrowArrayViewGetBytesUnsafe(self._ptr, i)
+                yield PyBytes_FromStringAndSize(item_view.data.as_char, item_view.size_bytes)
+
+    def _iter_str(self, int64_t offset, int64_t length):
```
WillAyd, line 306:
```suggestion
    def _iter_str(self, int64_t offset, int64_t length) -> str | None:
```

WillAyd (commented):
Nice work! I only have really minor comments

python/src/nanoarrow/_array.pyx
paleolimbot: I'll follow this up with a C library PR!

WillAyd (approved):

PR #638, opened 2024-09-30
feat: Add ArrowArrayView accessors to inspect buffer properties

paleolimbot:
This PR abstracts accessors for the buffer_view, buffer type, buffer data type, and element bit width for the `ArrowArrayView`. Before adding string/binary view support, this was done by directly accessing the `layout` and `buffer_view` members; however, this required special-casing + some duplicated code in the string view in the R/Python bindings.

This PR also removes the dependence on the `ArrowArrayView::array` member, since this member is optional (i.e., the data backing an `ArrowArrayView` need not be related to an actual `ArrowArray`).

### DIFF
--- a/src/nanoarrow/common/inline_types.h
+++ b/src/nanoarrow/common/inline_types.h
@@ -623,7 +623,8 @@ enum ArrowBufferType {
   NANOARROW_BUFFER_TYPE_UNION_OFFSET,
   NANOARROW_BUFFER_TYPE_DATA_OFFSET,
   NANOARROW_BUFFER_TYPE_DATA,
-  NANOARROW_BUFFER_TYPE_DATA_VIEW
+  NANOARROW_BUFFER_TYPE_VARIADIC_DATA,
+  NANOARROW_BUFFER_TYPE_VARIADIC_SIZE
 };
 
 /// \brief The maximum number of fixed buffers in an ArrowArrayView or ArrowLayout
@@ -815,6 +816,9 @@ struct ArrowArrayView {
 
   /// \brief Number of variadic buffers
   int32_t n_variadic_buffers;
+
+  /// \brief Pointers to variadic buffers of binary/string_view arrays
+  const void** variadic_buffers;
 
   /// \brief Size of each variadic buffer
   int64_t* variadic_buffer_sizes;
--- a/src/nanoarrow/nanoarrow.h
+++ b/src/nanoarrow/nanoarrow.h
@@ -1053,6 +1053,48 @@ ArrowErrorCode ArrowArrayViewSetArrayMinimal(struct ArrowArrayView* array_view,
                                              const struct ArrowArray* array,
                                              struct ArrowError* error);
 
+/// \brief Get the number of buffers
+///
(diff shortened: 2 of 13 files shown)


src/nanoarrow/common/inline_types.h
```diff
@@ -623,7 +623,8 @@ enum ArrowBufferType {
   NANOARROW_BUFFER_TYPE_UNION_OFFSET,
   NANOARROW_BUFFER_TYPE_DATA_OFFSET,
   NANOARROW_BUFFER_TYPE_DATA,
-  NANOARROW_BUFFER_TYPE_DATA_VIEW
+  NANOARROW_BUFFER_TYPE_VARIADIC_DATA,
```
WillAyd, line 626: If we are going to introduce new members like this should we think about renaming the existing member to `NANOARROW_BUFFER_TYPE_FIXED_DATA`? I think the distinction can get easily lost

src/nanoarrow/nanoarrow.h
```diff
@@ -1053,6 +1053,48 @@ ArrowErrorCode ArrowArrayViewSetArrayMinimal(struct ArrowArrayView* array_view,
...
+/// the data type of each buffer may be different between two arrays of the same type
+/// depending on the number of variadic buffers.
+static inline enum ArrowType ArrowArrayViewGetBufferDataType(
+    struct ArrowArrayView* array_view, int64_t i);
+
+/// \brief Get the element size (in bits) of a specific buffer in an ArrowArrayView
+///
+/// In may cases this can also be obtained from the ArrowLayout member of the
+/// ArrowArrayView or ArrowSchemaView; however, for binary view and string view types,
+/// the element width of each buffer may be different between two arrays of the same type
```
WillAyd, line 1093: Is the idea here to be forward-looking and support other non-char variadic buffers?

WillAyd (commented):
haven't looked in detail at the implementation just a few thoughts from the API

src/nanoarrow/common/inline_types.h
paleolimbot: That is definitely a better name, although a bit pedantic if you're not specifically dealing with view arrays (and most of the time, we're not!). There is room to revisit that name in the future...the buffer type is helpful for things like the R and Python bindings that are truly generic wrappers, but for the most part these names don't show up in user code (that I've seen).

(It could also be argued that the data buffer of a string/binary array is more similar to a variadic data buffer than a normal data buffer, which we could also change at some point).

src/nanoarrow/nanoarrow.h
paleolimbot: The element width I'm talking about here is how `ArrowArrayViewGetBufferElementBitWidth(array_view, 2)` might be 64 if the second buffer is the sizes or 0 if the second buffer is a data buffer. The existing (if imperfect) system is that string/binary buffer data has an element bitwidth of 0 instead of 8 to signify "blob" rather than "uint8".

WillAyd (approved):

WillAyd:
lgtm - merge away!

PR #643, opened 2024-10-02
chore(ci): Ensure integration workflow passes

### LINKED ISSUE #641: Integration workflow is failing for Rust and nanoarrow
The Rust errors are due to https://github.com/apache/arrow-rs/issues/5052, and unfortunately I think it just means we have to disable checking Rust compatibility until the community has the bandwidth to review the fix ( https://github.com/apache/arrow-rs/pull/6426 ).

https://github.com/apache/arrow-nanoarrow/actions/runs/11130234810/job/30929145413#step:3:10644

```
  FAILED TEST: primitive Rust producing,  nanoarrow consuming
  <class 'subprocess.CalledProcessError'>: Command '/build/nanoarrow/nanoarrow_ipc_integration < /tmp/tmpyz3mdbbr/afddb6c7_generated_primitive.producer_file_as_stream' returned non-zero exit status 22.
  
  FAILED TEST: primitive_zerolength Rust producing,  nanoarrow consuming
  <class 'subprocess.CalledProcessError'>: Command '/build/nanoarrow/nanoarrow_ipc_integration < /tmp/tmpyz3mdbbr/018cdedf_generated_primitive_zerolength.producer_file_as_stream' returned non-zero exit status 22.
  
  FAILED TEST: primitive_large_offsets Rust producing,  nanoarrow consuming
  <class 'RuntimeError'>: Command failed: /build/nanoarrow/nanoarrow_ipc_integration
  With output:
  --------------
  Validating that /tmp/tmpyz3mdbbr/2dffa3bd_generated_primitive_large_offsets.json_as_file reads identical to /tmp/arrow-integration-8826mv00/generated_primitive_large_offsets.json
  Command VALIDATE failed (22=Invalid argument): Footer flatbuffer verification failed (16) vector header out of range or unaligned
  
  --------------
```

The nanoarrow--nanoarrow error is something about converting swapped endian IPC streams to IPC files since it seems like buffer endian swapping isn't taking place.

https://github.com/apache/arrow-nanoarrow/actions/runs/11130234810/job/30929145413#step:3:10810

```
FAILED TEST: interval nanoarrow producing,  nanoarrow consuming
  <class 'RuntimeError'>: Command failed: /build/nanoarrow/nanoarrow_ipc_integration
  With output:
  --------------
  Validating that /arrow-integration/testing/data/arrow-ipc-stream/integration/1.0.0-bigendian/generated_interval.arrow_file reads identical to /tmp/tmpyz3mdbbr/1.0.0-bigendian_interval.gold.json
  Command VALIDATE failed (22=Invalid argument): Found 12 differences between batches:
  Path: Batch 0.children[0]
  - {"name": "f1", "count": 7, "VALIDITY": [1, 1, 1, 1, 1, 1, 0], "DATA": ["128", "-129", "-5758134827132877538", "-5984203302733857979", "7152141364016051091", "1946707673701025977", "0"]}
  + {"name": "f1", "count": 7, "VALIDITY": [1, 1, 1, 1, 1, 1, 0], "DATA": ["-9223372036854775808", "9223372036854775807", "2199428923299600048", "4988814404219237292", "-7785723983210331805", "-5113702496382483429", "0"]}
  
  Path: Batch 0.children[1]
  - {"name": "f2", "count": 7, "VALIDITY": [1, 1, 0, 1, 1, 0, 1], "DATA": ["128", "-129", "0", "-8588564788114424009", "4714234549317537763", "0", "-915230144468162553"]}
  + {"name": "f2", "count": 7, "VALIDITY": [1, 1, 0, 1, 1, 0, 1], "DATA": ["-9223372036854775808", "9223372036854775807", "0", "4034946345919434632", "-2085181525232227263", "0", "569773930746957043"]}
  
  Path: Batch 0.children[2]
  - {"name": "f3", "count": 7, "VALIDITY": [0, 0, 1, 1, 1, 1, 1], "DATA": ["0", "0", "-91386228020645345", "8297967703166746363", 
  
```

paleolimbot:
Closes #641.

Unfortunately we just have to skip checking Rust compatibility due to https://github.com/apache/arrow-rs/issues/5052 (e.g., https://github.com/apache/arrow-rs/pull/6449 ).

This PR also ensures compatibility with big endian Arrow files and Arrow files from before the continuation token. Support for those had already been added in the decoder but hadn't made it to the stream reader yet.

Local check:

```bash
# Assumes arrow-testing, arrow-nanoarrow, and arrow are all checked out in the same dir
export gold_dir=../arrow-testing/data/arrow-ipc-stream/integration 
export ARROW_NANOARROW_PATH=$(pwd)/build 
pip install -e "../arrow/dev/archery/[all]"
archery integration --with-nanoarrow=true --run-ipc \
    --gold-dirs=$gold_dir/0.14.1 \
    --gold-dirs=$gold_dir/0.17.1 \
    --gold-dirs=$gold_dir/1.0.0-bigendian \
    --gold-dirs=$gold_dir/1.0.0-littleendian \
    --gold-dirs=$gold_dir/2.0.0-compression \
    --gold-dirs=$gold_dir/4.0.0-shareddict
```

### DIFF
--- a/src/nanoarrow/ipc/reader.c
+++ b/src/nanoarrow/ipc/reader.c
@@ -230,6 +230,20 @@ static int ArrowIpcArrayStreamReaderNextHeader(
     // propagated higher (e.g., if the stream is empty and there's no schema message)
     ArrowErrorSet(&private_data->error, "No data available on stream");
     return ENODATA;
+  } else if (bytes_read == 4) {
+    // Special case very, very old IPC streams that used 0x00000000 as the
+    // end-of-stream indicator.
+    uint32_t last_four_bytes = 0;
+    memcpy(&last_four_bytes, private_data->header.data, sizeof(uint32_t));
+    if (last_four_bytes == 0) {
+      ArrowErrorSet(&private_data->error, "No data available on stream");
+      return ENODATA;
+    } else {
+      ArrowErrorSet(&private_data->error,
+                    "Expected 0x00000000 if exactly four bytes are available at the end "
+                    "of a stream");
+      return EINVAL;
+    }
   } else if (bytes_read != 8) {
     ArrowErrorSet(&private_data->error,
                   "Expected at least 8 bytes in remainder of stream");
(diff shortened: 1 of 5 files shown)


src/nanoarrow/ipc/reader.c
```diff
@@ -230,6 +230,20 @@ static int ArrowIpcArrayStreamReaderNextHeader(
     // propagated higher (e.g., if the stream is empty and there's no schema message)
     ArrowErrorSet(&private_data->error, "No data available on stream");
     return ENODATA;
+  } else if (bytes_read == 4) {
```
bkietz, line 233: I think this special case should specifically assert the metadata version we're currently reading. It's never valid for a V5 message to omit the continuation and a V4 message is invalid if it includes that.

And we can't write:

```suggestion
// DON'T commit; broken!
  } else if (private_data->decoder.metadata_version == NANOARROW_IPC_METADATA_VERSION_V4 && bytes_read == 4) {
```

because these peek etc functions are used before we've fully unpacked the schema to know what the metadata version is.

So I think what needs to happen is: the first time a decoder peeks a header and observes the lack of a continuation, we set `ArrowIpcDecoderPrivate::prefix_length = 4`. Later when we decode the metadata version, we can raise an error if the prefix length is not as expected. Also we should raise if the prefix length is inconsistent in any future message peeked by the decoder.

As a pleasant side effect, there will be one fewer argument for you to pass around :D

bkietz (changes requested):

src/nanoarrow/ipc/reader.c
paleolimbot: In a fun turn of events, not all V4 metadata messages have a 4 byte prefix size (i.e., not in the 0.17.1 golden files). I think I included all the other checks you suggested! (And also opened an issue with instructions for how to skip those cases in the future should we decide they don't need to be supported): https://github.com/apache/arrow-nanoarrow/issues/648

PR #644, opened 2024-10-03
python: Use meson-python instead of setuptools

WillAyd:
This PR originally started as a nanobind POC to see if we could convert from Cython over to that, but I scaled that back just to focus on the possiblity of porting the build system to meson (you can see some nanobind components in the history, if you cared to look)

The build system port may have some merits on its own. It is more tightly integrated with the build system of the C code, and has the advantage of being built in parallel. ~Technically it links to the main project by symlinking it as a subproject in the `subprojects` directory; if we cared to support source Python builds we could replace that symlink with a wrap file on distribution, or could even detect within the configuration if the symlink resolves. In the case it does not, we can fallback to a github release~

If the user wanted to develop locally, they could simply symlink the project root into the subprojects folder of Python (the git repo already does this). For source distributions, the wrap system can provide a fallback project to download and reference

If we choose to try and move off of Cython incrementally and use something like nanobind or even C extensions, I think Meson can handle those more capably than the setup.py script


python/src/nanoarrow/meson.build
```diff
@@ -0,0 +1,103 @@
...
+nanoarrow_src_dir = nanoarrow_proj.get_variable('_nanoarrow_src_dir')
+
+fs = import('fs')
+foreach cyf : cyfiles
+  py.extension_module(
+      fs.stem(cyf),
+      sources: [cyf],
+      cython_args: cython_args,
+      include_directories: nanoarrow_src_dir,
+      dependencies: [
```
WillAyd, line 69: This is a little too broad to use every dependency for every Cython file; can refine so that only IPC files need the IPC dep, Device files need the device dep, etc...

python/meson.build
```diff
@@ -0,0 +1,18 @@
...
+    'nanoarrow',
+    'cpp', 'cython',
+    version: '0.6.0.dev0',  # TODO: do we need git archive support?
+    license: 'Apache-2.0',
+    meson_version: '>=1.2.0',
+    default_options: [
+        'warning_level=2',
+        'cpp_std=c++17',
+        'default_library=static',
+        'nanoarrow:ipc=true', # https://github.com/mesonbuild/meson/issues/6728
```
WillAyd, line 11: Ideally we wouldn't need both of these. However, in the current configuration `arrow-nanoarrow` is assumed to be a symlink whereas `nanoarrow` is assumed to be a downloaded external dependency.

In theory we could just change `nanoarrow.wrap` to `arrow-nanoarrow.wrap` and let Meson do its magic, consolidating these options into 1. The problem though, is that the bootstrap.py logic needs to look in the subprojects folder to copy some files, and is unclear when you have multiple arrow-nanoarrow subproject directories which one the bootstrap.py script should reference

meson.build
```diff
@@ -103,7 +105,7 @@ if get_option('ipc')
     )
     nanoarrow_ipc_dep = declare_dependency(include_directories: [incdir],
                                            link_with: nanoarrow_ipc_lib,
-                                           dependencies: [nanoarrow_dep, flatcc_dep])
+                                           dependencies: [nanoarrow_dep])
```
WillAyd, line 108: I think it was a mistake to include the `flatcc_dep` transitively here; I think its only required to build the `nanoarrow_ipc` lib itself, but shouldn't be required by libraries that want to use `nanoarrow_ipc` (?)

python/subprojects/arrow-nanoarrow.wrap
```diff
@@ -14,12 +14,14 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-#
-# This file is part of 'miniver': https://github.com/jbweston/miniver
 
-# Replaced by version-bumping scripts at release time
-version = "0.6.0.dev0"
+[wrap-file]
+directory = arrow-nanoarrow-b78c0395a6fe74e776daf07933583f06567fb99c
```
WillAyd, line 19: This is just pointing to a single commit here that includes some of the changes made to the top level `meson.build` configuration. In a follow up ideally point to main or even a wrapdb release

WillAyd:
> Will this work with CUDA? I imagine it's possible to add `LDFLAGS` to get around this if needed.

Definitely not an area of expertise for me, but I don't think Meson provides that many abstractions over CUDA at the moment

> I do think we have to support building from sdist since I believe that's how the conda-forge setup currently works.

Yep should still be a part of this - when cloning the git repo if the arrow-nanoarrow subproject is there, meson should reference that. For source distributions, they should include the `arrow-nanoarrow.wrap` file which will download a copy form a source (currently this GH branch, but ideally main or better yet the WrapDB)

meson.build
```diff
@@ -56,6 +56,8 @@ endif
 subdir('src/nanoarrow')
 incdir = include_directories('src/')
 
+_nanoarrow_src_dir = include_directories('src/nanoarrow')  # only for use by Cython
```
WillAyd, line 59: Ideally this wouldn't exist at all; didn't look into how to resolve that closely though

.github/workflows/python-wheels.yaml
```diff
@@ -107,9 +107,16 @@ jobs:
...
       - uses: actions/setup-python@v5
+        if: ${{ matrix.config.label != "windows" }}
         with:
           python-version: "3.12"
 
+      - uses: actions/setup-python@v5
+        if: ${{ matrix.config.label == "windows" }}
+        with:
+          python-version: "3.12"
+          architecture: x64
```
WillAyd, line 118: Seems like setup-python still installs a 32 bit Python by default, while cibuildwheel is expected 64 bit on that host

WillAyd:
Putting back in draft mode for now as there _might_ be an upstream bug in meson we need to resolve https://github.com/mesonbuild/meson/issues/13746

...or my understanding is wrong. Either way, I think the way things stand now are a very close approximation of what this solution could look like

paleolimbot (commented):
Just a few thoughts! If we can solve the commit-level integration issue I think we should merge just after the release :slightly_smiling_face:

python/subprojects/arrow-nanoarrow.wrap
WillAyd: The commit level integration is managed through the symlink to the subprojects folder. A wrap file doesn't work like that - instead it just copies the library at a point in time to subprojects, and you would have to manually run `meson wrap update` constantly to get that

.github/workflows/python-wheels.yaml
WillAyd: I believe so...I think cibuildwheel controls that, and since we don't override anything in the pyproject.toml I think it should still generate the same types of wheels as specified here:

https://cibuildwheel.pypa.io/en/stable/options/#build-skip

Admittedly my knowledge of Windows and how it builds for different architectures is limited, so definitely something to verify when wheels actually get build

python/subprojects/arrow-nanoarrow.wrap
WillAyd: To try and bring that all together...I ideally wanted to have a subprojects folder that looked like:

```
python/
  subprojects/
    arrow-nanoarrow/  -- this is a symlink to the project root
    arrow-nanoarrow.wrap
```

Ideally, if meson finds the `arrow-nanoarrow` subproject it should not need the wrap file at all, and just happily use the symlink for a live integration. When it comes to sdists where the symlink will not exist, the wrap file should be responsible for resolving and unpacking a particular `arrow-nanoarrow` distribution to the subprojects folder (probably from the wrap db, but in this PR I have it pointing to this branch to get the config updates)

Due to the "bug" I linked elsewhere in this PR, Meson confusingly still downloads the contents of `arrow-nanoarrow.wrap` even when the `arrow-nanoarrow` symlink exists, so to work around that the local directory structure instead looks like:

```
python/
  subprojects/
    arrow-nanoarrow-dev/  -- this is a symlink to the project root
    arrow-nanoarrow.wrap
```

And the configuration looks like:

```python
nanoarrow_proj = subproject('arrow-nanoarrow-dev', required: false)
if not nanoarrow_proj.found()
  nanoarrow_proj = subproject('arrow-nanoarrow')
endif
```

When developing locally, the arrow-nanoarrow-dev symlink will always be prioritized. Of course, when we create source distributions we end up removing symlinks, so the wrap file comes into play in those instances.

Hope that makes sense but let me know if anything is unclear

paleolimbot: I think I get it, but I would like the sdist to have a full copy of the nanoarrow C library rather than download something. I can experiment a bit to see if I can get that to work (as well as fire up my Windows and CUDA test machines locally to ensure those both work too).

WillAyd: Ah OK. In that case we can probably get rid of the wrap file then; just need to figure out the right way/hook to install a copy into subprojects when generating the sdist

paleolimbot: Maybe in `bootstrap.py`? ADBC sets an environment variable when building a source distribution and we probably could too:

https://github.com/apache/arrow-adbc/blob/47e6a6a5318e9f85ef4eed7903844a05b7614e4f/python/adbc_driver_postgresql/setup.py#L37-L38

WillAyd: Meson-python and meson both actually have a dist option to include subprojects, which I think is where our best bet is

https://meson-python.readthedocs.io/en/stable/how-to-guides/meson-args.html

Of course, it doesn't seem to work right now (I think the symlink behavior tricks it into failure) but let me research that a bit more. Would rather let the build system take care of this if it can

jorisvandenbossche:
I don't really have a stake in this, but just to mention it: for pyarrow we are considering to switch from setuptools to scikit-build-core, and for maintenance overhead there might be some value in using the same build backend for the various python packages in the project (but in the end, we should also use the best tool in each case, and here there is less existing cmake compared to pyarrow, so using meson-python might be the simpler option)

python/subprojects/arrow-nanoarrow.wrap
rgommers: [`meson.add_dist_script`](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonadd_dist_script) may be useful here? It will run when building an sdist, and you can invoke a Python script to copy exactly the files/dirs you want under `subprojects/`.

WillAyd: Ah OK I think that's a great idea. The `--include-subprojects` flag I was looking at seems to just copy the symlink, but not deeply copy the subproject. So a custom script is probably the safest bet

.github/workflows/python-wheels.yaml
paleolimbot: You might have to update the `python-wheels.yaml` workflow to fire on `python/meson.build` instead of `python/setup.py` to get it to run on this PR (in which case we can inspect the results and find out!)

WillAyd:
Yea I think that or what @rgommers  suggested (run a script during Meson's dist hook to vendor what we need) are the two options. 

I'll take another look at this next week

python/generate_dist.py
```diff
@@ -0,0 +1,53 @@
...
+
+    subproj_subproj_dir = subproj_dir / "subprojects"
+    subproj_subproj_dir.mkdir()
+    for f in (src_dir / "subprojects").glob("*.wrap"):
+        shutil.copy(f, subproj_subproj_dir / f.name)
+
+    target_src_dir = subproj_dir / "src"
+    shutil.copytree(src_dir / "src", target_src_dir)
+
+    # this files are only needed by bootstrap.py
```
WillAyd, line 43: The amount of scripting here is unfortunate...would still be nice to replace bootstrap.py, but not sure I want to tackle that in this PR

paleolimbot: For this PR I think you will have the most success just renaming `bootstrap.py` to `generate_dist.py` and including a meson lib definition for nanoarrow in `python/meson.build` (i.e., don't try to build nanoarrow C using a subproject).

python/src/nanoarrow/meson.build
```diff
@@ -0,0 +1,99 @@
...
+# under the License.
+
+# Try to resolve the symlink to a subproject first and fall back
+# to the wrap entry if that is unsuccessful. Ideally Meson would
+# take care of this for us, but there is a bug upstream
+# https://github.com/mesonbuild/meson/issues/13746#issuecomment-2392510954
+nanoarrow_proj = subproject('arrow-nanoarrow')
+nanoarrow_dep = nanoarrow_proj.get_variable('nanoarrow_dep')
+nanoarrow_ipc_dep = nanoarrow_proj.get_variable('nanoarrow_ipc_dep')
+nanoarrow_device_dep = nanoarrow_proj.get_variable('nanoarrow_device_dep')
```
paleolimbot, line 25: Again, I think you want to vendor the files using `bootstrap.py` or whatever facility meson provides for that and not attempt to get the nanoarrow-as-a-subproject bit working (orthogonal battle to using meson-python vs. setuptools).

python/pyproject.toml
```diff
@@ -38,7 +38,11 @@ Changelog = "https://github.com/apache/arrow-nanoarrow/blob/main/CHANGELOG.md"
 
 [build-system]
 requires = [
-    "setuptools >= 61.0.0",
+    "meson-python",
```
paleolimbot, line 41: We probably need a version constraint here?

meson.build
paleolimbot: I'm unclear on the difference but yes, the flatcc headers are deliberately not included in nanoarrow's public headers such that a library using nanoarrow does not need `flatcc/*` files in the include path.

python/generate_dist.py
WillAyd: What do you mean by "including a meson lib definition?"

paleolimbot: Running `bootstrap.py` will get you `python/vendor/nanoarrow.c` and friends, and you should be able to put the appropriate `nanoarrow_c_lib = library(...)` in `python/meson.build`?

WillAyd: Ah OK - so you want to run bootstrap.py continuously, not just when the distribution is being created?

The current form uses the subproject symlink for live integration with the C project and only vendors when the distribution is created

paleolimbot: If you can get all the tests passing and all the wheels building with that go for it! I think you will have more success tackling those two battles separately but the time is yours :slightly_smiling_face:

python/src/nanoarrow/meson.build
```diff
@@ -0,0 +1,99 @@
...
+    'c_array_stream.py',
+    'c_buffer.py',
+    'c_schema.py',
+    'device.py',
+    'ipc.py',
+    'iterator.py',
+    '_repr_utils.py',
+    'schema.py',
+    'visitor.py',
+]
```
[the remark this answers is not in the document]
WillAyd: I don't think so. Here's some documentation in Meson as to why:

https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard

python/generate_dist.py
WillAyd: OK I understand your point of view looking at this some more. The bootstrapping mechanism is pretty tightly engrained as is, and would be an effort in and of itself to replace.

Still looking at the right balance here - might take me a bit

python/src/nanoarrow/__init__.py
```diff
@@ -75,7 +77,8 @@
 from nanoarrow.array import array, Array
 from nanoarrow.array_stream import ArrayStream
 from nanoarrow.visitor import nulls_as_sentinel, nulls_forbid, nulls_separate
-from nanoarrow._version import __version__  # noqa: F401
+
+__version__ = importlib.metadata.version("nanoarrow")
```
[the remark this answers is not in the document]
paleolimbot: I think it'll have to be `0.7.0.dev[0-9]+` to work with the nightly wheels!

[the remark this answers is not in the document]
WillAyd: As far as whether or not this approach is viable, here is the Python packaging documentation:

https://packaging.python.org/en/latest/discussions/single-source-version/#single-source-version

The date on that is pretty recent and I believe importlib.metadata was provisional up until 3.10, so a lot of established packages probably had to come up with their own solution

WillAyd:
OK I removed any of the subproject usage for now and went back to the bootstrap approach. This seems to work well when building in the repository, but the sdist installs are broken. 

Need to look through that a bit more but am open to ideas. The main limitation is that Meson does not allow you to do in-source builds, so the vendoring of files naturally does not happen until build time. But that is probably too late for the sdist

WillAyd:
I think the problem is that meson does not allow you to generate files in the source tree at all, so when doing a commit-level integration you need the bootstrap script to write the files to the build directory, but when generating an sdist the build directory doesn't yet exist

WillAyd:
Hmm do you mean the build tree or the source tree? The build tree won't exist until after the sdist is created, and you can't be sure what the build directory will actually be named since it can be controlled by the user

WillAyd:
I think your proposal would work if you only ever installed from an sdist, but wouldn't work when building from this repo. Let me see if I can clarify.

Assume we had a structure like this:

```
src/
  nanoarrow.c
python/
  src/
    nanoarrow_c.pxd
```

The PR as it currently stands will copy the `src/nanoarrow.c` file into the build folder and reference it from there:

```
src/
  nanoarrow.c
python/
  src/
    nanoarrow_c.pxd
  builddir/  # generated at build time
    nanoarrow.c
```

What you are proposing is to generate an sdist before installing the package with a layout that may look something like:

```
src/
  nanoarrow.c
python/
  src/
    nanoarrow_c.pxd
  vendor/
    nanoarrow.c
```

However, I don't know of a simple way for Meson/ninja to resolve to a source file potentially in the source tree (when installing from an sdist) while falling back to looking for it in the build folder (when building in this repo)

python/src/nanoarrow/__init__.py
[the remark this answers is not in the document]
jorisvandenbossche: > is there any issues with versioning in this way? Is there any cost to importing `importlib.metadata` by default?

I would personally advise to not use `importlib.metadata.version` for creating the `__version__` attribute, because it will add some avoidable overhead (without testing I don't know how big this would be, but in general we want to keep the import as lean as possible, and this seems unnecessary overhead to scan installed packages, when the `__version__` attribute can (in the built packages, i.e. injected during build-time) simply be a static string). 

In addition, when using editable installs, this might also not always work as expected (the metadata version does not update automatically).

> As far as whether or not this approach is viable, here is the Python packaging documentation:
> 
> https://packaging.python.org/en/latest/discussions/single-source-version/#single-source-version

Note that that page does not actually mention to use `importlib.metadata.version` to populate `__version__`, just that there is a desire that those return the same thing. 

There has been a long discussion on this the last months at https://discuss.python.org/t/please-make-package-version-go-away/58501, that resulted in the updated page linked above (it's long, don't read it all ;), but I do see that it does actually have some discussion on the overhead, estimated there to be 20-40ms (although it might also depend on how big your environment is))

.github/workflows/build-and-test-device.yaml
```diff
@@ -156,4 +157,4 @@ jobs:
         if: matrix.config.label == 'with-cuda'
         run: |
           cd python
-          python3 -m pytest -vv
+          python3 -m pytest -vv tests
```
WillAyd, line 160: Unfortunately pytest does discover tests recursively from the subproject if you try discovery from the root of the project; need to see if there's a more robust way of handling this

rgommers: `norecursedirs` in `pytest.ini`, like this: https://github.com/scipy/scipy/blob/main/pytest.ini#L3 ?

.github/workflows/python-wheels.yaml
```diff
@@ -107,9 +109,16 @@ jobs:
           fetch-tags: true
 
       - uses: actions/setup-python@v5
+        if: ${{ matrix.config.label != "windows" }}
         with:
           python-version: "3.12"
 
+      - uses: actions/setup-python@v5
+        if: ${{ matrix.config.label == "windows" }}
```
paleolimbot, line 117:
```suggestion
        if: ${{ matrix.config.label == 'windows' }}
```

...perhaps why the wheels CI job isn't buildings:

```

The workflow is not valid. .github/workflows/python-wheels.yaml (Line: 112, Col: 13): Unexpected symbol: '"windows"'. Located at position 24 within expression: matrix.config.label != "windows" .github/workflows/python-wheels.yaml (Line: 117, Col: 13): Unexpected symbol: '"windows"'. Located at position 24 within expression: matrix.config.label == "windows"
```

.github/workflows/python-wheels.yaml
```diff
@@ -107,9 +109,16 @@ jobs:
           fetch-tags: true
 
       - uses: actions/setup-python@v5
+        if: ${{ matrix.config.label != "windows" }}
```
paleolimbot, line 112:
```suggestion
        if: ${{ matrix.config.label != 'windows' }}
```

WillAyd:
@rgommers maybe you have an idea on the Windows failure? Tried looking through upstream discussions but its a rather broad topic:

https://github.com/apache/arrow-nanoarrow/actions/runs/11559159131/job/32173258300?pr=644#step:6:184

```
Need python for x86_64, but found x86
```

AFAIU the container which cibuildwheel installs a 32 bit Python but the compiler being used is 64 bit?

rgommers:
> AFAIU the container which cibuildwheel installs a 32 bit Python but the compiler being used is 64 bit?

Yes indeed. The tl;dr is that if you want to build wheels for 32-bit Python on 64-bit Windows, Meson doesn't automatically configure MSVC in 32-bit mode for you (rationale: there could be non-Python library or executable targets, and why would those be 32-bit because somewhere else in `meson.build` a 32-bit Python is detected).

To fix it, explicitly configure 32-bit MSVC in your `cibuildwheel`-using job: 
https://github.com/PyWavelets/pywt/blob/361cc9a4c80bc2603fae99e151f2a391c0b376f9/.github/workflows/wheel_tests_and_release.yml#L204-L214

meson.build
WillAyd: Yea this removal here essentially removes the flatcc headers from being included by downstream targets that use the nanoarrow_ipc_dep; this is akin to changing the flatcc target inclusion from PUBLIC to PRIVATE in CMake

paleolimbot (approved):
Truly heroic effort getting all of this to pass!

The only issue I see is that flatcc sources aren't vendored/not contained in the sdist. I think that is OK for now (solvable if somebody asks for it) and probably only directly affects somebody trying to do Python development on an airplane 🙂 .

WillAyd:
> The only issue I see is that flatcc sources aren't vendored/not contained in the sdist.

The latest commit should add that functionality now, relying on Meson's own functionality to bundle subprojects as part of dist creation. It's a bit wonky that flatcc is bundled that way and the rest of the files are created by our custom script reading from a CMake file...but I'm not sure that can be resolved in this PR. Definitely would like to tackle that in follow up(s)

paleolimbot (approved):

WillAyd:
Thanks @paleolimbot for the review and @rgommers for all of the guidance!

PR #645, opened 2024-10-03
docs(python): Add example of python package with nanoarrow C extension

WillAyd:


paleolimbot (commented):
Thank you!

Would it be appropriate to put a `pip install` line in the examples.yaml workflow?

https://github.com/apache/arrow-nanoarrow/blob/85adb56cdd095910a29c4227948a2109fbda5b8b/.github/workflows/examples.yaml#L114

paleolimbot (approved):
Thank you!

Issue #646, opened 2024-10-04
Re-enable CUDA checks on CI


paleolimbot:
The self-hosted "cuda" runner is currently not picking up any jobs. The runner was previously provided by Voltron Data; however, we may have to re-negotiate if or how it could be provided (or perhaps this is a temporary outage). In any case, we have to remove the workflow that ran there to keep CI rolling.

mike-wendt:
Hi @paleolimbot I'm looking into the cause of why this is no longer working. Our intent is to keep providing this service to the community. I'll let you know when I have more information about this outage.

mike-wendt:
For visibility this was an issue with our deployment where the runners were not synced and stopped working. It has been corrected and the runners for `arm` and `cuda` are active again. Apologies for the inconvenience

paleolimbot:
Thank you!

PR #647, opened 2024-10-04
chore(ci): Temporarily disable CUDA integration jobs that are not being picked up

paleolimbot:
The self-hosted "cuda" runner is currently not picking up any jobs, causing CI to hang at "yellow" for several days for each run until the job times out. Until this is resolved, we need to disable that job (https://github.com/apache/arrow-nanoarrow/issues/646).


assignUser (approved):

assignUser:
I'll investigate the status of the runners

assignUser:
The runners are back, this can be reverted :)

PR #667, opened 2024-10-29
chore: Improve tidiness of nanoarrow_testing.hpp

paleolimbot:
There were still a number of implementation details left in nanoarrow_testing.hpp. As part of the effort to improve the testing utilities and remove the Arrow C++ requirement, a few things might need to get added here and this seemed like a good place to start.


WillAyd (approved):
lgtm

Issue #599, opened 2024-08-27
Define the relationship between nanoarrow and C++


bkietz:
Currently it's not explicit what level of C++ is recommended for use in the nanoarrow project. The library itself is of course a minimal API in pure C, but a C++ helper library is packaged alongside and the unit tests are all C++.

This is part of general tension in the project between ergonomics and minimalism of the API. New helper functions are suspect for fear of bloating the API, but this leads to extremely verbose unit tests and code examples.

WRT using unit tests as documentation: although I think it's useful to have complete and self-contained examples of API usage, I think the priority for unit tests should be coverage rather than didactics.

I'd propose:
- we survey the community to find out whether the C++ layer is seriously used or could be left internal
  - if the C++ layer **is** widely used, does strict minimalism apply to that as well or can the C++ layer grow more freely?
  - document these decision in .rst
- we should write a large body of pure C example code, explicitly freeing up unit tests to be more terse and idiomatic
  - perhaps limiting unit tests to C++17, or relying more on the python layer for unit testing
- add DSLs available to the tests allowing schema and data literals to be written concisely

WillAyd:
This is a great question. I've documented using the C++ layer myself in some youtube videos on nanoarrow. I've personally been hoping to see it grow a little bit more to make it easier to iterate some of the nanoarrow structures (for example, I think it would be nice to have an iterator for StructArrays that helps you work with dataframe data), but I also don't have a clear point of view on where exactly the line should be drawn

paleolimbot:
Definitely a great point that this is not well defined! A few initial thoughts:

> this leads to extremely verbose unit tests and code examples.

Things like IPC, ADBC, and integration testing are disproportionately affected here, since they are essentially nanoarrow "users" more so than the nanoarrow C library itself. These are places that there is significant verbosity unrelated to the function being tested (e.g., creating an array to roundtrip through IPC), and where I wish I had things like `ArrayFromJSON()`. We are a C library and C is verbose, but I think the key to all of this is limiting the C-verbosity to the function being tested.

> find out whether the C++ layer is seriously used

I am aware of `nanoarrow::UniqueXXX` helper usage (Snowflake Python connector, cudf, ADBC) and array stream helpers (ADBC). I recently used the buffer-from-std-vector helpers for some GeoArrow tests also.

> does strict minimalism apply to that as well or can the C++ layer grow more freely?

[...]

PR #668, opened 2024-10-29
refactor: Split up nanoarrow.hpp into multiple .hpp files

paleolimbot:
This PR splits `nanoarrow.hpp`, which was getting large, into multiple files (bundling them, of course, when creating the bundled distribution). This is intended to make space for new helpers that are being proposed in #599 although I think it's also just a good idea in general since the utilities to create streams and buffers using templating are in a pretty different category from the things that keep you from leaking memory and many users will only need part of the API. Because it's header-only on top of the C runtime, I think this is probably a good pattern to enable.

This doesn't quite solve `nanoarrow_device.hpp` and `nanoarrow_ipc.hpp`...they currently still just include all of `nanoarrow.hpp`. I think we can probably solve that in the bundler but I might punt on that since it's not actually hurting anything at the moment.


meson.build
```diff
@@ -62,6 +62,16 @@ install_headers(
     subdir: 'nanoarrow',
 )
 
+install_headers(
```
WillAyd, line 65: Very minor but you could do:

```python
installl_subdir('src/nanoarrow/hpp', 'nanoarrow/hpp')
```

here if you want all the headers installled

src/nanoarrow/hpp/exception.hpp
WillAyd, line 1: Since array_stream and buffer are "first class" nanoarrow objects, but things like exceptions, operators, unique are not, maybe we should have another directory structure for `nanoarrow/hpp/util` to house the latter?

meson.build
paleolimbot: Will that also install other files in those directories or just headers? (I am about to move the tests to pair with the hpp files too, so there will be some cc files).

src/nanoarrow/hpp/exception.hpp
paleolimbot: I'm not sure what the best organization for that is yet and I am not sure that separating those things now will add much clarity. (Probably when more things are added there may come a separation that becomes clear).

meson.build
WillAyd: Hmm that's a good point - might do everything. This is fine for now

meson.build
```diff
@@ -210,7 +232,7 @@ if get_option('tests')
       ipc_test_files = {
           'ipc-decoder': {
               'src': 'decoder',
-              'deps': [nanoarrow_ipc_dep, arrow_dep, gtest_dep, gmock_dep],
+              'deps': [nanoarrow_ipc_dep, flatcc_dep, arrow_dep, gtest_dep, gmock_dep],
```
WillAyd, line 235: Nice catch on this - noticed in https://github.com/apache/arrow-nanoarrow/pull/673 as well

WillAyd (approved):

PR #669, opened 2024-10-29
refactor: Replace bootstrap script with explicit Cython declarations

WillAyd:


WillAyd:
Latest commit replaces our custom bootstrapping with autopxd2. 

If does generate the following warnings at runtime:

```
<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: UserWarning: enum class ArrowTimeUnit not importable from nanoarrow_c. You are probably using a cpdef enum declared in a .pxd file that does not have a .py  or .pyx file.
```

need to investigate further but I think this has the potential to greatly simplify our Python build system

PR #674, opened 2024-11-02
fix(r): sub-day precision Date should be floored when treated as integer

### LINKED ISSUE #665: [R] Sub-day precision Date class must be converted to integers by floor, not trunc
I noticed that the conversion of Date type with sub-day precision is not correct.
`arrow` package has the same problem (apache/arrow#44502).

``` r
d <- as.Date(-0.1)
d
#> [1] "1969-12-31"
as.vector(nanoarrow::as_nanoarrow_array(d))
#> [1] "1970-01-01"
```

<sup>Created on 2024-10-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

eitsupi:
Fix #665

### DIFF
--- a/r/R/as-array.R
+++ b/r/R/as-array.R
@@ -262,8 +262,14 @@ as_nanoarrow_array.Date <- function(x, ..., schema = NULL) {
   switch(
     parsed$type,
     date32 = {
+      int_vec <- if (is.integer(x)) {
+        x
+      } else {
+        floor(as.numeric(x))
+      }
+
       storage <- as_nanoarrow_array(
-        as.integer(x),
+        int_vec,
         schema = na_type(parsed$storage_type)
       )
       nanoarrow_array_set_schema(storage, schema)
--- a/r/R/as-array.R
+++ b/r/R/as-array.R
@@ -263,9 +263,9 @@ as_nanoarrow_array.Date <- function(x, ..., schema = NULL) {
     parsed$type,
     date32 = {
       int_vec <- if (is.integer(x)) {
-        x
+        as.integer(x)
       } else {
-        floor(as.numeric(x))
+        as.integer(floor(as.numeric(x)))
       }
 
       storage <- as_nanoarrow_array(
--- a/r/R/as-array.R
+++ b/r/R/as-array.R
@@ -263,7 +263,7 @@ as_nanoarrow_array.Date <- function(x, ..., schema = NULL) {
     parsed$type,
     date32 = {
       int_vec <- if (is.integer(x)) {
-        as.integer(x)
+        unclass(x)
       } else {
         as.integer(floor(as.numeric(x)))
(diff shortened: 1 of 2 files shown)


r/tests/testthat/test-as-array.R
```diff
@@ -464,7 +464,13 @@ test_that("as_nanoarrow_array() works for Date -> na_date32()", {
   expect_identical(as.raw(array$buffers[[1]]), as.raw(0x03))
   expect_identical(
     as.raw(array$buffers[[2]]),
-    as.raw(as_nanoarrow_buffer(c(10957L, 19391L, NA)))
+    as.raw(as_nanoarrow_buffer(c(10957L, 19391L, 0L)))
```
eitsupi, line 467: I do not know why this change occurred.
(I also do not understand what this test means.)

eitsupi (commented):
As a side note, it seems that date32 is currently output as a double vector in R, but I think integer is fine. (And that would be more efficient, right?)

r/R/as-array.R
```diff
@@ -262,8 +262,14 @@ as_nanoarrow_array.Date <- function(x, ..., schema = NULL) {
   switch(
     parsed$type,
     date32 = {
+      int_vec <- if (is.integer(x)) {
+        as.integer(x)
```
paleolimbot, line 266:
```suggestion
        x
```

paleolimbot (approved):
Thank you! Just one nit.

> As a side note, it seems that date32 is currently output as a double vector in R, but I think integer is fine. (And that would be more efficient, right?)

This is to better align with how base R normally handles this (although I think either are OK).

``` r
typeof(as.Date("2001-02-03"))
#> [1] "double"
typeof(Sys.Date())
#> [1] "double"
```

r/R/as-array.R
eitsupi: Thinking about it again, I think this is necessary to remove the class attribute (without it, wouldn't we be stuck in an endless loop?).

paleolimbot: Ah, I didn't get that from reading. Either a comment or an `unclass(x)` would work for me so that we don't accidentally simplify this later!

eitsupi:
> This is to better align with how base R normally handles this (although I think either are OK).
> 
> ```r
> typeof(as.Date("2001-02-03"))
> #> [1] "double"
> typeof(Sys.Date())
> #> [1] "double"
> ```

Indeed, it seems that when a number is converted to Date with `as.Date()`, the original numeric type is preserved.
So I think it does not seem surprising that the output of nanoarrow would be an integer.

``` r
typeof(as.Date(1L))
#> [1] "integer"
```

<sup>Created on 2024-11-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #677, opened 2024-11-09
test: Make Arrow C++ dependency optional

WillAyd:


CMakeLists.txt
```diff
@@ -422,26 +422,31 @@ if(NANOARROW_BUILD_TESTS)
...
+    message(STATUS "Arrow SO version: ${ARROW_FULL_SO_VERSION}")
 
-  # Give caller the option to link a static version of Arrow C++
-  if(NANOARROW_ARROW_STATIC)
-    set(NANOARROW_ARROW_TARGET arrow_static)
-  else()
-    set(NANOARROW_ARROW_TARGET arrow_shared)
-  endif()
+    # TODO: does this propoagate to projects using as a subproject?
+    add_compile_definitions("-DNANOARROW_ARROW_FOUND")
```
paleolimbot, line 431: I think it does. You probably want `target_compile_definitions(xxxx_test PRIVATE -DNANOARROW_ARROW_FOUND)` for every test because the tests are what actually need this define set. You could do something fancy like we do with the `nanoarrow_coverage_config` (define an interface target that the tests "link to"), or bite the bullet and use a `foreach()` loop to link tests (sort of like you do in the meson config), which we probably should have done a while ago anyway.

paleolimbot (commented):
I know you're still working on this, but just two high level comments to think about:

- We probably want a CMake option of `-DNANOARROW_TESTS_WITH_ARROW` (where if it's ON, configuration will fail if Arrow is not found, and if it's OFF, tests are never built with arrow). Having it be automatic makes it relatively easy to accidentally think you are testing with Arrow C++ when CMake actually failed to locate it.
- There are a few tests where the entire test is useless, but many of them only have Arrow C++ as a small component (notably: the schema init ones). It doesn't have to be perfect at the end of this PR, but before the next release we will ideally have full (or at least not reduced) test coverage without any of the Arrow C++ integration tests running.
- An alternative to compiling tests with a define is to move the (portions of) tests that need Arrow C++ to separate files (Arrow C++-specific integration tests, perhaps under integration/).

CMakeLists.txt
```diff
@@ -517,6 +520,27 @@ if(NANOARROW_BUILD_TESTS)
                         gmock_main
                         nanoarrow_coverage_config)
 
+  list(APPEND
```
WillAyd, line 523: Just kept the loop scoped to adding the define for now, although there is (as you've already mentioned) potential to use this for other parts of the config

src/nanoarrow/common/array_test.cc
```diff
@@ -1719,7 +1812,8 @@ TEST(ArrayTest, ArrayTestAppendToRunEndEncodedArray) {
   EXPECT_EQ(ArrowArrayFinishBuilding(&array, NANOARROW_VALIDATION_LEVEL_FULL, &error),
             NANOARROW_OK);
 
-#if !defined(ARROW_VERSION_MAJOR) || ARROW_VERSION_MAJOR < 12
+#if !defined(NANOARROW_BUILD_TESTS_WITH_ARROW) || !defined(ARROW_VERSION_MAJOR) || \
```
WillAyd, line 1815: Do we want to add this here or leave as is?

dev/release/verify-release-candidate.sh
```diff
@@ -243,12 +243,14 @@ test_cmake_project() {
 
 test_c() {
   show_header "Build and test C library"
-  test_cmake_project build . -DNANOARROW_BUILD_TESTS=ON -DNANOARROW_IPC=ON
+  test_cmake_project build . -DNANOARROW_BUILD_TESTS=ON -DNANOARROW_IPC=ON \
+                     -DNANOARROW_BUILD_TESTS_WITH_ARROW=ON
```
paleolimbot, line 247: For verification it would be useful to default this to OFF (because it significantly improves the instructions required for verifying). It can always be turned on (and perhaps we should turn it on in .github/workflows/verify.yaml) via the extra cmake args.

src/nanoarrow/common/array_test.cc
paleolimbot: Either is OK by me! I think it makes sense that if the version isn't defined that we didn't build with arrow C++ (but also it might make searching for all the Arrow C++ conditional test blocks easier to include the extra check).

paleolimbot (approved):
Sorry I missed the flip from draft -> ready!

This is great! Thank you!

PR #687, opened 2024-11-19
feat(python): Add map type constructor

paleolimbot:
When working on fixing a problem with Arrow C++'s map type import, I realized there was no way to create map types or any of the canonical extensions. Extensions are slightly different and I'll tackle them later, but map types are relatively straightforward and follow the existing pattern of all the other constructors:

```python
import nanoarrow as na
na.map_(na.string(), na.int32())
#> <Schema> map<entries: struct<key: string, value: int32>>
```


WillAyd (approved):
very nice

Issue #660, opened 2024-10-13
Improve release/post-release workflow


paleolimbot:
A few nits from the 0.6.0 release:

- The version updater is updating `meson.build`s versions incorrectly (inserting an extra `)` and updating the meson version as well as the nanoarrow version
- A script for uploading wheels would be great
- A script for uploading the documentation would be great
- Creating the `apache-arrow-nanoarrow-0.7.0.dev` tag is not documented
- Updating the R NEWS from the changelog would be nice as a script
- Separate the "sign" step from things that don't require somebody in the KEYS file (e.g., updating release changelog

paleolimbot:
It looks like all of these are still problems! I'll go through and fix some of these in the near future 🙂

paleolimbot:
We also need to add back the bit about removing release artifacts!

https://github.com/apache/arrow-nanoarrow/pull/767#discussion_r2188771102

paleolimbot:
...also make a note that the email to the announce mailing list has to be in plain text or it will be rejected!

PR #689, opened 2024-11-21
CI: Automate pypi wheel publishing

WillAyd:
ref https://github.com/apache/arrow-nanoarrow/issues/660

This is currently pointing to the test pypi environment and we can trigger manually for now


WillAyd:
Before we try this I think we will need to enable trusted publishing in pypi (better to start with the test environment)

https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#configuring-trusted-publishing

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,68 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Build and upload to PyPI
+
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
```
assignUser, line 23: Wheels are not signed during release right? Otherwise I'd say download the wheels from the release and push those. (Or make use of reproducible builds)

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,68 @@
...
+    environment:
+      name: testpypi
+      url: https://test.pypi.org/project/arrow-nanoarrow/
+    permissions:
+      id-token: write
+    # TODO: add back in the below condition when productionalized
+    # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+    steps:
+      - uses: actions/download-artifact@v4.1.8
+        with:
```
assignUser, line 58: The action supports trusted publishing, should be easy to setup: https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing

> This method can be used in automated environments and eliminates the need to use manually generated API tokens to authenticate with PyPI when publishing.

https://docs.pypi.org/trusted-publishers/

assignUser: (sorry comment one the wrong line ^^)

assignUser: also missed your comment above, very good!

paleolimbot (commented):
Thank you for looking into this!

I have to admit that what I had in mind was from the local end...something like "find the wheels job on the RC branch:

https://github.com/apache/arrow-nanoarrow/blob/e54b7df525fa1d310a96687bd99902823402b26c/dev/release/02-sign.sh#L50-L55

...then download artifacts ( https://cli.github.com/manual/gh_run_download ), then `twine` to upload.

Something like this can work too, although there is a bit of a chicken-and-egg problem with the Apache vote: the wheels will have already been built against the release commit before the release candidate vote happens, and we need to upload them after the release vote passes. (Not technically a problem since binaries aren't technically voted on, just not strictly necessary).

assignUser:
You could add the wheels to the github release (though I think they might need to be signed then?) and use these in the workflow (or just use gh to get them as in your example)

>  the wheels will have already been built against the release commit before the release candidate vote happens, and we need to upload them after the release vote passes.

I don't see the issue? I assume you publish the github release once the vote is finished? That would trigger this workflow and get the wheels voted on and upload them?

assignUser:
Ah yeah, we were talking about the same thing ^^ +1

PR #693, opened 2024-12-16
feat: Add ZSTD decompression support to IPC reader

paleolimbot:
This PR implements ZSTD buffer decompression in the `ArrowIpcDecoder` and in the `ArrowArrayStreamReader` when built with `-DNANOARROW_IPC_WITH_ZSTD=ON`. It also allows a user to inject support for these into the `ArrowIpcDecoder` if for whatever reason they don't have control over the build flags (or want to use ZSTD that has been made available to them in a different way).

This doesn't implement multithreaded decompression but does allow a user to implement it by not using the default `ArrowIpcSerialDecompressor()`. This could be included in header-only C++ if there is some interest.

A non-trivial example in Python bindings (where were also wired up to support it):

```python
import urllib.request
import nanoarrow as na
url = "https://github.com/geoarrow/geoarrow-data/releases/download/v0.1.0/ns-water-basin_point-wkb.arrow"

# Work around the 'no arrow file support'
with urllib.request.urlopen(url) as f:
    f.read(8)
    print(na.ArrayStream.from_readable(f).read_all())
#> nanoarrow.Array<non-nullable struct<OBJECTID: int64, FEAT_CODE: string, ...>[46]
#> {'OBJECTID': 1, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EB000', 'RIVER': 'BAR...
#> {'OBJECTID': 2, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EC000', 'RIVER': 'ROS...
#> {'OBJECTID': 3, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EA000', 'RIVER': 'TUS...
#> {'OBJECTID': 4, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01DA000', 'RIVER': 'MET...
#> {'OBJECTID': 5, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01ED000', 'RIVER': 'MER...
#> {'OBJECTID': 6, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EE000', 'RIVER': 'HER...
#> {'OBJECTID': 7, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EG000', 'RIVER': 'GOL...
#> {'OBJECTID': 8, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EF000', 'RIVER': 'LAH...
#> {'OBJECTID': 9, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EJ000', 'RIVER': 'SAC...
#> {'OBJECTID': 10, 'FEAT_CODE': 'WABA30', 'BASIN_NAME': '01EH000', 'RIVER': 'EA...
#> ...and 36 more items
```

This PR doesn't implement the R bindings (because adding a zstd dependency there is a can of worms better suited to another PR).


meson.build
```diff
@@ -118,6 +118,7 @@ if get_option('ipc')
 
     nanoarrow_ipc_lib = library(
```
WillAyd, line 119: Does this need the zstd dependency? Looks like it is in the wrapdb, so can just `meson wrap install zstd` to get the wrap file created then

```
zstd_dep = dependency('zstd')
nanoarrow_ipc_lib = library(
    ...,
    dependencies: [zstd_dep, ...]
)
````

Comment (paleolimbot) [reply meson.build:127]: Thanks!

File: python/src/nanoarrow/meson.build [hunk 2]
```diff
@@ -77,9 +78,11 @@ foreach cyf : cyfiles
     if stem in ['_array', '_device']
         cyfile_deps += [nanoarrow_device_dep]
     elif stem == '_ipc_lib'
-        cyfile_deps += [nanoarrow_ipc_dep, flatcc_dep]
+        cyfile_deps += [nanoarrow_ipc_dep, flatcc_dep, zstd_dep]
     endif
 
+    # TODO: Where does -DNANOARROW_IPC_WITH_ZSTD go?
```
paleolimbot, line 84: @WillAyd I can't seem to figure out how to get this to do anything? (Also, the zstd subproject doesn't show up in the log I get from `pip install -e .`?)

paleolimbot: Thank you! All of that led me to the problem that was...I'd mixed up `cpp_args` with `c_args` in nanoarrow (cpp in R stands for "c preprocessor" 😬 ). The compile_commands.json tip was bang on!

python/pyproject.toml
```diff
@@ -45,6 +45,11 @@ requires = [
 build-backend = "mesonpy"
 
 [tool.meson-python.args]
+# Consistent version for zstd, rather than attempting to delocate whatever
```
WillAyd, line 48: When you say delocate are you just referring to just to macOS or does this affect all platforms?

src/nanoarrow/ipc/decoder.c
```diff
@@ -1485,14 +1580,27 @@ static ArrowErrorCode ArrowIpcMakeBufferFromShared(struct ArrowIpcBufferFactory*
...
-  NANOARROW_UNUSED(error);
-
   struct ArrowIpcSharedBuffer* shared =
       (struct ArrowIpcSharedBuffer*)factory->private_data;
-  ArrowBufferReset(dst);
-  ArrowIpcSharedBufferClone(shared, dst);
-  dst->data += src->body_offset_bytes;
-  dst->size_bytes = src->buffer_length_bytes;
+
+  int needs_decompression = 0;
```
WillAyd, line 1587: Ultra nit but is there any reason why we shy away from using stdbool?

python/pyproject.toml
paleolimbot: I should check the logs, but I believe on both MacOS and Linux the wheel build attempts to copy shared library dependencies into the build. In the MacOS case, it was attempting to copy a version of zstd that would have required updating the minimum supported MacOS version. I'm not sure if this would have happened on Linux (because I'm not sure if that library was installed on manylinux2014), but it seems better to have the same version of zstd sitting in the wheel build? (Maybe handling this via build arguments in the CI job that builds the wheels is better?)

src/nanoarrow/ipc/decoder.c
paleolimbot: Early on I was trying to minimize the number of C headers that were included via `nanoarrow.h` directly. Because there are some boolean flag arguments there, the precedent spread elsewhere for consistency (sort of...I think there are one or two `char`s instead of `int`s).

Possibly better would be a `#define ArrowBoolean int` or `typedef ArrowBoolean int` (or char), or just include `stdbool.h`.

python/pyproject.toml
WillAyd: Ah ok. I can take a look too. 

Thinking out loud, we might need to be careful if it's the wheel build itself installing the libraries (because the subproject meson configurations instruct the build system to do so) or if it's the wheel repair jobs that are bundling improper libraries. I'm guessing it's the former. 

I think we could add the following to pyproject.toml to prevent the former issue:

[tool.meson-python.args]
install = ['--skip-subprojects']

Or possibly even statically link the zstd subproject (?)

src/nanoarrow/ipc/decoder.c
WillAyd: Sounds good. Not important enough to change I think. 

Can also wait 20 years for C23 to be fully supported by MSVC

python/pyproject.toml
paleolimbot: I think we have the --skip-subprojects in pyrpoject.toml already...I think the issue was that meson actively tries to avoid building and statically linking if it can avoid it (but this is exactly what we want for the wheel, I think!). I'm happy with this at the moment but we should definitely take a closer look at exactly what's packaged before we release the first version with the meson-python build.

python/pyproject.toml
```diff
@@ -45,6 +45,11 @@ requires = [
...
+# version was on the build host.
+# TODO: We need to override this somehow on conda-forge because we actually
+# do want the host fallback.
+setup = ['--force-fallback-for=zstd']
```
WillAyd, line 52: Giving this another review I think you should actually move this to the `meson.build` configuration on L28, right below the setting we have for `default_library=static`; that way, all of the configuration is contained in one place, rather than split between pyproject.toml and the meson config

src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -508,7 +595,7 @@ TEST_P(ArrowTypeParameterizedTestFixture, NanoarrowIpcArrowTypeRoundtrip) {
 #endif
 
 std::string ArrowSchemaMetadataToString(const char* metadata) {
-  struct ArrowMetadataReader reader {};
+  struct ArrowMetadataReader reader;
```
WillAyd, line 598: Not saying this is right or wrong, but I was under the impression clang-tidy would complain about use of an uninitialized variable here

WillAyd (commented):
Just a few more minor comments but this lgtm. Nice work - this was a good deal of effort

python/pyproject.toml
paleolimbot: Hmm...I get `../meson.build:18:0: ERROR: Unknown options: "force-fallback-for"`. Did I translate that incorrectly into a default option?

src/nanoarrow/ipc/decoder_test.cc
paleolimbot: It's pretty good about this kind of thing (where the initialization happens in a function whose body it doesn't have access to). The problem I'm working around here and should fix properly (by pinning the version in the pre-commit) is that `clang-format` on homebrew (me!) is very recent but `clang-format` on ubuntu-latest (CI) is very old, and the formatting of `struct XXXX {};` differs between them.

python/pyproject.toml
WillAyd: Looks like it needs to be `force_fallback_for` (underscores instead of hyphens)

See also https://mesonbuild.com/Builtin-options.html#core-options

WillAyd (approved):

python/pyproject.toml
paleolimbot: Thanks!

PR #694, opened 2024-12-26
CI: Use gcovr instead of lcov

WillAyd:


ci/docker/ubuntu.dockerfile
```diff
@@ -20,7 +20,7 @@ ARG NANOARROW_ARCH
 FROM --platform=linux/${NANOARROW_ARCH} ubuntu:latest
 
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
-    locales git cmake r-base gnupg curl valgrind gfortran python3-venv python3-dev doxygen pandoc lcov \
+    locales git cmake r-base gnupg curl valgrind gfortran python3-venv python3-dev doxygen pandoc gcovr \
```
WillAyd, line 23: I haven't been able to get the docker compose run to work locally - getting an error 

```sh
/coverage.sh: line 88: gcovr: command not found
```

Although I am installing here. Is this not the right place?

WillAyd: Ah OK  I see - this registry this pulls from is in the .env file

paleolimbot: It's the right place! The image this is using in docker-compose.yml is a multi-platform manifest (to in theory make it easier for anybody to run locally, M1 or AMD), but the multi-platform upload happens in CI when the image update is merged to main. If you want to check for sure in CI you could update docker-compose.yml temporarily to point to the Dockerfile instead of the remote image (or you can confirm it works locally and we can merge with the failing CI...I trust you!).

The only way to do this keeping CI passing the whole time would be to update the image dockerfile, merge, rebuild + upload the images using a manual workflow-dispatch, then PR the changes to the scripts. Feel free to do it all in one go, though 🙂

paleolimbot (approved):
Thank you!

ci/docker/ubuntu.dockerfile
WillAyd: Ah OK cool - just pushed it up using the local image for now. You might want to take a look at the CI output if that's been of value to you in the past. The gcovr output is slightly different than what lcov provides

WillAyd:
That should be done now. FWIW I think gcovr could also replace coverage.py, which I _think_ is being used although I can't see where that is installed.

Not something we need to solve in this PR, but might make a good follow up just to use the one tool

paleolimbot (approved):
Thank you!

I think that coverage gets installed via pytest-cov maybe (probably on the ubuntu image dockerfile). We used to have coverage on the .pyx files, too, although I am less concerned about that since once we switch to nanobind or similar we'll get coverage back.

PR #697, opened 2024-12-27
feat(r): Support native creation of more numeric Arrow arrays from integer vectors

paleolimbot:
Before we had been punting to the arrow R package for this which wasn't necessary! This was exposed by a test that had been added for matrix support that wasn't skipped if arrow wasn't installed.

``` r
library(nanoarrow)
as_nanoarrow_array(1:10, schema = na_double())
#> <nanoarrow_array double[10]>
#>  $ length    : int 10
#>  $ null_count: int 0
#>  $ offset    : int 0
#>  $ buffers   :List of 2
#>   ..$ :<nanoarrow_buffer validity<bool>[null] ``
#>   ..$ :<nanoarrow_buffer data<double>[10][80 b]> `1 2 3 4 5 6 7 8 9 10`
#>  $ dictionary: NULL
#>  $ children  : list()
```

<sup>Created on 2024-12-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/r/src/as_array.c
+++ b/r/src/as_array.c
@@ -48,10 +48,23 @@ static void call_as_nanoarrow_array(SEXP x_sexp, struct ArrowArray* array,
 
 static void as_array_int(SEXP x_sexp, struct ArrowArray* array, SEXP schema_xptr,
                          struct ArrowSchemaView* schema_view, struct ArrowError* error) {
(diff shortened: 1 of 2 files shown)


amoeba (approved):
Nice, no notes.

Issue #701, opened 2025-01-07
How to iterate arrays with an offset? Possible ViewArrayAsBytes bug


WillAyd:
Assuming we have the following program:

```cpp
#include <iostream>

#include <nanoarrow/nanoarrow.hpp>

using namespace nanoarrow::literals;

int main() {
  struct ArrowSchema schema;
  ArrowSchemaInit(&schema);
  ArrowSchemaSetType(&schema, NANOARROW_TYPE_LARGE_STRING);

  ArrowArray array1;
  ArrowArrayInitFromSchema(&array1, &schema, nullptr);
  ArrowArrayStartAppending(&array1);
  ArrowArrayAppendString(&array1, "foo"_asv);
  ArrowArrayAppendString(&array1, "bar"_asv);
  ArrowArrayFinishBuildingDefault(&array1, nullptr);

  ArrowArray array2;
  ArrowArrayInitFromSchema(&array2, &schema, nullptr);
  ArrowArrayStartAppending(&array2);
  ArrowArrayAppendString(&array2, "hopefully this does not show"_asv);
  ArrowArrayAppendString(&array2, "this should not show either"_asv);  
  ArrowArrayAppendString(&array2, "this is a new string"_asv);
  ArrowArrayAppendString(&array2, "this is a newer string"_asv);
  ArrowArrayFinishBuildingDefault(&array2, nullptr);
  array2.offset = 2;

  struct ArrowArrayStream array_stream;
  ArrowBasicArrayStreamInit(&array_stream, &schema, 2);
  ArrowBasicArrayStreamSetArray(&array_stream, 0, &array1);
  ArrowBasicArrayStreamSetArray(&array_stream, 1, &array2);

  std::cout << "Here are my results with ViewArrayAsBytes" << std::endl;
  nanoarrow::ViewArrayStream array_stream_view(&array_stream);
  for (const auto &array : array_stream_view) {
    for (const auto val : nanoarrow::ViewArrayAsBytes<64>(&array)) {
      std::cout << std::string_view{(*val).data,
static_cast<size_t>((*val).size_bytes)} << std::endl;
    }
  }

  /*
  std::cout << "Here is with standard ArrowArrayView usage" << std::endl;
  nanoarrow::ViewArrayStream array_stream_view(&array_stream);
  nanoarrow::UniqueSchema array_stream_schema{};
  array_stream.get_schema(&array_stream, array_stream_schema.get());
  nanoarrow::UniqueArrayView array_view{};
[...]

PR #702, opened 2025-01-07
fix(cpp): Fix offset handling in ViewArrayAs Range Helpers

WillAyd:
closes https://github.com/apache/arrow-nanoarrow/issues/701

### DIFF
--- a/src/nanoarrow/hpp/view.hpp
+++ b/src/nanoarrow/hpp/view.hpp
@@ -80,7 +80,7 @@ struct RandomAccessRange {
   };
 
   const_iterator begin() const { return {offset, this}; }
-  const_iterator end() const { return {size, this}; }
+  const_iterator end() const { return {offset + size, this}; }
 };
 
 template <typename Next>
--- a/src/nanoarrow/hpp/view.hpp
+++ b/src/nanoarrow/hpp/view.hpp
@@ -65,6 +65,7 @@ class Maybe {
 template <typename Get>
 struct RandomAccessRange {
   Get get;
+  int64_t offset;
   int64_t size;
 
   using value_type = decltype(std::declval<Get>()(0));
@@ -79,7 +80,7 @@ struct RandomAccessRange {
   };
 
   const_iterator begin() const { return {0, this}; }
-  const_iterator end() const { return {size, this}; }
+  const_iterator end() const { return {size - offset, this}; }
 };
 
 template <typename Next>
@@ -157,6 +158,7 @@ class ViewArrayAs {
                 array_view->buffer_views[1].data.data,
                 array_view->offset,
             },
+            array_view->offset,
             array_view->length,
         } {}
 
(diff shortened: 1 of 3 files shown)


src/nanoarrow/hpp/view.hpp
```diff
@@ -157,6 +158,7 @@ class ViewArrayAs {
                 array_view->buffer_views[1].data.data,
                 array_view->offset,
             },
+            array_view->offset,
```
WillAyd, line 161: This is an easy fix but maybe not the API we want, since we are providing the offset to both the Range and the `Get` class template (?)

paleolimbot: This seems like the right thing to put here (at least one of Range or Get need it?)

src/nanoarrow/hpp/view.hpp
```diff
@@ -79,7 +80,7 @@ struct RandomAccessRange {
   };
 
   const_iterator begin() const { return {0, this}; }
-  const_iterator end() const { return {size, this}; }
+  const_iterator end() const { return {size - offset, this}; }
```
paleolimbot, line 83: I would have expected this to be:

```c
  const_iterator begin() const { return {offset, this}; }
  const_iterator end() const { return {offset + size, this}; }
```

...or to have it be `[0, size]` and to have the offset handled by `range->get()`.

src/nanoarrow/hpp/view_test.cc
```diff
@@ -135,3 +135,109 @@ TEST(NanoarrowHppTest, NanoarrowHppViewArrayStreamTest) {
...
+  ASSERT_EQ(ArrowArrayAppendBytes(array.get(), {{"qux"}, FixedSize}), NANOARROW_OK);
+  ASSERT_EQ(ArrowArrayFinishBuildingDefault(array.get(), nullptr), NANOARROW_OK);
+  array->offset = 2;
+
+  ArrowStringView expected[] = {"baz"_asv, "qux"_asv};
+  int i = 0;
+  for (auto slot : nanoarrow::ViewArrayAsFixedSizeBytes(array.get(), FixedSize)) {
+    EXPECT_EQ(slot, expected[i]);
+    i++;
+  }
```
paleolimbot, line 232: Can you use `::testing::ElementsAre("baz"_asv, "qux"_asv)`? (Not sure if the templates work for string views but they might)

WillAyd: Ah that's cool

src/nanoarrow/hpp/view.hpp
WillAyd: Not quite what you asked for but I think I've simplified this a bit by taking the offset out of the `Get` class and making it so that the range bounds start at the offset and end at `size`

WillAyd:
Hmm I think I might be missing something simple with the offest. So when we have an array of `{"foo", "bar", "baz", "qux"}` and an offset of 2, does that mean that the length of the array is supposed to be reported as 2 as well, or is it still 4?

paleolimbot (approved):
This looks great...thank you!

WillAyd:
Awesome - thanks for the help on review

PR #708, opened 2025-01-24
chore(ci): Use Github linux arm64 runners instead of the self-hosted ones

assignUser:
The linux arm64 runners are now GA for open source repos: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/


assignUser:
Yeah seems to happen consistently :/

paleolimbot (approved):
Hmm...there is a lot of QEMU segfaulting during compile on various projects I'm working on. I'll sort that out in a follow-up...thanks again!

Issue #633, opened 2024-09-23
ViewArrayAsBytes Specializations for Binary View Types


WillAyd:
Still need to be implemented in C++ header

PR #709, opened 2025-01-29
Implement ViewBinaryViewArrayAsBytes

WillAyd:
Closes https://github.com/apache/arrow-nanoarrow/issues/633

### DIFF
--- a/src/nanoarrow/hpp/view.hpp
+++ b/src/nanoarrow/hpp/view.hpp
@@ -234,6 +234,64 @@ class ViewArrayAsBytes {
   value_type operator[](int64_t i) const { return range_.get(i); }
 };
 
+class ViewBinaryViewArrayAsBytes {
+ private:
+  struct Get {
+    const uint8_t* validity;
+    const union ArrowBinaryView* inline_data;
+    const void** variadic_buffers;
+
+    internal::Maybe<ArrowStringView> operator()(int64_t i) const {
+      if (validity == nullptr || ArrowBitGet(validity, i)) {
+        const union ArrowBinaryView bv = inline_data[i];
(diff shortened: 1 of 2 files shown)


src/nanoarrow/hpp/view_test.cc
```diff
@@ -82,6 +82,58 @@ TEST(NanoarrowHppTest, NanoarrowHppViewArrayAsBytesTest) {
...
+
+  i = 0;
+  for (auto slot : nanoarrow::ViewBinaryViewArrayAsBytes(array_view.get())) {
+    if (i == 1 || i == 3) {
+      EXPECT_EQ(slot, nanoarrow::NA);
+    } else {
+      EXPECT_EQ(slot, expected[i]);
+    }
+    ++i;
+  }
```
paleolimbot, line 130: Should you increment the offset here and test again to make sure the offset is applied correctly? (I'm always paranoid that something isn't taking into account an offset 🙂 )

src/nanoarrow/hpp/view.hpp
```diff
@@ -234,6 +234,60 @@ class ViewArrayAsBytes {
   value_type operator[](int64_t i) const { return range_.get(i); }
 };
 
+class ViewBinaryViewArrayAsBytes {
```
paleolimbot, line 237: Should this be `ViewBinaryArray*View*AsBytes`?

paleolimbot (approved):
Two comments for consideration. Thank you!

src/nanoarrow/hpp/view.hpp
paleolimbot: Got it!

WillAyd:
Thanks @paleolimbot for the review

PR #710, opened 2025-02-04
feat: Implement LIST_VIEW and LARGE_LIST_VIEW support

WillAyd:


src/nanoarrow/common/inline_types.h
```diff
@@ -630,7 +636,8 @@ enum ArrowBufferType {
   NANOARROW_BUFFER_TYPE_DATA_OFFSET,
   NANOARROW_BUFFER_TYPE_DATA,
   NANOARROW_BUFFER_TYPE_VARIADIC_DATA,
-  NANOARROW_BUFFER_TYPE_VARIADIC_SIZE
+  NANOARROW_BUFFER_TYPE_VARIADIC_SIZE,
+  NANOARROW_BUFFER_TYPE_SIZE,
```
WillAyd, line 640: Not sure if we wanted to add a new buffer type for this or not - is there an official document upstream we follow for this?

src/nanoarrow/common/inline_array.h
```diff
@@ -752,19 +763,43 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+            ArrowArrayBuffer(array, 2), (int32_t)child_length - prev_offset));
+      }
       break;
+    }
     case NANOARROW_TYPE_LARGE_LIST:
+    case NANOARROW_TYPE_LARGE_LIST_VIEW: {
       child_length = array->children[0]->length;
       NANOARROW_RETURN_NOT_OK(
           ArrowBufferAppendInt64(ArrowArrayBuffer(array, 1), child_length));
+      if (private_data->storage_type == NANOARROW_TYPE_LARGE_LIST_VIEW) {
```
WillAyd, line 791: Generally not sure how much we want the implementation of the list views tied to their respective plain list types. Here I've kept the cases group together and specialize within the branch, but also happy to move the view types to their own case statements

src/nanoarrow/common/inline_array.h
```diff
@@ -290,18 +293,27 @@ static inline ArrowErrorCode _ArrowArrayAppendEmptyInternal(struct ArrowArray* a
...
-        // Append the current value at the end of the offset buffer for each element
-        NANOARROW_RETURN_NOT_OK(ArrowBufferReserve(buffer, size_bytes * n));
-
-        for (int64_t j = 0; j < n; j++) {
-          ArrowBufferAppendUnsafe(buffer, buffer->data + size_bytes * (array->length + j),
-                                  size_bytes);
+      case NANOARROW_BUFFER_TYPE_SIZE:
+        NANOARROW_RETURN_NOT_OK(ArrowBufferAppendFill(buffer, 0, size_bytes * n));
+        continue;
+      case NANOARROW_BUFFER_TYPE_DATA_OFFSET: {
```
WillAyd, line 299: The existing offset type is tightly bound to the idea that you have n+1 elements in the buffer to provide all open and closed intervals. The list views do not require that, so I've specialized this here. But we could alternatively make a new buffer enum member for offsets that aren't double ended (?)

src/nanoarrow/common/array_test.cc
```diff
@@ -1555,6 +1555,149 @@ TEST(ArrayTest, ArrayTestAppendToLargeListArray) {
...
+  EXPECT_EQ(ArrowArrayAppendEmpty(&array, 1), NANOARROW_OK);
+
+  // Make sure number of children is checked at finish
+  array.n_children = 0;
+  EXPECT_EQ(ArrowArrayFinishBuildingDefault(&array, &error), EINVAL);
+  EXPECT_STREQ(ArrowErrorMessage(&error),
+               "Expected 1 child of list_view array but found 0 child arrays");
+  array.n_children = 1;
+
+  // Make sure final child size is checked at finish
```
WillAyd, line 1591: This was just copied from the existing list types, where it is trivial to check the last offsets against the size of the array. The list view check I don't think is as simplistic - maybe we should just remove this from the validation?

src/nanoarrow/common/array_test.cc
```diff
@@ -1555,6 +1555,149 @@ TEST(ArrayTest, ArrayTestAppendToLargeListArray) {
 #endif
 }
 
+TEST(ArrayTest, ArrayTestAppendToListViewArray) {
```
WillAyd, line 1558: These tests were copy/paste from their list counterparts; in a follow up commit or PR I'd like to try and parametrize these

WillAyd:
The current implementation here is really simplistic, and just continually appends children elements to the end of the array rather than trying to consolidate them in any way

src/nanoarrow/common/array_test.cc
paleolimbot: I think we probably want the check, but because the check requires iterating along every item it should be checked in the `ValidateFull` rather than the `ValidateDefault`. That check is important for IPC because it prevents (if done correctly) out-of-bounds buffer access!

src/nanoarrow/common/inline_array.h
paleolimbot: I think we want a new buffer type for this one because it has different rules than the existing `DATA_OFFSET` buffer (maybe `NANOARROW_BUFFER_TYPE_VIEW_OFFSET`?)

src/nanoarrow/common/inline_array.h
```diff
@@ -772,6 +787,59 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+      }
+
+      const int32_t current_offset = (int32_t)private_data->list_view_offset;
+      int32_t offset;
+      if (current_offset == 0) {
+        offset = 0;
+      } else {
+        struct ArrowBuffer offsets_buf, sizes_buf;
+        offsets_buf = *ArrowArrayBuffer(array, 1);
+        sizes_buf = *ArrowArrayBuffer(array, 2);
```
paleolimbot, line 803: I don't think you want to dereference a buffer pointer here (that will mean that the changes that you make to `offsets_buf` won't be syncrhonized back to the array!). I think `struct ArrowBuffer* offsets_buf` will work here?

src/nanoarrow/common/inline_array.h
```diff
@@ -772,6 +787,59 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+    case NANOARROW_TYPE_LARGE_LIST_VIEW: {
+      child_length = array->children[0]->length;
+      const int64_t current_offset = private_data->list_view_offset;
+      int64_t offset;
+      if (current_offset == 0) {
+        offset = 0;
+      } else {
+        struct ArrowBuffer offsets_buf, sizes_buf;
+        offsets_buf = *ArrowArrayBuffer(array, 1);
+        sizes_buf = *ArrowArrayBuffer(array, 2);
```
paleolimbot, line 827: I think these should be pointers as well

src/nanoarrow/common/inline_types.h
paleolimbot: Nope...this is nanoarrow native! I think a new buffer type is a good idea (for the view offsets as well!)

src/nanoarrow/common/inline_types.h
```diff
@@ -864,6 +871,9 @@ struct ArrowArrayPrivateData {
 
   // Size of each variadic buffer in bytes
   int64_t* variadic_buffer_sizes;
+
+  // The current offset used to build list views
+  int64_t list_view_offset;
```
paleolimbot, line 876: Would it be possible to just use the length of the child array instead of add a parameter here?

paleolimbot (commented):
Thank you for taking this on! On the next round I will give this a check-out + run through 🙂

src/nanoarrow/common/inline_array.h
WillAyd: Happy to change these but there shouldn't be any updates to the original data through these. Can also add const

paleolimbot: Got it! I think it's still a good change for future code readers (the `ArrowBuffer` should really never be copied by value!)

src/nanoarrow/common/inline_types.h
WillAyd: What we need to keep track of somehow is the last valid offset preceding the value currently being appended, so we can calculate the `current_child_length - previous_child_length` to get the size of the current element.

The private data is necessary in the current implementation because we aren't keeping track of the most recently valid offset when Null/Empty data is being appended. Those types of data are currently inserting an offset and size of 0 for simplicity.

We can certainly change that append to keep repeating the most recently offset value with a size of 0 for empty elements, but I think that would either still require us to use private data OR backtrack to the last element contained within a `NANOARROW_BUFFER_TYPE_VIEW_OFFSET` and repeat that as many times as needed during `_ArrowArrayAppendEmptyInternal`

WillAyd:
Not a problem. There is literally no rush on this at all. Doesn't need to be over the weekened

src/nanoarrow/common/inline_array.h
```diff
@@ -772,6 +792,31 @@ static inline ArrowErrorCode ArrowArrayFinishElement(struct ArrowArray* array) {
...
+      if (child_length > INT32_MAX) {
+        return EOVERFLOW;
+      }
+
+      const int32_t last_valid_offset = (int32_t)private_data->list_view_offset;
+      NANOARROW_RETURN_NOT_OK(
+          ArrowBufferAppendInt32(ArrowArrayBuffer(array, 1), last_valid_offset));
+      NANOARROW_RETURN_NOT_OK(ArrowBufferAppendInt32(
+          ArrowArrayBuffer(array, 2), (int32_t)child_length - last_valid_offset));
+      private_data->list_view_offset = (int32_t)child_length;
```
paleolimbot, line 806:
```suggestion
      private_data->list_view_offset = child_length;
```

paleolimbot (commented):
Apologies for the turnaround here...great work and thank you for working on this!

A few comments here, and I think there are a few more test categories that need implementing for the new type:

ArrayView checks:

https://github.com/apache/arrow-nanoarrow/blob/d36829207c1920e2871fe7cc58d944aaefcc3784/src/nanoarrow/common/array_test.cc#L2804

SchemaInit checks:

https://github.com/apache/arrow-nanoarrow/blob/d36829207c1920e2871fe7cc58d944aaefcc3784/src/nanoarrow/common/schema_test.cc#L130

SchemaViewInit checks:

https://github.com/apache/arrow-nanoarrow/blob/d36829207c1920e2871fe7cc58d944aaefcc3784/src/nanoarrow/common/schema_test.cc#L1309

src/nanoarrow/common/inline_array.h
```diff
@@ -310,7 +313,23 @@ static inline ArrowErrorCode _ArrowArrayAppendEmptyInternal(struct ArrowArray* a
...
+        if (array->length == 0) {
+          const int64_t zero_val = 0;
+          for (int64_t j = 0; j < n; j++) {
+            ArrowBufferAppendUnsafe(buffer, &zero_val, size_bytes);
+          }
+        } else {
+          for (int64_t j = 0; j < n; j++) {
+            ArrowBufferAppendUnsafe(
+                buffer, buffer->data + size_bytes * (array->length + j - 1), size_bytes);
+          }
```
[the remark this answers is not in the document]
WillAyd: Yes it can be. The only advantage I think to doing it this way is it aligns better with how the current LIST types work, possibly making the tests easier to parametrize. 

We can cross that bridge when we get there though; for now setting to 0 should work fine

src/nanoarrow/common/array.c
```diff
@@ -1157,9 +1175,31 @@ static int ArrowArrayViewValidateDefault(struct ArrowArrayView* array_view,
           return EINVAL;
         }
       }
+
+      if ((array_view->storage_type == NANOARROW_TYPE_LIST_VIEW) &&
```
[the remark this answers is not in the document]
WillAyd: I might be misunderstanding. To clarify further what I was thinking, lists and strings check the first/last `offset` in default validation, but have no size checking, given they do not have a dedicated size buffer. The list view repeats that same check for the `offset` buffer and extends it over to the `size` buffer.

So do you want to keep the first/last check of offsets for the list view in the default validation and move the size check to full validation? Or should we eliminate the size check altogether?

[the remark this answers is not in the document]
paleolimbot: I also might be misunderstanding 🙂 

You do validate both the offsets and the sizes in the full validation check, and my point is that here the first and last offsets and/or sizes are not more or less important than any of the others so I think it makes sense to do them all once, in the same place.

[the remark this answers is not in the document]
WillAyd: > You do validate both the offsets and the sizes in the full validation check

I think you meant to say _default validation check_ here, in which case understood and happy to move those checks to full. The only nuance there again is that it probably will make the test parametrization more complicated, as the list types keep the first/last offset checks in their default validation implementation and test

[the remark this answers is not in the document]
paleolimbot: Got it! I think the current diff looks great, thanks for bearing with me on that one!

paleolimbot (approved):
Apologies for the turnaround again, but thank you!

PR #712, opened 2025-02-08
fix(docs): fix download link in the README

wgtmac:
The download link in the README is wrong and the version is outdated.


paleolimbot (approved):
Thank you! (And apologies if you tried that and it didn't work off the bat!)

wgtmac:
Thanks for merging it! Yes I am trying to use it in iceberg-cpp :)

PR #717, opened 2025-03-09
feat: Add support for Decimal32/64 to R package

paleolimbot:
This PR adds support for decimal32, 64, and 256, and migrates support for decimal128 to use nanoarrow's internals rather than call Arrow. This removes the last dependency on arrow for conversion from Arrow -> R, for which some (undiagnosed as of yet) bug results in very slow conversions when many batches are involved ( https://github.com/apache/arrow-adbc/issues/2508 ). Decimal conversions show up in database land quite lot because decimal types are common in database output (notably: Snowflake).

To support the types that the arrow package does not yet support (Decimal32, 64, and 256), this PR also implements conversion from R -> Arrow for decimal types. This conversion is a little hacky because it involves strings, but the basic idea is to use the fact that R has all the functionality needed (rounding, appending adding zeroes, etc.). This code path is probably rare because one has to explicitly opt-in to a decimal type but I put it in because it's helpful to ensure we (and others) can create test data or specific typed input for a database append if required.

``` r
library(nanoarrow)

# Can create a decimal array from integer or double
(dec_array <- as_nanoarrow_array(-10:10, schema = na_decimal32(9, 2)))
#> <nanoarrow_array decimal32(9, 2)[21]>
#>  $ length    : int 21
#>  $ null_count: int 0
#>  $ offset    : int 0
#>  $ buffers   :List of 2
#>   ..$ :<nanoarrow_buffer validity<bool>[null] ``
#>   ..$ :<nanoarrow_buffer data<decimal32>[21][84 b]> `-1000 -900 -800 -700 -6...`
#>  $ dictionary: NULL
#>  $ children  : list()

# Default conversion to R is a double, which matches what arrow does
# (and what nanoarrow did before this change for decimal128)
str(convert_array(dec_array))
#>  num [1:21] -10 -9 -8 -7 -6 -5 -4 -3 -2 -1 ...

# One can also convert to character(), which keeps better fidelity for
# out-of-range types.
str(convert_array(dec_array, character()))
#>  chr [1:21] "-10.00" "-9.00" "-8.00" "-7.00" "-6.00" "-5.00" "-4.00" ...

# Print method for decimal buffers!
dec_array$buffers[[2]]
#> <nanoarrow_buffer data<decimal32>[21][84 b]> `-1000 -900 -800 -700 -600 -500...`
```

<sup>Created on 2025-03-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


r/tests/testthat/test-type.R
```diff
@@ -74,7 +74,18 @@ test_that("timestamp type passes along timezone parameter", {
 })
 
 test_that("decimal types pass along precision and scale", {
+  schema <- na_decimal32(12, 10)
```
amoeba, line 77: How are you thinking about validation with these helpers? This is technically an invalid precision but nanoarrow currently passes it through. I think other na_* also don't validate inputs.

r/NAMESPACE
```diff
@@ -63,7 +63,6 @@ S3method(as_nanoarrow_schema,nanoarrow_schema)
 S3method(as_nanoarrow_schema,nanoarrow_vctr)
 S3method(c,nanoarrow_vctr)
 S3method(convert_array,default)
-S3method(convert_array,double)
```
amoeba, line 66: I had initially thought this might be breaking but I see the way you have things hooked up here `convert_array(..., to = double())` works just like before. 👍

amoeba, line 66: I had initially thought this might be breaking but I see the way you have things hooked up here `convert_array(..., to = double())` works just like before. 👍

amoeba (approved):
Nice, LGTM. Built, tested, and checked locally.

Do you have a PR up to update the implementation status page?

r/tests/testthat/test-type.R
paleolimbot: That's a great point. The validation happens in C (`ArrowSchemaSetTypeDecimal()`) and now is a good as a time as any to fix it!

r/NAMESPACE
paleolimbot: Yes, we have a funny two-pass system to ensure we skip S3 dispatch wherever possible (but we still need it to work so that other packages can implement custom behaviour without strictly taking on a nanoarrow dependency).

amoeba:
Right, I realize that was a bit confusing. The current [implementation status page](https://arrow.apache.org/docs/status.html#data-types) shows that nanoarrow doesn't support Decimal32/64. Should that get updated?

amoeba:
Oh yes, it is. 🤦. Don't mind me!

PR #719, opened 2025-03-10
feat: Improve shared linkage support

### LINKED ISSUE #495: Export symbols from C library when compiling as a shared library on Windows

### LINKED ISSUE #718: The c library doesn't export any symbols
On windows (at least using msvc) symbols are not exported by default, contrary to linux/macos/*nix (actually depends on `-fvisibility` afaik).

To still use a nanoarrow dll on windows, there are two options:

 - Using `-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE`
 - Using [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html)

It would be nice to use the second approach to explicitly flag symbols which are meant for external use as such.

paleolimbot:
This PR makes linking to the usual targets static linkage and adds a paired `_shared` target for users that want to opt in to shared nanoarrow linkage. This is intended to mirror what ADBC and Arrow do (with separated `_shared` and `_static` targets), except nanoarrow already had some existing targets that were intended to be used with static linkage anyway, so those target names were left as is.

This PR also adds the appropriate import/export attribute for Windows (most of the changed lines in this PR).

While I had my Windows machine fired up I also fixed a few compiler warnings, and I added a Windows and MacOS run for the `examples` workflow.

Closes #495, closes #718


WillAyd:
Love the individual commit names - really takes you along for the ride as a reader. Hope all is well!

src/nanoarrow/nanoarrow.h
```diff
@@ -139,6 +139,20 @@
...
+#else
+#define NANOARROW_DLL __declspec(dllimport)
+#endif  // defined(NANOARROW_EXPORT_DLL)
+#elif !defined(NANOARROW_DLL)
+#if __GNUC__ >= 4
+#define NANOARROW_DLL __attribute__((visibility("default")))
+#else
+#define NANOARROW_DLL
+#endif  // __GNUC__ >= 4
+#endif
```
m-kuhn, line 154: To avoid keeping this up to date, I would recommend the usage of https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html which creates this code via cmake.

```
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
generate_export_header(nanoarrow)

install(FILES
 ${PROJECT_BINARY_DIR}/nanoarrow_export.h DESTINATION ${INCLUDE_INSTALL_DIR}
)
```

```
class NANOARROW_EXPORT SomeClass { ... }
```

I do think that with this we get away with one exported target (no `nanoarrow_shared` target, or is there a reason we want to install static and shared in parallel?) and less boilerplate code to maintain

CMakeLists.txt
```diff
@@ -416,6 +452,13 @@ foreach(target
     # Ensure NANOARROW_DEBUG is defined for debug builds
     target_compile_definitions(${target} PUBLIC "$<$<CONFIG:Debug>:NANOARROW_DEBUG>")
 
+    # Ensure NANOARROW_DLL is set when building and linking to the library
+    if(target MATCHES "_shared$")
```
assignUser, line 456: Targets have a `TYPE` property that you could check here instead of the name: https://cmake.org/cmake/help/latest/prop_tgt/TYPE.html

assignUser (commented):
I have some thoughts about target naming and such but I need to take some more time and think about it and look some things up. 

One thing is that I see a bunch of repeating stuff that should probably be refactored in a functions, but that's true in general (see L436 onwards) so that could be done as a follow up.

In general I think this is mergable so don't wait for me ^^

src/nanoarrow/nanoarrow.h
m-kuhn: if cross-build-system support is a concern, I am with you the current approach is better 👍

.github/workflows/examples.yaml
```diff
@@ -107,15 +113,18 @@ jobs:
...
           ./build.sh
-          for dir in scratch/build*; do
-            ./${dir}/minimal_cpp_app;
-            echo ;
-          done
+          ./run.sh
 
       - name: Meson example
+        # This example doesn't seem to work on Windows
+        if: matrix.config.label != 'windows'
```
paleolimbot, line 126: @WillAyd  I'll open an issue for this! I think it works but needs some environment-on-ci debugging!

assignUser:
I think you should change `nanoarrow` explicitly to `nanoarrow_static` to be consistent and then add a `nanoarrow::nanoarrow` alias to it (I assume static will be the default? Or you could base that on `BUILD_SHARED_LIBS` like this
```cmake
if(BUILD_SHARED_LIBS)
  add_library(nanoarrow::nanoarrow ALIAS nanoarrow_shared)
else()
  add_library(nanoarrow::nanoarrow ALIAS nanoarrow_static)
endif()
```
This is mostly relevant when nanoarrow is built as a subproject. You could also add an option that allows toggling the building of either version (gflags does that). That way I don't have to compile everything twice if I only care about static.

For packaging the shared bit could be made a component that also changes which target gets the main alias. (again something gflags does `find_package(gflags COMPONENTS shared)`

nit: You should be able to create an INTERFACE target for the sources and use that instead of variables: https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SOURCES.html

I haven't used an interface target but it's a nice solution to avoid raw variables (which I am always happy about...)

m-kuhn:
Thanks for all the work!

For my understanding, for downstream libraries that link to nanoarrow, this will mean to be used like this:

```
target_link_libraries(main PRIVATE $<IF:$<TARGET_EXISTS:nanoarrow::nanoarrow_shared>,nanoarrow::nanoarrow_shared,nanoarrow::nanoarrow_static>)
```

and there's no generic target that takes whatever has been built?

```
target_link_libraries(main PRIVATE nanoarrow::nanoarrow)
```

I'm asking as I never had the need to build static AND shared libraries side-by-side but had the need to build either one or the other depending on the target platform (e.g. shared for linux, static for ios) and for these cases the short version is nice in terms of readability.

PR #721, opened 2025-03-14
Add skips for Suggests packages

MichaelChirico:
This completes the existing set of `skip_if_not_installed()` directives


paleolimbot (approved):
Thank you!

PR #723, opened 2025-03-14
Add missing dplyr dependency

### LINKED ISSUE #722: 'dplyr' is used in optional test, but omitted from DESCRIPTION
https://github.com/apache/arrow-nanoarrow/blob/cf0518724fa2147eb38cbf5bf420e94605479f3a/r/tests/testthat/test-ipc.R#L335

I would call {dplyr} a `Suggests` dependency in this case, since you are calling `requireNamespace("dplyr")`. One could cook up a scenario where a user unexpectedly winds up with `isNamespaceLoaded("dplyr")` not matching their expectations as a result.

MichaelChirico:
Closes #722

### DIFF
--- a/r/DESCRIPTION
+++ b/r/DESCRIPTION
@@ -26,6 +26,7 @@ Suggests:
     arrow (>= 9.0.0),
     bit64,
     blob,
+    dplyr,
     hms,
     jsonlite,
     rlang,


paleolimbot (approved):
Thank you!

PR #731, opened 2025-03-20
feat: Use hidden symbol linkage in Meson configuration

WillAyd:
This is a follow up to #719 enabling hidden symbol visibility in the Meson configuration


paleolimbot (approved):
Thank you! Meson to the rescue again!

PR #736, opened 2025-03-28
fix: Fix potential integer overflows

### LINKED ISSUE #735: Possible integer overflow
Found by a static analyser. Not sure whether it's a real issue, but there are various places where an operation is performed on `int32_t` types and assigned to a `int64_t` variable:

https://github.com/apache/arrow-nanoarrow/blob/08a64ff44fce354dfca9c011443dbe555f3a8715/src/nanoarrow/common/schema.c#L1356

> The value of an arithmetic expression `schema_view->fixed_size * 8` is a subject to overflow because its operands are not cast to a larger data type before performing arithmetic

https://github.com/apache/arrow-nanoarrow/blob/08a64ff44fce354dfca9c011443dbe555f3a8715/src/nanoarrow/common/inline_array.h#L1220-L1221

https://github.com/apache/arrow-nanoarrow/blob/08a64ff44fce354dfca9c011443dbe555f3a8715/src/nanoarrow/common/inline_array.h#L1260-L1261

> The value of an arithmetic expression `offsets_view->data.as_int32[i + 1] - offsets_view->data.as_int32[i]` is a subject to overflow because its operands are not cast to a larger data type before performing arithmetic

https://github.com/apache/arrow-nanoarrow/blob/08a64ff44fce354dfca9c011443dbe555f3a8715/src/nanoarrow/ipc/decoder.c#L1084

https://github.com/apache/arrow-nanoarrow/blob/08a64ff44fce354dfca9c011443dbe555f3a8715/src/nanoarrow/ipc/decoder.c#L1206

> The value of an arithmetic expression `decoder->header_size_bytes - prefix_size_bytes` is a subject to overflow because its operands are not cast to a larger data type before performing arithmetic

Gumix:
This patch extends int32 operands to 64 bits in arithmetic expressions where the result is assigned to int64 members, or compared with them.

Closes #735

### DIFF
--- a/src/nanoarrow/common/inline_array.h
+++ b/src/nanoarrow/common/inline_array.h
@@ -1218,7 +1218,7 @@ static inline struct ArrowStringView ArrowArrayViewGetStringUnsafe(
     case NANOARROW_TYPE_BINARY:
       view.data = data_view + offsets_view->data.as_int32[i];
       view.size_bytes =
-          offsets_view->data.as_int32[i + 1] - offsets_view->data.as_int32[i];
+          (int64_t)offsets_view->data.as_int32[i + 1] - offsets_view->data.as_int32[i];
       break;
     case NANOARROW_TYPE_LARGE_STRING:
     case NANOARROW_TYPE_LARGE_BINARY:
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
Thank you!

PR #739, opened 2025-03-31
feat: Use feature options instead of boolean in Meson

WillAyd:
The feature option is pretty similar to the previous boolean setting, but allows three states of enabled, disabled, and auto.

auto by default is disabled, but you can set `-Dauto_features=enabled` to have that toggle to enabled. By doing so, you can let users either start with nothing and opt-in to everything, or opt-in by default and then opt-out of the components they do not want

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,11 +159,11 @@ cd builddir
 After setting up your project, be sure to enable the options you want:
 
 ```sh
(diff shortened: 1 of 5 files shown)


paleolimbot (commented):
Cool! (Looks like there's one more maybe in the docs builder script that needs updating?)

PR #744, opened 2025-04-22
clean: Use Meson disabler objects instead of conditions

WillAyd:
This is a minor usability enhancement to use Meson's `disabler()` object rather than repeating a lot of if statements to enable/disable tests


paleolimbot (approved):
Thanks!

PR #750, opened 2025-05-02
fix: Fix valgrind error suggesting use of an uninitialized value

### LINKED ISSUE #749: Meson Sanitzer error on MaybeTest
Noted in CI on #746.

Ref: https://github.com/apache/arrow-nanoarrow/actions/runs/14802193206/job/41563284683?pr=746

```
stderr:
  ==4371== Memcheck, a memory error detector
  ==4371== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
  ==4371== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
  ==4371== Command: /home/runner/work/arrow-nanoarrow/arrow-nanoarrow/_meson_builddir/utils-test
  ==4371== 
  ==4371== Conditional jump or move depends on uninitialised value(s)
  ==4371==    at 0x121DA1: CmpHelperNE<nanoarrow::internal::Maybe<int>, int> (gtest.h:1457)
  ==4371==    by 0x121DA1: MaybeTest_ConstructionAndConversion_Test::TestBody() (utils_test.cc:790)
  ==4371==    by 0x16DFBE: HandleSehExceptionsInMethodIfSupported<testing::Test, void> (gtest.cc:2612)
  ==4371==    by 0x16DFBE: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2648)
  ==4371==    by 0x15D5E5: Run (gtest.cc:2687)
  ==4371==    by 0x15D5E5: testing::Test::Run() (gtest.cc:2677)
  ==4371==    by 0x15D82C: testing::TestInfo::Run() (gtest.cc:2836)
  ==4371==    by 0x15FEA1: testing::TestSuite::Run() [clone .part.0] (gtest.cc:3015)
  ==4371==    by 0x163AFD: Run (gtest.cc:2969)
  ==4371==    by 0x163AFD: testing::internal::UnitTestImpl::RunAllTests() (gtest.cc:5920)
  ==4371==    by 0x16E586: HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (gtest.cc:2612)
  ==4371==    by 0x16E586: bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (gtest.cc:2648)
  ==4371==    by 0x15D90F: testing::UnitTest::Run() (gtest.cc:5484)
  ==4371==    by 0x11D463: RUN_ALL_TESTS (gtest.h:2317)
  ==4371==    by 0x11D463: main (gtest_main.cc:64)
  ==4371==  Uninitialised value was created by a stack allocation
  ==4371==    at 0x121D2E: MaybeTest_ConstructionAndConversion_Test::TestBody() (utils_test.cc:780)
  ==4371== 
```

The code is here:

https://github.com/apache/arrow-nanoarrow/blob/4c1e484ca7d575250444e0a8eee5884e13489104/src/nanoarrow/common/utils_test.cc#L780-L809

paleolimbot:
I'm not sure that anything was actually wrong here, but writing it differently helped valgrind understand what's going on. I also added a pin for meson since the latest version results in failing CI when running valgrind (and opened #753 to track removing the pin).

Closes #749.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -127,7 +127,9 @@ jobs:
           path: build/Testing/Temporary/MemoryChecker.*.log
 
   meson-build:
-    runs-on: ubuntu-latest
+    # Can't quite be ubuntu-latest yet
+    # https://github.com/apache/arrow-nanoarrow/issues/753
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -139,6 +141,8 @@ jobs:
           sudo apt-get update && sudo apt-get install -y gcovr ninja-build valgrind
 
       - name: Install meson
+        # Meson 1.8.0 breaks build
+        # https://github.com/apache/arrow-nanoarrow/issues/753
         run: |
           python3 -m pip install "meson<1.8.0"
 
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -126,8 +126,7 @@ jobs:
           name: nanoarrow-memcheck
           path: build/Testing/Temporary/MemoryChecker.*.log
 
-  verify-meson:
-    name: meson-build
+  meson-build:
     # Workaround until https://github.com/apache/arrow-nanoarrow/issues/663 is solved
     # (after which we can use ubuntu-latest)
     runs-on: ubuntu-22.04
@@ -143,7 +142,7 @@ jobs:
 
       - name: Install meson
         run: |
-          python3 -m pip install meson
+          python3 -m pip install "meson<1.8.0"
 
       - name: Cache Arrow C++ Build
         id: cache-arrow-build
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -127,9 +127,7 @@ jobs:
           path: build/Testing/Temporary/MemoryChecker.*.log
 
   meson-build:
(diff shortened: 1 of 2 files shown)


WillAyd:
I believe that error message occurs when you build libraries with sanitizer support but the executable you are using is not built with sanitizer support (ex: I see it all the time when loading shared libraries through a Python process that is not build with sanitizer flags)

For LD_PRELOAD, I would usually do something like `LD_PRELOAD="$(gcc -print-file-name=libasan.so)" <the_executable> ...` to populate the ASAN runtime, although in this case we could also consider not running ASAN and Valgrind together. I vaguely recall seeing a FAQ stating those should not be run together; if I can find it I will link back

WillAyd:
OK here it is:

https://bugs.kde.org/show_bug.cgi?id=380037#c1

WillAyd:
Hmm it looks like we do set `-Db_sanitize=none` in the valgrind job:

https://github.com/apache/arrow-nanoarrow/blob/4c1e484ca7d575250444e0a8eee5884e13489104/ci/scripts/build-with-meson.sh#L84

So the fact that those two are running together is unexpected. Meson just recently released version 1.8.0 with a major change to the arguments that `-Db_sanitize=` accepts, so I wonder if this is a regression with that version

PR #755, opened 2025-05-15
fix: Fix ignored offset in ArrowArrayViewGetIntervalUnsafe

### LINKED ISSUE #754: Interval Types do not respect Array View Offsets
I noticed with some user reported bugs in my own library that the `ArrowArrayViewGetIntervalUnsafe` does not work properly when you have an array view that carries an offset. The offset is completely ignored, so you end up taking data from the beginning of the array when you iterate across multiple batches

WillAyd:
closes #754


WillAyd:
I am able to reproduce the Python failures on main, so I am guessing they have something to do with dependency versions

WillAyd:
Confirmed the Python regression is caused by Cython 3.1.0. See also https://github.com/apache/arrow-nanoarrow/issues/756

paleolimbot (approved):
Thank you!

PR #759, opened 2025-05-30
fix(python): Ensure source distribution builds with Cython 3.1.0

### LINKED ISSUE #756: Python Package fails with Cython 3.1.0
The Python package fails to appropriate install from an sdist when using Cython 3.1.0. Importing nanoarrow will throw an error like:

```
python/tests/test_visitor.py:20: in <module>
    import nanoarrow as na
../../miniforge3/envs/scratchpad/lib/python3.13/site-packages/nanoarrow/__init__.py:28: in <module>
    from nanoarrow.c_array import c_array_from_buffers, c_array
../../miniforge3/envs/scratchpad/lib/python3.13/site-packages/nanoarrow/c_array.py:571: in <module>
    _types.BOOL: "_append_using_buffer_builder",
E   AttributeError: module 'nanoarrow._types' has no attribute 'BOOL'
```

paleolimbot:
In Cython 3.1.0, cpdef enums are no longer copied to the parent module. We're using the types module a bit like an enum and kind of want this behaviour, so this PR implements the workaround suggested in the Cython docs: https://cython.readthedocs.io/en/stable/src/userguide/language_basics.html

Closes #756, superceeds #757 as long as Will is OK with this!

### DIFF
--- a/python/src/nanoarrow/_types.pyx
+++ b/python/src/nanoarrow/_types.pyx
@@ -24,6 +24,11 @@ from struct import calcsize
 from sys import byteorder as sys_byteorder
 
 
+# The default changed in Cython 3.1.0 such that the member of an
+# enum are no longer automatically copied to the parent module.
+globals().update(getattr(CArrowType, '__members__'))
+
+
 cdef equal(int type_id1, int type_id2):
     """Check if two type identifiers are equal
 


WillAyd (approved):
lgtm

PR #763, opened 2025-05-30
fix: Make auto_features more user friendly in Meson

WillAyd:
Right now if a user wants to user the `-Dauto_features=enabled` option, some of the more obscure features have to be disabled in the majority of cases.

Rather than failing the build, I think its better to warn for some of the lesser used or available features (cuda, metal, arrow)

### DIFF
--- a/ci/scripts/build-with-meson.sh
+++ b/ci/scripts/build-with-meson.sh
@@ -69,55 +69,39 @@ function main() {
     meson configure \
           -Dbuildtype=debugoptimized \
           -Db_sanitize="address,undefined" \
-          -Dtests=enabled \
-          -Dtests_with_arrow=enabled \
-          -Dipc=enabled \
-          -Ddevice=enabled \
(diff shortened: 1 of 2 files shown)


WillAyd:
@paleolimbot I vaguely recall you doing some update to the cached Arrow C++ version - was that in this project or another repo?

I _think_ the current CI failure is unrelated to this change, but instead may come from linking against a version of Arrow that was compiled with gcc 13 (the default on ubuntu 24.04) while our current Meson image is limited to ubuntu 22.04 with gcc 11

WillAyd:
I think this will be fixed after https://github.com/apache/arrow-nanoarrow/pull/766

WillAyd:
I just realized on main that we never set the `-Dtests_with_arrow` option in CI, so I think that explains why these don't show on main. I'll take a look at setting that as a precursor

paleolimbot (approved):
Thank you!

Issue #753, opened 2025-05-03
Remove version pins required for meson script to pass


paleolimbot:
We now have two pins for the meson-build job:

- meson < 1.8.0 because something about running the valgrind job also runs with sanitizers by accident?)
- ubuntu-22.04 because of something to do with coverage. This doesn't seem to have been fixed by the PR linked in https://github.com/apache/arrow-nanoarrow/issues/663 ?

I imagine the first point will be fixed by a meson version update...not sure about the second one! Our coverage script could potentially be simpler which may help.

WillAyd:
I believe the root problem with 1.8.0 is this issue:

https://github.com/mesonbuild/meson/issues/14558

Will track upstream

PR #766, opened 2025-06-02
ci: Try bumping Meson to ubuntu-latest

WillAyd:


WillAyd:
OK all green - this works around the issue in https://github.com/apache/arrow-nanoarrow/issues/753

paleolimbot (approved):
Thanks!

Issue #773, opened 2025-06-09
Flatcc not linked properly on clang-cl when built with meson


paleolimbot:
From https://github.com/mesonbuild/wrapdb/pull/2150

https://github.com/mesonbuild/wrapdb/actions/runs/15544016418/job/43761308181?pr=2150#step:6:5173

```
[35/97] Linking target subprojects/arrow-nanoarrow-4bfd77c82793eb91c804d6f893f7ab324e0105a0/nanoarrow_ipc.dll
FAILED: subprojects/arrow-nanoarrow-4bfd77c82793eb91c804d6f893f7ab324e0105a0/nanoarrow_ipc.dll subprojects/arrow-nanoarrow-4bfd77c82793eb91c804d6f893f7ab324e0105a0/nanoarrow_ipc.pdb 
[...]

PR #774, opened 2025-06-10
ci: Clean up existing Meson CI job and add Windows coverage

WillAyd:


WillAyd:
This should also fix https://github.com/apache/arrow-nanoarrow/issues/773

WillAyd:
Ah OK...if our chatbot overlords are correct then I think I understand the problem.

PR #776, opened 2025-06-10
Force static compilation on Windows with Meson builds

WillAyd:


paleolimbot (approved):
Thank you!

WillAyd:
Hmm I still think something is wrong here. CI shows an error, although the job itself doesn't fail (?)

From reading up on it some more, I think the problem is that our nanoarrow testing dependency links libraries against both libnanoarrow and libnanoarrow_testing. The former can be shared while the latter is now forced to be static.

The problem with the former being shared is that it would then require `NANOARROW_BUILD_DLL` to be defined, but the latter cannot have this defined (or else the symbols in a target are generated with dllimport).

Maybe for now the best solution is to just force static libraries across the board on Windows?

meson.build
```diff
@@ -46,8 +46,21 @@ if get_option('buildtype') in ['debug', 'debugoptimized']
...
+    # static libraries for now. See also
+    # https://github.com/apache/arrow-nanoarrow/pull/776#issuecomment-2962559614
+    target_type = 'static_library'
+else
+    target_type = 'library'
+endif
+
 nanoarrow_dep_args = []
-if host_machine.system() == 'windows' and get_option('default_library') == 'shared'
+if (host_machine.system() == 'windows'
```
WillAyd, line 60: The formatting enforced by meson-format here is a bit unfortunate, but will be fixed when version 1.9.0 gets released

PR #778, opened 2025-06-11
fix: Linking to nanoarrow-testing-shared on Windows

paleolimbot:
We missed one issue in the last PR for this to work out of the box. I imagine people put std::string in headers all the time and there's another workaround here (or maybe people just ignore this warning all the time)...perhaps a better long-term solution would be to just expose a C interface here.


WillAyd (approved):
niceeeeeee!

PR #779, opened 2025-06-11
ci: Set Werror in Windows Meson configuration

WillAyd:


src/nanoarrow/ipc/reader_test.cc
```diff
@@ -115,8 +115,13 @@ TEST(NanoarrowIpcReader, InputStreamFile) {
   ASSERT_EQ(ArrowIpcInputStreamInitFile(&stream, nullptr, 1), EINVAL);
 
   uint8_t input_data[] = {0x01, 0x02, 0x03, 0x04, 0x05};
+#ifdef _WIN32
+  FILE* file_ptr;
+  ASSERT_EQ(tmpfile_s(&file_ptr), 0);
```
WillAyd, line 120: I'm not sure if Windows expects tmpfile_s to be explicitly closed or not; at least in their documentation that does not seem to be called:

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/tmpfile-s?view=msvc-170

paleolimbot: I would be very surprised if this file weren't supposed to be closed (and it is closed in almost all cases because our inputstream implementation will do it for us). The `FileCloser` was added because some static analyzer somewhere complained about the file not being closed.

src/nanoarrow/ipc/reader_test.cc
```diff
@@ -115,8 +115,13 @@ TEST(NanoarrowIpcReader, InputStreamFile) {
   ASSERT_EQ(ArrowIpcInputStreamInitFile(&stream, nullptr, 1), EINVAL);
 
   uint8_t input_data[] = {0x01, 0x02, 0x03, 0x04, 0x05};
+#ifdef _WIN32
```
paleolimbot, line 118: Probably this should be `_MSC_VER` (with a comment explaining that this is to silence a warning)?

paleolimbot (approved):
Thanks for sorting these warnings! I have a few suggestions but it's also just test code and fine to merge if we'd like to move on.

PR #791, opened 2025-06-20
fix(ci): Meson windows example CI

WillAyd:


WillAyd:
From looking at the previous logs, I think the meson build is running fine. Maybe its the location in the build directory that is incorrect when trying to execute the program?

.github/workflows/examples.yaml
```diff
@@ -134,8 +132,17 @@ jobs:
...
 
+      - name: Meson example (Windows)
+        if: matrix.config.label == 'windows'
+        run: |
+          python3 -m pip install meson ninja
+          cd examples/meson-minimal
+          meson setup builddir
+          meson compile -C builddir
+
+          start .\builddir\example_meson_minimal_app.exe
```
WillAyd, line 143: There might be a smarter way to combine this step with the preceding one, but I couldn't get the github actions syntax quite right. 

Hopefully punt-able for now but let me know if any objection

Issue #793, opened 2025-06-27
Symlink in Meson Python configuration causes release verification problems


paleolimbot:
The release tarball can't be extracted on windows, for example:

https://github.com/apache/arrow-nanoarrow/actions/runs/15918271144/job/44899850928#step:10:136

```
  apache-arrow-nanoarrow-0.7.0.tar.gz: OK
  + tar xf apache-arrow-nanoarrow-0.7.0.tar.gz
  tar: apache-arrow-nanoarrow-0.7.0/python/subprojects/arrow-nanoarrow: Cannot create symlink to '../..': Too many levels of symbolic links
  tar: Exiting with failure status due to previous errors
```

...and I believe locally this was also the issue I had with Python verification on Windows/Git Bash.

paleolimbot:
Can we use the files that are already vendored for this? I can't quite get

```

nanoarrow_pyx_dep = declare_dependency(sources: vendored_files)
nanoarrow_lib = library('nanoarrow', meson.current_build_dir() + '/nanoarrow.c', dependencies: nanoarrow_pyx_dep)
```

to work:

```
  ../src/nanoarrow/meson.build:48:16: ERROR: File /Users/dewey/gh/arrow-nanoarrow/python/.mesonpy-yhi2ybnt/src/nanoarrow/nanoarrow.c does not exist.
```

...but clearly the generated .pxd file is there when the Cython is built so the files must be somewhere?

WillAyd:
Hmm nothing should have to change with the Meson configuration. Let me take a closer look - not super familiar with the Apache release process yet, but probably good to learn more about

WillAyd:
Oh...OK so my mistake on terminology. When I before talked about the `bootstrap` script what I meant to refer to was the `generate_dist.py` script. I see we separately have a bootstrap script that does something else, but the generate_dist.py script is what replaces symlinks with actual files

paleolimbot:
Ah...that requires the symlink:

```
/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/Resources/Python.app/Contents/MacOS/Python: can't open file '/Users/dewey/gh/arrow-nanoarrow/python/subprojects/arrow-nanoarrow/ci/scripts/bundle.py': [Errno 2] No such file or directory
```

WillAyd:
[...]

PR #794, opened 2025-06-27
ci: Replace symlinks with files when packaging source

WillAyd:
https://github.com/apache/arrow-nanoarrow/issues/793

### DIFF
--- a/dev/release/source_build.sh
+++ b/dev/release/source_build.sh
@@ -45,6 +45,8 @@ main() {
     mv "${base_name}/" "${base_name}.tmp/"
     cp -R -d "${base_name}.tmp" "${base_name}"
     rm -rf "${base_name}.tmp/"
+    MESON_SOURCE_ROOT="${base_name}/python" MESON_DIST_ROOT="${base_name}/python" \
+                     python "${base_name}/python/generate_dist.py"
 
     # Create new tarball
     tar czf "${tar_ball}" "${base_name}/"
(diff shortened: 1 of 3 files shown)


dev/release/source_build.sh
```diff
@@ -45,6 +45,8 @@ main() {
     mv "${base_name}/" "${base_name}.tmp/"
     cp -R -d "${base_name}.tmp" "${base_name}"
     rm -rf "${base_name}.tmp/"
+    MESON_SOURCE_ROOT="${base_name}/python" MESON_DIST_ROOT="${base_name}/python" \
```
WillAyd, line 48: My bash knowledge is really bad, so happy to be shown a better way of specifying these variables to be passed to the command :-)

dev/release/source_build.sh
```diff
@@ -45,6 +45,8 @@ main() {
...
+                     python "${base_name}/python/generate_dist.py"
```
WillAyd, line 49: Perhaps a more descriptive name for `generate_dist.py` would be `replace_subproject_symlinks.py`. Happy to make that change (or any other) here

WillAyd (commented):
OK I think this fixes the issue @paleolimbot

dev/release/source_build.sh
paleolimbot: Not just you..that's how to do it!

PR #803, opened 2025-08-27
fix: Assorted updates to improve Meson WrapDB entry

WillAyd:
After having submitted more wraps and gotten feedback from the Meson core team, I found out there are a few things we can do to improve our wrap submission and make it easier to include nanoarrow in other projects


meson.build
```diff
@@ -91,12 +91,20 @@ nanoarrow_lib = library(
...
+    description: 'Helpers for Arrow C Data & Arrow C Stream interfaces',
+)
+
 nanoarrow_dep = declare_dependency(
     include_directories: [incdir],
     link_with: nanoarrow_lib,
     compile_args: nanoarrow_dep_args,
 )
 
+meson.override_dependency('nanoarrow', nanoarrow_dep)
```
WillAyd, line 106: In today's wrap we have a provides section that looks like:

```ini
[provides]
nanoarrow = nanoarrow_dep
nanoarrow_ipc = nanoarrow_ipc_dep
```

But since Meson version 0.54.0, they have advised that a project internally call `meson.override_dependency` instead, so that the project has total control over their variable naming. As such, on the next release of Meson, we can simplify our wrap entry to just:

```ini
[provides]
dependency_names = [nanoarrow, nanoarrow-ipc, ...]
```

and decouple the internal variable name from the wrap system

meson.build
```diff
@@ -135,11 +143,17 @@ if get_option('ipc').enabled()
         c_args: ipc_lib_c_args,
         gnu_symbol_visibility: 'hidden',
     )
+    pkg.generate(
+        nanoarrow_ipc_lib,
+        filebase: 'nanoarrow-ipc',
```
WillAyd, line 148: I am possibly overlooking or misunderstanding the current CMake configuration, but I couldn't find anywhere that we were producing pkgconfig files. 

As such, its worth double checking if we want to use hyphens or underscores for the pkgconfig file name. I am using hyphens here as that is what arrow does (ex; `arrow-compute.pc`)

WillAyd:
@bgilbert if you have anything else to add here let me know. Happy to make any changes you think are best for the wrap system

meson.build
```diff
@@ -21,7 +21,7 @@ project(
     'cpp',
     version: '0.8.0-SNAPSHOT',
     license: 'Apache 2.0',
-    meson_version: '>=1.3.0',
+    meson_version: '>=1.1.0',
```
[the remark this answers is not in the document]
bgilbert: `meson.options` is undoubtedly a nicer name but IMO it's not worth constraining the Meson version just to avoid `meson_options.txt`.

meson.build
```diff
@@ -135,11 +143,14 @@ if get_option('ipc').enabled()
...
         gnu_symbol_visibility: 'hidden',
     )
+    pkg.generate(nanoarrow_ipc_lib, filebase: 'nanoarrow-ipc')
+
     nanoarrow_ipc_dep = declare_dependency(
         include_directories: [incdir],
         link_with: nanoarrow_ipc_lib,
         dependencies: [nanoarrow_dep],
     )
+    meson.override_dependency('nanoarrow-ipc', nanoarrow_dep)
```
bgilbert, line 153:
```suggestion
    meson.override_dependency('nanoarrow-ipc', nanoarrow_ipc_dep)
```

meson.build
```diff
@@ -21,7 +21,7 @@ project(
     'cpp',
     version: '0.8.0-SNAPSHOT',
     license: 'Apache 2.0',
```
bgilbert, line 23: SPDX license expression is preferred.

```suggestion
    license: 'Apache-2.0',
```

meson.build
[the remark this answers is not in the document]
WillAyd: Roger that - dropped to 0.58 as a minimum then, which we need for str.replace in the config

meson.build
WillAyd: This comment might have been dropped by subsequent changes, so just want to make sure you see it @paleolimbot in case you disagree with the chosen naming convention

paleolimbot: I think the arrow convention with hyphens makes sense!

paleolimbot (approved):
Thank you!

PR #808, opened 2025-09-19
fix: Ensure ArrowArrayBuffer() and ArrowArraySetBuffer() work for variadic buffers

paleolimbot:
The high-level goal of this PR is to fix a bug where the R package couldn't create a shallow copy of a string view array. The R and Python packages create shallow copies by borrowing buffers from the original and constructing an array using `ArrowArraySetBuffer()`. Because `ArrowArraySetBuffer()` errored for any buffer other than 0, 1, and 2, string view arrays caused cryptic errors when they hit certain places in the R package.

After this PR, `ArrowArrayBuffer()` and `ArrowArraySetBuffer()` work for any buffer, variadic or no. This means it's possible to create string/binary view arrays by buffer, even if it's mostly something we just do internally.

I'd hoped this would be a more straightforward fix, but the assumption that there will always be <=3 buffers ran fairly deep.

### DIFF
--- a/python/tests/test_c_array.py
+++ b/python/tests/test_c_array.py
@@ -502,8 +502,8 @@ def test_c_array_from_buffers_recursive():
 
 
 def test_c_array_from_buffers_string_view():
-    # Creating the actual view buffer is hard, but we can at least make sure the buffers are able to
-    # be roundtripped.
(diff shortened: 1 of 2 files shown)


WillAyd (commented):
Nice! I think this generally makes a lot of sense

PR #830, opened 2025-12-05
exception.hpp: add override qualifier to what() method

rouault:
Enables to build with -Wsuggest-override


paleolimbot (approved):
Thank you!

PR #839, opened 2026-01-14
chore: Re-enable CUDA tests with runs-on

### LINKED ISSUE #814: CUDA Runner is no longer available?
All of the latest CUDA CI jobs seem to be timing out/unable to find the runner tagged `["self-hosted", "cuda"]`. I think these were also used for crossbow/nightlies and I know there was a change there...is there any other configuration that we need to update to get these back? (Or do we just have to drop the CUDA CI?)

@raulcd (Or perhaps you know the right person to ping?)

paleolimbot:
Closes #814.


rok:
Great! Happy to see we have this back!

PR #863, opened 2026-04-06
chore: add libuv-dev to alpine-s390x dockerfile

amoeba:
fs >= 2.0.0 now prefers system libuv and you appear to have to opt into
building with the bundled version. Adding libuv-dev to the Dockefile
seems like the best fix.


paleolimbot (approved):
Thank you!

(I'll merge + trigger the build workflow manually on main. I can't recommend setting it up this way but it seemed like a good idea 3 years ago and it doesn't come up very often...)

PR #864, opened 2026-04-06
feat: Add shared array and buffer to nanoarrow.h

paleolimbot:
For #861 (actual decoding of dictionary arrays), we really need shared arrays for this to be reasonable (e.g., to not deep copy each dictionary for each batch!).

This PR (1) moves shared buffers to the C library instead of the IPC extension, (2) implements a second kind of shared buffer, which is borrowed from an owned reference-counted array, and (3) implements a "clone" that mostly uses the second concept to explode an array into 100% reference counted buffers that we can clone.

I had hoped we could replace the R and Python versions of these but the logic is fuzzier there because once an array has been referenced we can't mess with it or any of its parents (or it will cause a crash). My mind explodes (and I get a lot of failing R tests cases) whenever I mess with that piece and so I'll try to deal with that a different day.

This still needs tests for the shared array move and clone.


WillAyd (approved):
nice work - i think this looks pretty good

PR #877, opened 2026-04-18
chore: Fix CMake linking to static-only nanoarrow install with BUILD_SHARED_LIBS

### LINKED ISSUE #875: BUILD_SHARED_LIBS in the config should be set at build time
Currently the CMake config uses the value of `BUILD_SHARED_LIBS` whenever the config is executed, which means it is the value from other builds. However, what we really want here is to use the value of `BUILD_SHARED_LIBS` when nanoarrow itself was built since that controls which libraries are actually present. What we probably want is to add some logic [after the `configure_file`](https://github.com/apache/arrow-nanoarrow/blob/3306ac6e884392c8f7303d9dd3355af546276b65/CMakeLists.txt#L846) to write an explicit target alias to the generated file based on the value of `BUILD_SHARED_LIBS` during the nanoarrow build.

paleolimbot:
Closes #875.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -830,6 +830,13 @@ endif()
 # Generate package files for the build and install trees.
 include(CMakePackageConfigHelpers)
 
+# Capture BUILD_SHARED_LIBS at configure time for use in the config file
+if(BUILD_SHARED_LIBS)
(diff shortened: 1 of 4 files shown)


vyasr (approved):
Yup this looks great, thanks!

PR #882, opened 2026-05-12
feat: Encode dictionary schemas

paleolimbot:
This PR adds support in the IPC writer for dictionary schemas, verifying support by removing skips and adding cases for all the valid dictionary index types in the parameterized test cases.


sounkou-bioinfo:
> This PR adds support in the IPC writer for dictionary schemas, verifying support by removing skips and adding cases for all the valid dictionary index types in the parameterized test cases.

this is great ! it should remove some brittle codepaths in the IPC conversion path in this project https://github.com/sounkou-bioinfo/Rducks/

PR #883, opened 2026-05-16
fix: Correctly subscript ViewArrayAs with offset

borchero:
# Motivation

There is currently a bug in `ViewArrayAs`: when using subscripting, it ignores the underlying array's offset, potentially yielding incorrect values.


paleolimbot (approved):
Thank you!

Issue #622, opened 2024-09-18
Support dictionary-encoded types in the IPC reader and writer


paleolimbot:
Neither the IPC reader nor the IPC writer support dictionary encoding. The lack of writer support is problematic for the R bindings because the `factor` type is converted by default to a dictionary-encoded string; the lack of reader support is problematic because the IPC reader is currently forced to error for streams produced elsewhere, many of which will happily emit dictionary messages. This also complicates our integration testing since currently we have to skip any file that includes a dictionary.

Delta dictionaries are probably out of scope for the initial level of support.

white123:
Hi @paleolimbot,

Following up on this — is there a current plan or timeline for adding dictionary support to the IPC writer/reader? Just trying to understand the expected direction.

Thanks!

rustyconover:
Dictionary write support would unlock a few ideas I've had, i.e. changing duckdb-wasm to use nanoarrow rather than arrow c++ for IPC would be one of them.  Commenting for following.

paleolimbot:
Thanks for the +1! I think it would be great too. I took me a few months to plug through read support and it will probably take a few months to plug through the write support as well.

PR #926, opened 2026-08-25
feat: Add DictionaryBatch write support to IPC encoder and writer

niekverw:
Hi, I had implemented dictionary batch writing for a project that I was working on, I noticed a thread (https://github.com/apache/arrow-nanoarrow/issues/622) about it, so I just wanted to share what I had in case it is useful.  I needed this for a project that emits dictionary-encoded columns over IPC.  I use ArrowIpcWriterWriteDictionaryBatch directly, for the PR I wired it into WriteArrayStream  so dictionaries get emitted automatically on the high-level path and added a write→read round-trip test. Happy to also adjust.

### DIFF
--- a/src/nanoarrow/ipc/writer.c
+++ b/src/nanoarrow/ipc/writer.c
@@ -317,6 +317,73 @@ ArrowErrorCode ArrowIpcWriterWriteArrayView(struct ArrowIpcWriter* writer,
   return NANOARROW_OK;
 }
 
+ArrowErrorCode ArrowIpcWriterWriteDictionaryBatch(
+    struct ArrowIpcWriter* writer, int64_t dictionary_id, char is_delta,
+    const struct ArrowArrayView* values_view, struct ArrowError* error) {
+  NANOARROW_DCHECK(writer != NULL && writer->private_data != NULL && values_view != NULL);
+  struct ArrowIpcWriterPrivate* private =
+      (struct ArrowIpcWriterPrivate*)writer->private_data;
+
+  NANOARROW_ASSERT_OK(ArrowBufferResize(&private->buffer, 0, 0));
+  NANOARROW_ASSERT_OK(ArrowBufferResize(&private->body_buffer, 0, 0));
+
+  NANOARROW_RETURN_NOT_OK(ArrowIpcEncoderEncodeSimpleDictionaryBatch(
+      &private->encoder, dictionary_id, is_delta, values_view, &private->body_buffer,
+      error));
+  NANOARROW_RETURN_NOT_OK_WITH_ERROR(
+      ArrowIpcEncoderFinalizeBuffer(&private->encoder, /*encapsulate=*/1,
+                                    &private->buffer),
+      error);
+
+  if (private->writing_file) {
+    _NANOARROW_CHECK_RANGE(private->buffer.size_bytes, 0, INT32_MAX);
+    struct ArrowIpcFileBlock block = {
+        .offset = private->bytes_written,
+        .metadata_length = (int32_t) private->buffer.size_bytes,
+        .body_length = private->body_buffer.size_bytes,
+    };
+    NANOARROW_RETURN_NOT_OK_WITH_ERROR(
+        ArrowBufferAppend(&private->footer.dictionary_blocks, &block, sizeof(block)),
+        error);
+  }
+  private->bytes_written += private->buffer.size_bytes;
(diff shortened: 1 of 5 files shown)


rustyconover:
Hi @niekverw I took where you started and iterated on it a bit.  Thanks for the taking the first crack at this dictionary write support.

src/nanoarrow/ipc/writer_test.cc
```diff
@@ -204,3 +204,180 @@ TEST(NanoarrowIpcWriter, FileWriting) {
...
+  EXPECT_EQ(ArrowArrayViewGetIntUnsafe(indices_view, 0), 0);
+  EXPECT_EQ(ArrowArrayViewGetIntUnsafe(indices_view, 1), 1);
+  EXPECT_EQ(ArrowArrayViewGetIntUnsafe(indices_view, 2), 0);
+
+  struct ArrowStringView v0 = ArrowArrayViewGetStringUnsafe(values_view, 0);
+  struct ArrowStringView v1 = ArrowArrayViewGetStringUnsafe(values_view, 1);
+  EXPECT_EQ(std::string(v0.data, v0.size_bytes), "foo");
+  EXPECT_EQ(std::string(v1.data, v1.size_bytes), "bar");
+
+  ASSERT_EQ(ArrowArrayStreamGetNext(reader.get(), roundtrip_array.get(), &error),
```
paleolimbot, line 379: I believe the CI error is is a memory leak coming from here (just need to reset the array before reading into it)

```suggestion

  roundtrip_array.reset();
  ASSERT_EQ(ArrowArrayStreamGetNext(reader.get(), roundtrip_array.get(), &error),
```

src/nanoarrow/ipc/writer.c
```diff
@@ -317,6 +317,73 @@ ArrowErrorCode ArrowIpcWriterWriteArrayView(struct ArrowIpcWriter* writer,
...
+      &private->output_stream, ArrowBufferToBufferView(&private->body_buffer), error));
+  return NANOARROW_OK;
+}
+
+// Walk the array in the same depth-first order the schema encoder uses to assign
+// dictionary ids (see ArrowIpcDictionaryEncodingsAppendSchema): a dictionary-encoded
+// node claims the next id before descending into its children and then its values.
+// Emitting a full (non-delta) DictionaryBatch for each dictionary before every
+// RecordBatch keeps each batch's indices valid against the dictionary that precedes
+// it, which is required because each array in the stream carries its own dictionary.
```
paleolimbot, line 364: I think this is a good scope for this PR, but we will want to ensure we have some way to avoid emitting duplicate dictionaries as tracked follow-up.

src/nanoarrow/ipc/writer.c
```diff
@@ -317,6 +317,73 @@ ArrowErrorCode ArrowIpcWriterWriteArrayView(struct ArrowIpcWriter* writer,
...
+
+  for (int64_t i = 0; i < array_view->n_children; i++) {
+    NANOARROW_RETURN_NOT_OK(ArrowIpcWriterWriteDictionariesForArrayView(
+        writer, array_view->children[i], next_id, error));
+  }
+
+  if (array_view->dictionary != NULL) {
+    NANOARROW_RETURN_NOT_OK(ArrowIpcWriterWriteDictionariesForArrayView(
+        writer, array_view->dictionary, next_id, error));
+  }
```
paleolimbot, line 382: The encoder rejects dictionary values that have a dictionary. Should this error instead? (Or if it works, can we add a test?)

paleolimbot (approved):
This is great work!

The CI failure is a small test issue...other than that just a question to make sure I'm understanding the nested dictionary situation.

@rustyconover I think we can merge this after fixing the test leak an integrate your additional work on top. What do you think?

rustyconover:
I think so too.  I thought this PR was a great start thats why I did a bunch of work on top of it!  

I'd like to get dictionary write support landed, because then I'll roll it out across a few of my Query Farm's projects and really start to use nanoarrow a lot more!

niekverw:
Hey all, no problem, just go with Rusty’s PR - I just wanted to share this as I was using it and seemed to work for me - no need to merge this specific PR if rusty has built a better solution. 
thank you!!

rustyconover:
Thanks @niekverw

paleolimbot (approved):
Thank you!

PR #927, opened 2026-08-28
feat(kotoba): add v1 Arrow IPC schema subset

com-junkawasaki:
Adds `kotoba/` next to `python/` and `r/`.

Kotoba cannot FFI libarrow. v1 is an in-language Arrow IPC / schema subset: empty schema, one nullable int32 field, and an int32 RecordBatch header. Host owns files.

Kotoba CI on the fork is green. An ASF ICLA may apply for a first-time committer.

Fork PR: https://github.com/kotoba-lang/arrow-nanoarrow/pull/1

### DIFF
--- a/src/nanoarrow/ipc/decoder.c
+++ b/src/nanoarrow/ipc/decoder.c
@@ -81,6 +81,10 @@ struct ArrowIpcDecoderPrivate {
   int64_t n_union_fields;
   // A pointer to the last flatbuffers message.
   const void* last_message;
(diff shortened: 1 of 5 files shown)


paleolimbot (commented):
Thank you for opening!

I'm glad that nanoarrow is a good fit for kotoba language bindings! I don't have any experience with kotoba, so I think this repository might not be a good fit. Is there a repository in a different organization that would have review capacity to take this one? (Or could this stay in a repository under your account with clear licensing/commit history until we find a place in the arrow organization to donate it to that has the requisite review capacity?)

Issue #824, opened 2025-11-12
StringView not supported in IPC reader or writer


paleolimbot:
Noted in https://github.com/paleolimbot/duckdb-nanoarrow/issues/13 ...if an IPC file contains the string view type it will error on read.

After https://github.com/apache/arrow-nanoarrow/pull/808 we should be able to use the same buffer infrastructure for setting/getting buffers as we do for all the other types.

PR #928, opened 2026-09-02
feat(ipc): support dictionary updates and deltas

rustyconover:
## Summary

- replace the #926 per-batch dictionary emission path with cached dictionary state
- suppress unchanged dictionaries and emit full replacements when values change
- emit append-only dictionary changes as delta batches
- decode dictionary delta batches by concatenating storage with `ArrowArrayAppendStorageFromArrayView()`
- preserve dictionary metadata, nested dictionary ordering/invalidation, file-mode rules, and validation coverage

## Relationship to #930

#930 has merged, and this branch has been rebased onto the resulting `main`. This PR now uses the public `ArrowArrayAppendStorageFromArrayView()` API from #930; the array-view appender implementation and #930 commits are no longer part of this diff.

The remaining changes in this PR are the dictionary IPC update and delta functionality.

## IPC behavior

Streams cache dictionary values, omit unchanged dictionaries, emit append-only suffixes with `DictionaryBatch.isDelta`, and advertise `DICTIONARY_REPLACEMENT` when a full replacement may occur. Nested dictionaries are emitted dependency-first and dependents are invalidated when needed.

Files permit append-only delta dictionaries but reject full dictionary replacement after a dictionary ID has been written.

Delta decoding supports the storage types handled by `ArrowArrayAppendStorageFromArrayView()`. Dense union deltas currently return `ENOTSUP`, matching #930 support.

## Validation

- pinned clang-format hook passed
- native CMake/CTest: 393/393 passed (16 optional codec tests skipped)
- Arrow C++ delta interoperability: 12/12 passed, including the explicit dense-union rejection case
- Arrow-enabled writer suite: 18 passed (4 optional codec tests skipped)


paleolimbot (commented):
Thank you for this!

I will take a closer look in the next few days, but can I ask that the delta dictionary and/or appender piece be separated from the IPC write piece? All three are independently useful and I think it will be a smoother process if they're PRed separately as well.

rustyconover:
Yep I'll break them out!

Rusty

rustyconover:
Split completed as requested:

- #930 contains the generic `ArrowArrayAppendArrayView()` functionality as a one-commit core-array PR.
- This PR has been rebuilt on current `main` and now contains only dictionary IPC encoding/writing (no decoder changes, appender, quadratic-delta helper, or automatic delta emission).
- The dictionary-delta decoder is preserved on `Query-farm:feat/ipc-dictionary-delta-decoding`; I will submit it after #930 merges so its upstream diff contains only decoder work.

The original full branch is preserved as `Query-farm:split/pr928-full`.

src/nanoarrow/ipc/decoder.c
```diff
@@ -337,17 +352,117 @@ static ArrowErrorCode ArrowIpcDictionaryReplace(struct ArrowIpcDictionary* dicti
...
+  // to be regenerated by the next ArrowArrayFinishBuildingDefault().
+  if (array_view->storage_type == NANOARROW_TYPE_BINARY_VIEW ||
+      array_view->storage_type == NANOARROW_TYPE_STRING_VIEW) {
+    ArrowBufferReset(ArrowArrayBuffer(array, array->n_buffers - 1));
+  }
+
+  for (int64_t i = 0; i < array->n_children; i++) {
+    ArrowIpcArrayPrepareForAppend(array->children[i], array_view->children[i]);
+  }
+}
```
paleolimbot, line 385: I don't think we support views yet for IPC decoding at all yet? https://github.com/apache/arrow-nanoarrow/issues/824

src/nanoarrow/ipc/decoder.c
```diff
@@ -337,17 +352,117 @@ static ArrowErrorCode ArrowIpcDictionaryReplace(struct ArrowIpcDictionary* dicti
...
+    ArrowArrayMove(&dictionary->current_value, &combined);
+
+    ArrowIpcArrayPrepareForAppend(&combined, array_view);
+    ArrowErrorCode result = ArrowArrayReserve(&combined, value->length);
+    if (result == NANOARROW_OK) {
+      result = ArrowArrayAppendStorageFromArrayView(&combined, array_view, error);
+    }
+    if (result == NANOARROW_OK) {
+      result = ArrowIpcArraySetDictionaries(&combined, value);
+    }
```
paleolimbot, line 412: For the PR that does contain this, a good pattern is to define a function that can fail and doesn't own any input pointers:

```c
ArrowErrorCode ArrowIpcDoAppend(struct ArrowArray* combined, struct ArrowArrayView array_view, struct ArrowError* error) {
  NANOARROW_RETURN_NOT_OK(...);
  return NANOARROW_OK;
}
```

then here you can just check once and release the temporary

```c
  ArrowErrorCode result = ArrowIpcDoAppend(...);
  if (result != NANOARROW_OK) {
    // release stuff
    return result;
  }
```

src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -2099,4 +2164,279 @@ INSTANTIATE_TEST_SUITE_P(NanoarrowIpcTest, ArrowTypeIdParameterizedTestFixture,
...
+  kInt64WithNull,
+  kUInt64,
+  kDouble,
+  kString,
+  kBinary,
+  kDecimal128,
+  kList,
+  kStruct,
+  kFixedSizeList
+};
```
paleolimbot, line 2180: Can you reuse `enum ArrowType` for this one?

src/nanoarrow/ipc/decoder_test.cc
```diff
@@ -2099,4 +2164,279 @@ INSTANTIATE_TEST_SUITE_P(NanoarrowIpcTest, ArrowTypeIdParameterizedTestFixture,
...
+  auto maybe_array2 = arrow::DictionaryArray::FromArrays(
+      dictionary_type, MakeDeltaDictionaryIndices(true), values2);
+  ASSERT_TRUE(maybe_array1.ok()) << maybe_array1.status();
+  ASSERT_TRUE(maybe_array2.ok()) << maybe_array2.status();
+  AssertReadsArrowCppDeltaStream(maybe_array1.ValueUnsafe(), maybe_array2.ValueUnsafe());
+}
+
+TEST(NanoarrowIpcTest, RejectsArrowCppDenseUnionDictionaryDelta) {
+  arrow::Int8Builder type_ids_builder;
+  arrow::Int32Builder offsets_builder;
```
paleolimbot, line 2395: I think we can skip this test unless there's something special about rejecting a type in the decoder implementation

src/nanoarrow/ipc/decoder.c
```diff
@@ -337,17 +352,117 @@ static ArrowErrorCode ArrowIpcDictionaryReplace(struct ArrowIpcDictionary* dicti
...
+  }
+
+  // In the usual streaming loop, the previously returned batch has been released
+  // before the next one is requested. Recover the mutable backing array and append
+  // directly so a sequence of small deltas grows geometrically instead of copying
+  // the complete dictionary for every message. If an older batch is still alive,
+  // keep the copy-on-write path below to preserve its dictionary snapshot.
+  if (ArrowArrayInternalTryUnshare(&dictionary->current_value)) {
+    struct ArrowArray combined;
+    ArrowArrayMove(&dictionary->current_value, &combined);
```
paleolimbot, line 403: The usual streaming loop should be using `ArrowIpcDecoderDecodeArrayViewWithDictionaries()`, which in a perfect world never was shared (however, it probably had a shared backing buffer on its first decode because most decoding happens from a shared buffer).

Rather than trying to unshare, can we check whether this array can be appended to (and if possibly delay the sharing of it until it is requested as an array that is not just a view?). Totally ok if not possible, but if so, we should punt on the trying to unshare piece and implement that in a follow up.

src/nanoarrow/ipc/decoder.c
```diff
@@ -337,17 +352,117 @@ static ArrowErrorCode ArrowIpcDictionaryReplace(struct ArrowIpcDictionary* dicti
...
-  NANOARROW_RETURN_NOT_OK(ArrowIpcDictionaryReplace(dictionary, value, error));
-  return NANOARROW_OK;
+  struct ArrowArray combined;
+  combined.release = NULL;
+  NANOARROW_RETURN_NOT_OK(ArrowArrayInitFromArrayView(&combined, array_view, error));
+  ArrowErrorCode result = ArrowArrayStartAppending(&combined);
+  if (result == NANOARROW_OK) {
+    result =
+        ArrowArrayReserve(&combined, dictionary->current_value.length + value->length);
+  }
```
paleolimbot, line 436: The same pattern applies here

src/nanoarrow/ipc/encoder.c
```diff
@@ -216,6 +218,14 @@ ArrowErrorCode ArrowIpcEncoderSetCompression(
   return ArrowIpcEncoderSetCompressor(encoder, &compressor);
 }
 
+void ArrowIpcEncoderSetDictionaryReplacement(struct ArrowIpcEncoder* encoder,
+                                             char enabled) {
+  NANOARROW_DCHECK(encoder != NULL && encoder->private_data != NULL);
+  struct ArrowIpcEncoderPrivate* private =
+      (struct ArrowIpcEncoderPrivate*)encoder->private_data;
+  private->dictionary_replacement = enabled != 0;
+}
```
paleolimbot, line 227: This is a great and super useful change, but a separate one from the decoding or encoding of delta dictionaries

src/nanoarrow/ipc/files_test.cc
```diff
@@ -608,10 +614,10 @@ INSTANTIATE_TEST_SUITE_P(
         TestFile::OK("cpp-21.0.0/generated_primitive_zerolength.stream"),
         TestFile::OK("cpp-21.0.0/generated_recursive_nested.stream"),
         TestFile::OK("cpp-21.0.0/generated_union.stream"),
-        TestFile::ReadOnly("cpp-21.0.0/generated_dictionary.stream"),
-        TestFile::ReadOnly("cpp-21.0.0/generated_dictionary_unsigned.stream"),
+        TestFile::OK("cpp-21.0.0/generated_dictionary.stream"),
+        TestFile::OK("cpp-21.0.0/generated_dictionary_unsigned.stream"),
         TestFile::ReadOnly("cpp-21.0.0/generated_extension.stream"),
-        TestFile::ReadOnly("cpp-21.0.0/generated_nested_dictionary.stream"),
+        TestFile::OK("cpp-21.0.0/generated_nested_dictionary.stream"),
```
paleolimbot, line 620: Just curious: what feature is needed to roundtrip the generated extension beyond what's here?

src/nanoarrow/ipc/writer.c
```diff
@@ -357,57 +430,386 @@ ArrowErrorCode ArrowIpcWriterWriteDictionaryBatch(
...
-  NANOARROW_RETURN_NOT_OK(ArrowIpcOutputStreamWrite(
-      &private->output_stream, ArrowBufferToBufferView(&private->buffer), error));
-  NANOARROW_RETURN_NOT_OK(ArrowIpcOutputStreamWrite(
-      &private->output_stream, ArrowBufferToBufferView(&private->body_buffer), error));
   return NANOARROW_OK;
 }
 
+static void ArrowIpcWriterCanonicalizeBitmapPadding(
+    struct ArrowArray* array, const struct ArrowArrayView* array_view) {
+  int64_t remainder = array->length % 8;
```
paleolimbot, line 490: It would help to put the comment about canonicalizing the padding helping to minimize reemitting a dictionary here.

Other parts of nanoarrow typically avoid this by zeroing out the last byte of a bitmap when reserving but it is sometimes hard to guarantee that everywhere.

src/nanoarrow/ipc/writer.c
```diff
@@ -357,57 +430,386 @@ ArrowErrorCode ArrowIpcWriterWriteDictionaryBatch(
...
+
+  struct ArrowArrayView slice = *src;
+  slice.offset += offset;
+  slice.length = length;
+  slice.null_count = -1;
+
+  ArrowErrorCode result = ArrowArrayInitFromArrayView(out, src, error);
+  if (result == NANOARROW_OK) {
+    result = ArrowArrayStartAppending(out);
+  }
```
paleolimbot, line 533: The same nested function strategy for avoiding these repeated result checks applies here, too

src/nanoarrow/ipc/writer.c
```diff
@@ -357,57 +430,386 @@ ArrowErrorCode ArrowIpcWriterWriteDictionaryBatch(
...
+
+  if (result != NANOARROW_OK && out->release != NULL) {
+    ArrowArrayRelease(out);
+  }
+  return result;
+}
+
+static ArrowErrorCode ArrowIpcWriterCompareMaterializedArrays(
+    const struct ArrowArray* lhs, const struct ArrowArray* rhs,
+    const struct ArrowArrayView* shape, int* out, struct ArrowError* error) {
```
paleolimbot, line 558: Rather than compare by value (possibly slower than just writing another dictionary to the stream), probably comparing the buffer pointers makes more sense. If it's a shared array it should be pointing to the same data.

src/nanoarrow/ipc/writer.c
```diff
@@ -357,57 +430,386 @@ ArrowErrorCode ArrowIpcWriterWriteDictionaryBatch(
...
+    }
+
+    if (prefix_equal) {
+      result = ArrowIpcWriterMaterializeArrayView(
+          values_view, cached->values.length,
+          current_values.length - cached->values.length, &delta_values, error);
+      if (result != NANOARROW_OK) {
+        goto cleanup;
+      }
+      is_delta = 1;
```
paleolimbot, line 657: You can use the nested function approach for shared cleanup (we don't use `goto` in nanoarrow, or at least we haven't yet)

src/nanoarrow/ipc/writer.c
```diff
@@ -357,57 +430,386 @@ ArrowErrorCode ArrowIpcWriterWriteDictionaryBatch(
...
+    goto cleanup;
+  }
+
+  if (allow_delta && !force_emit && cached != NULL &&
+      current_values.length > cached->values.length) {
+    result = ArrowIpcWriterMaterializeArrayView(values_view, 0, cached->values.length,
+                                                &prefix_values, error);
+    if (result != NANOARROW_OK) {
+      goto cleanup;
+    }
```
paleolimbot, line 641: I don't think that the writer should emit delta dictionaries in this way...this is a kind of calculation that application code should do, which may have access to things like C++ that can do it more efficiently.

In any case, emitting delta dictionaries is a third scope here for a separate PR

paleolimbot (commented):
Thank you for this!

I think this PR contains 3 orthogonal changes:

- Decoding delta dictionaries + some small fixes to the decoder
- Avoiding dictionary replacement when emitting dictionaries that are identical + assorted fixes to the encoder for nested dictionaries
- Encoding/writing delta dictionaries

I did a first pass on everything and I think we can get all these merged soon!

PR #929, opened 2026-09-02
ci: authenticate Codecov uploads with OIDC

rustyconover:
The coverage workflow on `main` currently completes successfully even though Codecov rejects its upload with `Token required because branch is protected`. This leaves the repository coverage baseline stale and causes PR patch reports to compare against an old commit.

Configure the Codecov action to use GitHub OIDC, grant the required permissions, and fail the job when an upload is rejected. Fork pull requests continue to use Codecov tokenless uploads because the action detects fork events and skips OIDC for them.

The failed protected-branch upload is visible in https://github.com/apache/arrow-nanoarrow/actions/runs/32611406823.

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -37,6 +37,9 @@ jobs:
   coverage:
 
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
 
     steps:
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
Thank you for this...the codecov setup is wildly out of date.

PR #930, opened 2026-09-02
feat: append array views to arrays

rustyconover:
## Summary

- add `ArrowArrayAppendArrayView()` as a core array-building API
- append primitive, nested, union, list-view, and run-end encoded values from an `ArrowArrayView`
- preserve source slicing semantics and report unsupported sliced run-end encoded inputs
- add direct native coverage for primitive, nested/sliced, union, and run-end encoded arrays

## Context

This extracts the generic array-view appender from #928 as requested in review. It is independently useful and will be the prerequisite for a separate dictionary-delta decoding PR; #928 no longer contains the appender or decoder changes.

## Validation

- native CMake build passed
- native CTest suite: 334/334 passed (4 optional codec tests skipped)

### DIFF
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -436,7 +436,7 @@ static ArrowErrorCode ArrowArrayAppendValidityFromArrayView(
 
   NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowBitmapReserve(dst_validity, src->length),
                                      error);
-  int8_t validity[1024];
+  int8_t validity[1024] = {0};
   for (int64_t offset = 0; offset < src->length;) {
     int64_t remaining = src->length - offset;
     int64_t chunk_size =
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -186,6 +186,14 @@ static ArrowErrorCode ArrowArrayCheckCanAppendStorageFromArrayView(
   struct ArrowArrayPrivateData* private_data =
       (struct ArrowArrayPrivateData*)dst->private_data;
   enum ArrowType dst_type = private_data->storage_type;
+
+  if (src->storage_type == NANOARROW_TYPE_DENSE_UNION ||
+      src->storage_type == NANOARROW_TYPE_SPARSE_UNION) {
+    ArrowErrorSet(error, "Appending array views is not supported for %s",
+                  ArrowTypeString(src->storage_type));
+    return ENOTSUP;
+  }
+
   if (!ArrowArrayCanAppendStorageType(dst_type, src->storage_type)) {
     ArrowErrorSet(error, "Can't append %s storage to an array with %s storage",
                   ArrowTypeString(src->storage_type), ArrowTypeString(dst_type));
@@ -299,25 +307,6 @@ static ArrowErrorCode ArrowArrayAppendStorageFromArrayViewElement(
       NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowArrayAppendInterval(dst, &interval), error);
       return NANOARROW_OK;
     }
-    case NANOARROW_TYPE_DECIMAL32:
-    case NANOARROW_TYPE_DECIMAL64:
-    case NANOARROW_TYPE_DECIMAL128:
-    case NANOARROW_TYPE_DECIMAL256: {
-      int32_t bitwidth = 32;
-      if (src->storage_type == NANOARROW_TYPE_DECIMAL64) {
-        bitwidth = 64;
-      } else if (src->storage_type == NANOARROW_TYPE_DECIMAL128) {
-        bitwidth = 128;
-      } else if (src->storage_type == NANOARROW_TYPE_DECIMAL256) {
-        bitwidth = 256;
-      }
-
-      struct ArrowDecimal decimal;
-      ArrowDecimalInit(&decimal, bitwidth, /*precision=*/0, /*scale=*/0);
-      ArrowArrayViewGetDecimalUnsafe(src, i, &decimal);
-      NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowArrayAppendDecimal(dst, &decimal), error);
-      return NANOARROW_OK;
-    }
     case NANOARROW_TYPE_STRUCT:
       for (int64_t child_i = 0; child_i < src->n_children; child_i++) {
         NANOARROW_RETURN_NOT_OK(ArrowArrayAppendStorageFromArrayViewElement(
@@ -357,45 +346,12 @@ static ArrowErrorCode ArrowArrayAppendStorageFromArrayViewElement(
       NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowArrayFinishElement(dst), error);
       return NANOARROW_OK;
     }
+    case NANOARROW_TYPE_DECIMAL32:
+    case NANOARROW_TYPE_DECIMAL64:
+    case NANOARROW_TYPE_DECIMAL128:
+    case NANOARROW_TYPE_DECIMAL256:
     case NANOARROW_TYPE_DENSE_UNION:
-    case NANOARROW_TYPE_SPARSE_UNION: {
-      int8_t type_id = ArrowArrayViewUnionTypeId(src, i);
-      int8_t child_index = ArrowArrayViewUnionChildIndex(src, i);
-      int64_t child_offset = ArrowArrayViewUnionChildOffset(src, i);
-      NANOARROW_RETURN_NOT_OK(ArrowArrayAppendStorageFromArrayViewElement(
-          dst->children[child_index], src->children[child_index], child_offset, error));
-      NANOARROW_RETURN_NOT_OK_WITH_ERROR(
-          ArrowBufferAppend(ArrowArrayBuffer(dst, 0), &type_id, sizeof(type_id)), error);
-      if (src->storage_type == NANOARROW_TYPE_DENSE_UNION) {
-        int64_t dst_child_offset = dst->children[child_index]->length - 1;
-        if (dst_child_offset < 0 || dst_child_offset > INT32_MAX) {
-          ArrowErrorSet(error,
-                        "Expected dense union child offset to fit in int32 but "
-                        "found %" PRId64,
-                        dst_child_offset);
-          return EOVERFLOW;
-        }
-        NANOARROW_RETURN_NOT_OK_WITH_ERROR(
-            ArrowBufferAppendInt32(ArrowArrayBuffer(dst, 1), (int32_t)dst_child_offset),
-            error);
-      } else {
-        for (int64_t child_i = 0; child_i < dst->n_children; child_i++) {
-          if (child_i != child_index && dst->children[child_i]->length == dst->length) {
-            NANOARROW_RETURN_NOT_OK_WITH_ERROR(
-                ArrowArrayAppendEmpty(dst->children[child_i], 1), error);
-          }
-          if (dst->children[child_i]->length != dst->length + 1) {
-            ArrowErrorSet(error,
-                          "Expected sparse union child length of %" PRId64
-                          " but found %" PRId64,
-                          dst->length + 1, dst->children[child_i]->length);
-            return EINVAL;
-          }
-        }
-      }
-      dst->length++;
-      return NANOARROW_OK;
-    }
+    case NANOARROW_TYPE_SPARSE_UNION:
     case NANOARROW_TYPE_RUN_END_ENCODED:
     case NANOARROW_TYPE_UNINITIALIZED:
     default:
@@ -409,7 +365,8 @@ static int ArrowArrayCanAppendFixedWidthStorage(struct ArrowArray* dst,
                                                 const struct ArrowArrayView* src) {
   struct ArrowArrayPrivateData* private_data =
       (struct ArrowArrayPrivateData*)dst->private_data;
-  return private_data->storage_type == src->storage_type && src->n_children == 0 &&
+  return private_data->storage_type == src->storage_type && dst->n_buffers == 2 &&
+         src->n_children == 0 &&
          src->layout.buffer_type[1] == NANOARROW_BUFFER_TYPE_DATA &&
          src->layout.element_size_bits[1] > 0 &&
          src->layout.element_size_bits[1] % 8 == 0;
--- a/src/nanoarrow/common/array.c
+++ b/src/nanoarrow/common/array.c
@@ -169,6 +169,14 @@ static int ArrowArrayCanAppendStorageType(enum ArrowType dst_type,
 
 static ArrowErrorCode ArrowArrayCheckCanAppendStorageFromArrayView(
     struct ArrowArray* dst, const struct ArrowArrayView* src, struct ArrowError* error) {
+  if (src->offset < 0 || src->length < 0 || src->offset > INT64_MAX - src->length) {
+    ArrowErrorSet(error,
+                  "Expected source offset and length to describe a valid int64 "
+                  "range but found %" PRId64 " and %" PRId64,
+                  src->offset, src->length);
+    return EINVAL;
+  }
+
   if (!ArrowArrayIsInternal(dst)) {
     ArrowErrorSet(error, "Expected destination to be an internal ArrowArray");
     return EINVAL;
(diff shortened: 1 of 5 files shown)


src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,174 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+      ArrowErrorSet(error, "Appending array views is not supported for %s",
+                    ArrowTypeString(src->storage_type));
+      return ENOTSUP;
+  }
+}
+
+ArrowErrorCode ArrowArrayAppendArrayView(struct ArrowArray* dst,
+                                         const struct ArrowArrayView* src,
+                                         struct ArrowError* error) {
+  if (src->storage_type == NANOARROW_TYPE_RUN_END_ENCODED) {
```
paleolimbot: At the top of this function we should sanity check that the append makes sense, because I am not sure that all combinations of spurious comparisons will generate a reasonable error here. ArrowArrayView only gives you storage type, so we probably should name the function accordingly (`ArrowArrayAppendStorageFromArrayView()`) to make it a bit more obvious to the caller that they're on their own with respect to whether they *should* do this or not.

Also a note that this needs to check `src->dictionary`, which I don't believe is reflected in the storage type (for better or worse).

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,174 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+    if (src->offset != 0) {
+      ArrowErrorSet(error, "Can't append a sliced run-end encoded array view");
+      return ENOTSUP;
+    }
```
paleolimbot: No need to get too into the weeds supporting this one, but we do have a utility that I think can resolve the first run:

https://github.com/apache/arrow-nanoarrow/blob/f55f85be3a1008bee99f556eb3cda3949356bbd8/src/nanoarrow/common/inline_buffer.h#L53-L54

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,174 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
   return array->release == &ArrowArrayReleaseInternal;
 }
 
+static ArrowErrorCode ArrowArrayAppendArrayViewElement(struct ArrowArray* dst,
+                                                       const struct ArrowArrayView* src,
+                                                       int64_t i,
+                                                       struct ArrowError* error) {
```
paleolimbot: While we are here, I think it is worth the effort to special case primitive fixed-size appends of the same storage type (e.g., int32 getting appended to an int32). These should be much faster because they are just appending buffers to each other (the nulls bitmap is a bit more complex...may have to go through a uint8 buffer for the bitmap append if either the source or the target has any nulls).

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,174 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+    if (src->offset != 0) {
+      ArrowErrorSet(error, "Can't append a sliced run-end encoded array view");
+      return ENOTSUP;
+    }
+
+    int64_t run_end_offset = dst->length;
+    for (int64_t i = 0; i < src->children[0]->length; i++) {
+      NANOARROW_RETURN_NOT_OK(ArrowArrayAppendInt(
+          dst->children[0],
+          run_end_offset + ArrowArrayViewGetIntUnsafe(src->children[0], i)));
```
paleolimbot: These (and below) should use `NANOARROW_RETURN_NOT_OK_WITH_ERROR()`...we are in general very careful about ensuring that the contents of `error` are populated if an error code is returned.

src/nanoarrow/common/array_test.cc
```diff
@@ -5087,3 +5087,232 @@ TEST(ArrayMoveSharedTest, ArrayWithNullBuffers) {
 
   ArrowArrayRelease(&shared);
 }
+
+static ArrowErrorCode AppendArrayViewForTest(const struct ArrowArrayView* src,
+                                             struct ArrowArray* dst,
+                                             struct ArrowError* error) {
```
paleolimbot: In addition to these lovely nanoarrow only tests, an Arrow C++ gated test I think would help cover some of the things that can happen here in a way that can be parameterized to cover more cases. Perhaps one test per type family (unigned int, signed int, strings, binaries, lists), testing against the Arrow C++ builder for the matrix of source and destination types? Arrow C++ makes asserting the equality a bit easier too.

src/nanoarrow/common/array_test.cc
```diff
@@ -5087,3 +5087,232 @@ TEST(ArrayMoveSharedTest, ArrayWithNullBuffers) {
...
+
+  ArrowArrayViewReset(&dst_view);
+  ArrowArrayRelease(&dst);
+  ArrowArrayViewReset(&src_view);
+}
+
+TEST(ArrayTest, ArrayAppendArrayViewPrimitiveTypes) {
+  struct ArrowArray array;
+
+  ASSERT_EQ(ArrowArrayInitFromType(&array, NANOARROW_TYPE_INT64), NANOARROW_OK);
```
paleolimbot: Some things that I am not sure are covered here are:

- Appending an integer that is out of range to a narrower integer type (should pass on the error from the integer appender)
- Appending too many elements to a narrower list type (you can use a list of Null type to avoid allocating huge buffers), which should pass on an error from the list element finisher
- Dictionary encoded source or target

paleolimbot (commented):
Thank you!

Some things to take care of here but I think this is great functionality.

I can follow up with R/Python bindings, which are mostly useful because testing this stuff in C++ is a pain. I think this may also make a few of the R and Python constructors quite a bit simpler.

src/nanoarrow/common/array.c
rustyconover: Addressed in 4974cd7. Matching byte-aligned primitive storage takes a bulk buffer-copy path. Validity is copied in bounded uint8 chunks so sliced bitmaps and all four source/destination bitmap-presence combinations are handled.

src/nanoarrow/common/array_test.cc
rustyconover: Addressed in e56f47b. Added Arrow C++-gated source/destination matrices for signed integers, unsigned integers, string/large_string, binary/large_binary, and list/large_list, with Arrow C++ equality checks on the imported result.

src/nanoarrow/common/array_test.cc
rustyconover: Addressed in e56f47b. Added tests for integer narrowing overflow, int32 list-offset overflow using a Null child, successful dictionary storage copying with caller-managed dictionary values, and rejection when only one side is dictionary-encoded.

src/nanoarrow/common/array.c
rustyconover: Addressed in bab29ad. The API is now ArrowArrayAppendStorageFromArrayView, with recursive preflight checks for compatible storage, child layouts, fixed-size widths, and dictionary presence. Dictionary values are deliberately caller-managed: both source and destination must be dictionary-encoded, and the API docs state that values are not copied. Follow-up 693718d narrows fixed-size layout equality checks to fixed-size-to-fixed-size appends and covers fixed-to-variable binary/list conversion.

src/nanoarrow/common/array.c
rustyconover: Addressed in bab29ad. Sliced run-end encoded views now resolve the first run with ArrowResolveChunk32 or ArrowResolveChunk64 (with the int16 fallback), clip the boundary runs to the slice, and rewrite run ends relative to the destination.

src/nanoarrow/common/array.c
rustyconover: Addressed in bab29ad and refined in e56f47b. Direct append and buffer operations use NANOARROW_RETURN_NOT_OK_WITH_ERROR so failures populate the error; calls to helpers that already populate it use the plain propagation macro so their more specific message is preserved.

rustyconover:
CI follow-up: b4f4adc fixes the shared failure behind clang-tidy and the four Valgrind jobs. ArrowBitmapAppendInt8Unsafe could read more values than supplied when a short append began in a partially filled byte; the same issue existed in the int32 variant. Both helpers now bound the partial-byte copy, with regression tests for short unaligned appends. Native, Arrow C++, AddressSanitizer, and local clang-tidy checks pass.

rustyconover:
clang-tidy follow-up: the Valgrind failures are resolved by b4f4adc, but clang-tidy 18 could not prove that ArrowBitsUnpackInt8 initialized every element later read from the scratch buffer. Commit 0603bc1 explicitly zero-initializes that buffer; this preserves behavior while guaranteeing analyzer-visible initialization.

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,498 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+      } else if (src->storage_type == NANOARROW_TYPE_DECIMAL256) {
+        bitwidth = 256;
+      }
+
+      struct ArrowDecimal decimal;
+      ArrowDecimalInit(&decimal, bitwidth, /*precision=*/0, /*scale=*/0);
+      ArrowArrayViewGetDecimalUnsafe(src, i, &decimal);
+      NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowArrayAppendDecimal(dst, &decimal), error);
+      return NANOARROW_OK;
+    }
```
paleolimbot: I think you can only do this if the precision and scale mach, in which case the fixed-size append case should take care of it. It might be possible to go through the string representation to handle some differences, but I think this maybe should just not be supported (except through the fixed size append case) for now.

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,498 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+            dst->children[0], src->children[0], child_offset + child_i, error));
+      }
+      NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowArrayFinishElement(dst), error);
+      return NANOARROW_OK;
+    }
+    case NANOARROW_TYPE_DENSE_UNION:
+    case NANOARROW_TYPE_SPARSE_UNION: {
+      int8_t type_id = ArrowArrayViewUnionTypeId(src, i);
+      int8_t child_index = ArrowArrayViewUnionChildIndex(src, i);
+      int64_t child_offset = ArrowArrayViewUnionChildOffset(src, i);
```
paleolimbot: While I appreciate the effort here, I think that it is safer to not support unions yet unless you absolutely need them for what you're trying to do here.

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,498 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+
+static int ArrowArrayCanAppendFixedWidthStorage(struct ArrowArray* dst,
+                                                const struct ArrowArrayView* src) {
+  struct ArrowArrayPrivateData* private_data =
+      (struct ArrowArrayPrivateData*)dst->private_data;
+  return private_data->storage_type == src->storage_type && src->n_children == 0 &&
+         src->layout.buffer_type[1] == NANOARROW_BUFFER_TYPE_DATA &&
+         src->layout.element_size_bits[1] > 0 &&
+         src->layout.element_size_bits[1] % 8 == 0;
+}
```
paleolimbot, line 373: This should probably also check for `n_buffers == 2` just to be safe.

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,498 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+
+  if (src_validity == NULL) {
+    NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowBitmapAppend(dst_validity, 1, src->length),
+                                       error);
+    return NANOARROW_OK;
+  }
+
+  NANOARROW_RETURN_NOT_OK_WITH_ERROR(ArrowBitmapReserve(dst_validity, src->length),
+                                     error);
+  int8_t validity[1024] = {0};
```
paleolimbot, line 396: Just making sure that there is also a test that exercises >1 chunk here.

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,498 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+          0, run_ends->length);
+    case NANOARROW_TYPE_INT16:
+    default: {
+      int64_t run = 0;
+      while (run + 1 < run_ends->length &&
+             ArrowArrayViewGetIntUnsafe(run_ends, run) <= logical_offset) {
+        run++;
+      }
+      return run;
+    }
```
paleolimbot, line 477: We should at least file an issue to add a ArrowResolveChunk16

paleolimbot (commented):
A few more things from this pass but I think it's close! On the next pass I'll send the AIs over it as well to look for coverage or correctness I'm not sure about.

src/nanoarrow/common/array.c
rustyconover: Addressed in dacf383. Added a 2,050-element validity test, which exercises three chunks (1,024, 1,024, and 2 values) and checks values/nulls on both chunk boundaries.

src/nanoarrow/common/array.c
rustyconover: Opened #934 to add ArrowResolveChunk16 as a separate, focused change, with parity tests matching the existing 32- and 64-bit resolvers.

src/nanoarrow/common/array.c
rustyconover: Addressed in dacf383. Removed the per-element decimal path; decimals are now handled only by the identical-storage fixed-width bulk path. Because precision and scale are not represented by ArrowArrayView, the storage API continues to leave logical compatibility to the caller.

src/nanoarrow/common/array.c
rustyconover: Addressed in dacf383. The fixed-width bulk-path predicate now also requires destination n_buffers == 2.

src/nanoarrow/common/array.c
rustyconover: Addressed in dacf383. Dense and sparse unions are no longer supported by this API and are rejected during recursive preflight, before any destination mutation. The former union success test now covers both rejection cases.

rustyconover:
@paleolimbot I think this is ready for your AI based review.

paleolimbot (commented):
A few issues left:

- Null storage appended to run-end encoded storage creates an invalid array. A run of nulls has to get added or you could also just reject this combination.
- The conversion pathway is different in the nested case (e.g., list of decimal or struct containing decimals converts using the element-wise appender). I *think* the easiest way to do this is to update the signature of `ArrowArrayAppendStorageFromArrayView()` to accept an additional `int64 offset` and `int64 length` and recurse back using that (this is what the R and Python conversion functions do), but if there is another relatively non-verbose way to deal with that I'm happy to take a look.
- I believe the compatibility check doesn't take into account struct names, and it probably should (appending `xmin`, `xmax` vs `xmax`, `xmin` is likely to be a problem should it be used for that).

When this merges I'll file a follow up for non-correctness issues that should get handled at some point, like the ability to handle non-storage conversions (e.g., appending timestamps with non-equal units), or speeding up conversions between integer widths / string on string appends.

rustyconover:
Addressed the latest review in 2ae9cd6.

- Null storage is now rejected when the destination uses run-end encoded storage, before any destination mutation.
- Nested struct/list/fixed-size-list children now recurse through an explicit offset/length range helper, so nested fixed-width storage such as decimals follows the same validated bulk path as top-level storage.
- Added regressions for null-to-REE rejection and decimal storage nested in both a struct and a list.
- For struct field names: ArrowArrayView does not carry field names, and this API intentionally operates on storage rather than logical schema metadata. I documented that struct children are matched positionally and that callers must validate field names (as well as decimal precision/scale). The public API signature remains unchanged; offset/length are internal to the recursive helper.

Validation:
- Native CMake/CTest: 280/280 passed.
- Arrow C++-gated CMake/CTest: 320/320 passed.

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,483 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+  }
+}
+
+static ArrowErrorCode ArrowArrayAppendStorageFromArrayViewRange(
+    struct ArrowArray* dst, const struct ArrowArrayView* src, int64_t offset,
+    int64_t length, struct ArrowError* error) {
+  struct ArrowArrayView src_slice = *src;
```
paleolimbot, line 552: This line is a little janky (it would be cleaner to pass the offset and length through the internal calls), but also OK as long as it is qualified. I can see how it's possibly less error prone to apply the slice once.

```suggestion
  // Note: src_slice must not be freed: this is used as a convenience to reduce the
  // number of times the offset and length must be compose on top of src.
  struct ArrowArrayView src_slice = *src;
```

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,483 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+          logical_offset + 1, run_ends->buffer_views[1].data.as_int64 + run_ends->offset,
+          0, run_ends->length);
+    case NANOARROW_TYPE_INT16:
+    default: {
+      int64_t run = 0;
+      while (run + 1 < run_ends->length &&
+             ArrowArrayViewGetIntUnsafe(run_ends, run) <= logical_offset) {
+        run++;
+      }
+      return run;
```
paleolimbot: Can you merge upstream/main and update this now that `ArrowResolveChunk16()` is merged?

src/nanoarrow/common/array.c
```diff
@@ -77,6 +77,483 @@ int ArrowArrayIsInternal(struct ArrowArray* array) {
...
+static int ArrowArrayCanAppendFixedWidthStorage(struct ArrowArray* dst,
+                                                const struct ArrowArrayView* src) {
+  struct ArrowArrayPrivateData* private_data =
+      (struct ArrowArrayPrivateData*)dst->private_data;
+  return private_data->storage_type == src->storage_type && dst->n_buffers == 2 &&
+         src->n_children == 0 &&
+         src->layout.buffer_type[1] == NANOARROW_BUFFER_TYPE_DATA &&
+         src->layout.element_size_bits[1] > 0 &&
+         src->layout.element_size_bits[1] % 8 == 0;
+}
```
paleolimbot, line 417: I think the other case worth handling out of the gate here (or one level up) is a struct appended to a struct, which I believe is trivial (feel free to punt if it is not).

paleolimbot (commented):
A few remaining issues...thank you for sticking with this! This is hard...you are basically implementing cast to/from all arrow types. Also very useful (we can use to implement explicit batch sizes on write, casting, and delta dictionaries!).

Codex turned up a few more things:

- self-append causes use-after-free. I think this one can just be a comment in the docs...it is not normal/supported behaviour to view an array that is in append mode.
- Two REE-related issues: REE destination-length overflow is undefined behavior and the Null→REE fix does not cover REE nested beneath a null parent. We've gone pretty far down the supporting REEs here so it's probably worth fixing, but you could also not support REEs here.

src/nanoarrow/common/array.c
rustyconover: Added the qualifying comment in 85ea0f7, including that the shallow slice must not be freed and why it is used.

src/nanoarrow/common/array.c
rustyconover: Merged current upstream/main in 9690f71 and replaced the int16 linear scan with ArrowResolveChunk16() in 85ea0f7. The REE append test now exercises int16 run ends.

src/nanoarrow/common/array.c
rustyconover: Added a bulk struct-to-struct path in 85ea0f7. It appends each child range and the parent validity in bulk; the regression test also covers a null struct slot with an REE child.

rustyconover:
Addressed the remaining review points in 85ea0f7 (after merging current upstream/main in 9690f71):

- documented that the source view must not reference destination-owned storage, since an append can reallocate it
- qualified the shallow src_slice copy
- switched int16 REE run lookup to ArrowResolveChunk16()
- added bulk struct-to-struct appends, including valid child storage beneath null struct slots
- reject other null appends before they can recursively create invalid REE child storage
- guard REE destination length addition before any signed overflow or mutation

Added regression coverage for int16 REE lookup, struct/REE storage beneath a null parent, nested Null-to-REE rejection, and destination-length overflow.

Local validation: all 14 focused append tests pass; the full C/C++ suite passes 284/284 tests (3 expected thread-safety skips).

src/nanoarrow/common/array_test.cc
```diff
@@ -5087,3 +5088,906 @@ TEST(ArrayMoveSharedTest, ArrayWithNullBuffers) {
...
+    ArrowSchemaRelease(&schema);
+  }
+}
+
+TEST(ArrayTest, ArrayAppendStorageFromArrayViewRunEndEncoded) {
+  struct ArrowError error;
+  struct ArrowSchema schema;
+  ArrowSchemaInit(&schema);
+  ASSERT_EQ(ArrowSchemaSetTypeRunEndEncoded(&schema, NANOARROW_TYPE_INT16), NANOARROW_OK);
+  ASSERT_EQ(ArrowSchemaSetType(schema.children[1], NANOARROW_TYPE_STRING), NANOARROW_OK);
```
paleolimbot: Here the int32 REE and int64 REE branches aren't covered. I think a `for()` over `NANOARROW_TYPE_INT16`, `NANOARROW_TYPE_INT32` and `NANOARROW_TYPE_INT64` should cover those.

paleolimbot (commented):
Codex could only find one test coverage gap here (inline below). I'll open some follow on issues for R and Python integration (which should help get this code covered in a wider variety of scenarios). Thanks!

paleolimbot (approved):
Thank you!

PR #934, opened 2026-09-05
feat: add ArrowResolveChunk16

rustyconover:
Adds the int16 counterpart to ArrowResolveChunk32 and ArrowResolveChunk64.

This follows up on the run-end encoded array review in #930, where int16 run ends currently require a linear scan. The helper uses the same binary-search semantics as the existing resolvers and includes boundary-parity tests.

Validation:
- 270/270 C/C++ tests passed
- clang-format and git diff checks passed


paleolimbot (approved):
Thank you!

PR #935, opened 2026-09-08
feat(ipc): support writing LZ4 and ZSTD compressed IPC streams and files

pdet:
This PR adds LZ4 and ZSTD compression to nanoarrow’s IPC encoder and writer, allowing applications to write compressed Arrow streams and files. The motivation comes from the duckdb-arrow extension, where we want nanoarrow to handle compression while the extension takes care of the SQL options. To support this, callers can select a codec and compression level, query the supported level range, and convert between codec names and enum values without needing to include the compression libraries’ headers themselves.

Compression happens per buffer within each record batch, following the Arrow IPC format. If compressing a buffer does not make it smaller, we store it uncompressed, and empty buffers remain empty. The implementation provides a serial compressor by default, with an interface for applications to supply their own. Unsupported codecs and out-of-range levels are rejected when configuring the built-in compressor, so callers can report these errors before writing data. The added tests cover compression and decompression roundtrips, compressed streams and files, compression levels, and builds with either codec unavailable.

### DIFF
--- a/src/nanoarrow/ipc/codecs.c
+++ b/src/nanoarrow/ipc/codecs.c
@@ -191,6 +191,42 @@ ArrowIpcCompressFunction ArrowIpcGetLZ4CompressionFunction(void) {
 #endif
 }
 
+ArrowErrorCode ArrowIpcGetCompressionLevelRange(
+    enum ArrowIpcCompressionType compression_type, int* min_level_out,
+    int* max_level_out) {
+  NANOARROW_DCHECK(min_level_out != NULL && max_level_out != NULL);
+  NANOARROW_UNUSED(min_level_out);
+  NANOARROW_UNUSED(max_level_out);
+
+  switch (compression_type) {
+    case NANOARROW_IPC_COMPRESSION_TYPE_ZSTD:
+#if defined(NANOARROW_IPC_WITH_ZSTD)
+#if ZSTD_VERSION_NUMBER >= 10400
+      *min_level_out = ZSTD_minCLevel();
+#else
+      // Negative (fast) levels can't be queried before zstd 1.4.0
+      *min_level_out = NANOARROW_IPC_COMPRESSION_LEVEL_DEFAULT;
+#endif
+      *max_level_out = ZSTD_maxCLevel();
+      return NANOARROW_OK;
+#else
+      return ENOTSUP;
+#endif
+    case NANOARROW_IPC_COMPRESSION_TYPE_LZ4_FRAME:
+#if defined(NANOARROW_IPC_WITH_LZ4)
+      // A negative level selects an acceleration of 1 - level, which lz4 caps at 65537
+      // (LZ4_ACCELERATION_MAX, which is not part of its public headers)
+      *min_level_out = 1 - 65537;
+      *max_level_out = LZ4F_compressionLevel_max();
+      return NANOARROW_OK;
+#else
+      return ENOTSUP;
+#endif
+    default:
+      return EINVAL;
+  }
+}
+
 // The serial decompressor and compressor keep one function per codec, indexed by
 // enum ArrowIpcCompressionType (NONE is never a codec)
 static int ArrowIpcCompressionTypeIsCodec(enum ArrowIpcCompressionType compression_type) {
--- a/src/nanoarrow/ipc/codecs.c
+++ b/src/nanoarrow/ipc/codecs.c
@@ -304,8 +304,7 @@ static void ArrowIpcSerialCompressorRelease(struct ArrowIpcCompressor* compresso
 }
 
 ArrowErrorCode ArrowIpcSerialCompressor(struct ArrowIpcCompressor* compressor) {
-  compressor->compress = &ArrowIpcSerialCompressorCompress;
-  compressor->release = &ArrowIpcSerialCompressorRelease;
+  compressor->release = NULL;
   compressor->private_data = ArrowMalloc(sizeof(struct ArrowIpcSerialCompressorPrivate));
   if (compressor->private_data == NULL) {
     return ENOMEM;
@@ -317,6 +316,8 @@ ArrowErrorCode ArrowIpcSerialCompressor(struct ArrowIpcCompressor* compressor) {
   ArrowIpcSerialCompressorSetFunction(compressor,
                                       NANOARROW_IPC_COMPRESSION_TYPE_LZ4_FRAME,
                                       ArrowIpcGetLZ4CompressionFunction());
+  compressor->compress = &ArrowIpcSerialCompressorCompress;
+  compressor->release = &ArrowIpcSerialCompressorRelease;
   return NANOARROW_OK;
 }
 
--- a/src/nanoarrow/ipc/codecs.c
+++ b/src/nanoarrow/ipc/codecs.c
@@ -249,7 +249,7 @@ const char* ArrowIpcCompressionTypeToString(
     case NANOARROW_IPC_COMPRESSION_TYPE_ZSTD:
       return "zstd";
     default:
-      return NULL;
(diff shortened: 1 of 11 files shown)


src/nanoarrow/ipc/codecs.c
```diff
@@ -113,6 +183,99 @@ ArrowIpcDecompressFunction ArrowIpcGetLZ4DecompressionFunction(void) {
...
+    enum ArrowIpcCompressionType compression_type) {
+  switch (compression_type) {
+    case NANOARROW_IPC_COMPRESSION_TYPE_NONE:
+      return "none";
+    case NANOARROW_IPC_COMPRESSION_TYPE_LZ4_FRAME:
+      return "lz4";
+    case NANOARROW_IPC_COMPRESSION_TYPE_ZSTD:
+      return "zstd";
+    default:
+      return NULL;
```
paleolimbot: It's better to have this return `""` or `"<unknown>"` for the inevitable user or contributor who uses this in a format string without checking for a NULL. I had to fix ArrowTypeToString recently because of gcc warnings.

src/nanoarrow/ipc/encoder.c
```diff
@@ -69,6 +79,11 @@ ArrowErrorCode ArrowIpcEncoderInit(struct ArrowIpcEncoder* encoder) {
   ArrowBufferInit(&private->nodes);
   ArrowIpcDictionaryEncodingsInit(&private->dictionary_encodings);
   ArrowBufferInit(&private->message_metadata);
+  private->codec = NANOARROW_IPC_COMPRESSION_TYPE_NONE;
+  private->compressor.release = NULL;
+  private->custom_compressor = 0;
+  private->compression_level = NANOARROW_IPC_COMPRESSION_LEVEL_DEFAULT;
+  private->flatbuf_codec = ns(CompressionType_LZ4_FRAME);
```
paleolimbot: Is there another value we can use for `flatbuf_codec` (or can we omit the field entirely and recalculate it when needed) that better communicates defaultness (rather than a specific value)?

src/nanoarrow/ipc/encoder.c
```diff
@@ -49,6 +49,16 @@ struct ArrowIpcEncoderPrivate {
...
+  // Compression applied to the body buffers of subsequently encoded RecordBatches
+  enum ArrowIpcCompressionType codec;
+  // Compressor used when codec != NONE (release is NULL until one is needed)
+  struct ArrowIpcCompressor compressor;
+  // Whether compressor was provided by ArrowIpcEncoderSetCompressor()
+  int custom_compressor;
+  // Compression level passed to the compressor when codec != NONE
+  int compression_level;
+  // The flatbuffer equivalent of codec (only meaningful when codec != NONE)
+  ns(CompressionType_enum_t) flatbuf_codec;
```
paleolimbot: Can we reduce this to only `codec` and `compressor` or even just `compressor` by inlining the codec and/or level into the `ArrowIpcCompressor` struct?

src/nanoarrow/ipc/encoder.c
```diff
@@ -744,6 +886,15 @@ static ArrowErrorCode ArrowIpcEncoderEncodeRecordBatch(
...
 
+  if (private->codec != NANOARROW_IPC_COMPRESSION_TYPE_NONE) {
+    FLATCC_RETURN_UNLESS_0(RecordBatch_compression_start(builder), error);
+    FLATCC_RETURN_UNLESS_0(BodyCompression_codec_add(builder, private->flatbuf_codec),
+                           error);
+    FLATCC_RETURN_UNLESS_0(
+        BodyCompression_method_add(builder, ns(BodyCompressionMethod_BUFFER)), error);
+    FLATCC_RETURN_UNLESS_0(RecordBatch_compression_end(builder), error);
+  }
+
```
paleolimbot, line 1038: In addition to these changes, I believe the use of compression should also be declared when writing the Schema message.

src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -412,6 +432,102 @@ ArrowIpcSerialDecompressorSetFunction(struct ArrowIpcDecompressor* decompressor,
...
+  /// \brief Compress a buffer
+  ///
+  /// Compresses src using compression_type at compression_level and appends the
+  /// compressed bytes to dst. Any content already in dst must be preserved (i.e.,
+  /// implementations may only append to dst). See ArrowIpcCompressFunction for the
+  /// interpretation of compression_level.
+  ArrowErrorCode (*compress)(struct ArrowIpcCompressor* compressor,
+                             enum ArrowIpcCompressionType compression_type,
+                             int compression_level, struct ArrowBufferView src,
+                             struct ArrowBuffer* dst, struct ArrowError* error);
```
paleolimbot: I think it would be better for force any parameters (like level and compression type) to be captured in the private data, perhaps with a member for compression_type since the encoder will need to declare it (callback seems like overkill for that).

If there is any way to potentially parallelize compression, it would be great to have this struct/its use support that even though the default implementation won't. You might need two passes (`compress_start()`, `compress_finish()`, passing `i` to identify the buffer). I don't want to spend too much time on the best possible parallel encoding, I just want to avoid changing the ABI for a trivial two pass setup. I don't mind the exact contract as long as there's at least one plausible setup where this works. Also fine if there is no easy plausible setup (we can try later pre 1.0).

src/nanoarrow/nanoarrow_ipc.h
```diff
@@ -412,6 +432,102 @@ ArrowIpcSerialDecompressorSetFunction(struct ArrowIpcDecompressor* decompressor,
...
+NANOARROW_DLL ArrowIpcCompressFunction ArrowIpcGetZstdCompressionFunction(void);
+
+/// \brief Get the compression function for LZ4
+///
+/// The result will be NULL if nanoarrow was not built with NANOARROW_IPC_WITH_LZ4.
+NANOARROW_DLL ArrowIpcCompressFunction ArrowIpcGetLZ4CompressionFunction(void);
+
+/// \brief An ArrowIpcCompressor implementation that performs compression in serial
+NANOARROW_DLL ArrowErrorCode
+ArrowIpcSerialCompressor(struct ArrowIpcCompressor* compressor);
```
paleolimbot: I think this can take a `compression_type` and `level` parameter if we inline those parameters into the ArrowIpcSerialCompressor private data.

paleolimbot (commented):
Thank you!

A few conceptual questions (feel free to push back of they don't make sense!)...I'll follow up with a review focused on correctness with the help of the robots after

This needs a rebase or a merge from main (sorry!) from the initial work on Dictionary writes.

pdet:
Thanks for the review @paleolimbot! Merged with the latest main and walked through your points.

src/nanoarrow/ipc/encoder.c
```diff
@@ -695,40 +800,207 @@ static ArrowErrorCode ArrowIpcEncoderBuildContiguousBodyBufferCallback(
...
+      continue;
+    }
+
+    int result = private->compressor.compress_add(&private->compressor, views[i],
+                                                  &private->compressed_buffers[i], error);
+    if (result != NANOARROW_OK) {
+      // don't leave queued work referring to our buffers behind
+      struct ArrowError ignored;
+      NANOARROW_UNUSED(
+          private->compressor.compress_wait(&private->compressor, -1, &ignored));
```
paleolimbot, line 867: tiny nit: You should be able to pass `NULL` for the error here.

paleolimbot (approved):
Thank you!

Codex turned up an unrelated issue surfaced here we can tackle separately ( https://github.com/apache/arrow-nanoarrow/issues/936 ).

PR #940, opened 2026-09-11
fix: allow null data for empty binary and string views during append

fornwall:
[ArrowBufferView](https://arrow.apache.org/nanoarrow/latest/reference/c.html#_CPPv4N15ArrowBufferView4dataE) and [ArrowStringView](https://arrow.apache.org/nanoarrow/latest/reference/c.html#_CPPv4N15ArrowStringView4dataE) allow a null data pointer when their size is zero:

> If size_bytes is 0, this value may be NULL.

Appending these views to binary-view or string-view arrays currently passes the null pointer to `memcpy`, which is undefined behaviour even for a zero-byte copy as flagged by UBSan.

Skip the inline copy for empty values to fix that.

### DIFF
--- a/src/nanoarrow/common/array_test.cc
+++ b/src/nanoarrow/common/array_test.cc
@@ -1193,6 +1193,30 @@ TEST(ArrayTest, ArrayTestAppendToStringViewArray) {
   });
 };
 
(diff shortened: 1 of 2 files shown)


paleolimbot (approved):
Thank you!

PR #942, opened 2026-09-15
fix: relax validation for missing validity buffer when null_count is -1

### LINKED ISSUE #855: Relax validation for null validity buffer when null_count is -1
This can happen in the wild ( https://github.com/firelab/gdalraster/issues/918#issuecomment-4071676813 ) and the current validation is tighter than what Arrow C++ accepts on import (I checked arrow-rs and it also does not mind).

We error if null_count != 0 but it is probably sufficient to check that the null count is not > 0 for all practical purposes. The use case in the issue is that GDAL has been producing arrays like this for several years and there's no real reason we shouldn't accept these arrays.

and2049:
Allow a missing validity buffer when the null count is unknown, matching
Arrow C++’s import behavior. Arrays reporting a positive null count still
require a validity buffer.

Add regression coverage for unknown null counts with a missing validity
buffer, including sliced arrays and positive-null-count rejection.

Fixes #855


src/nanoarrow/common/array_test.cc
```diff
@@ -2441,6 +2441,70 @@ TEST(ArrayTest, ArrayViewTestBasic) {
...
+      ArrowArrayViewValidate(string_view.get(), NANOARROW_VALIDATION_LEVEL_FULL, &error),
+      NANOARROW_OK)
+      << error.message;
+  EXPECT_EQ(ArrowArrayViewGetBufferView(string_view.get(), 0).size_bytes, 0);
+  EXPECT_EQ(ArrowArrayViewComputeNullCount(string_view.get()), 0);
+  EXPECT_EQ(ArrowArrayViewGetStringUnsafe(string_view.get(), 0), "bb"_asv);
+  EXPECT_EQ(ArrowArrayViewGetStringUnsafe(string_view.get(), 1), "ccc"_asv);
+
+  array->null_count = 1;
+  EXPECT_EQ(ArrowArrayViewSetArray(string_view.get(), array.get(), &error), EINVAL);
```
paleolimbot: Can you to use the exact same array as above and just modify the null count? I don't think we need to use a different type here to cover the changed lines.

paleolimbot (commented):
Thank you!

One note on the test here...I think you're testing two things but we only need to update the null count to ensure your change is covered.

src/nanoarrow/common/array_test.cc
and2049: I've removed the string-array case, and the test now uses the same int32 array and only changes the null count.

paleolimbot (approved):
Thank you!

